PR #27, opened 2012-12-04
check_ssh: make use of libssh (SF.net feature request #3574197)

hjanuschka:
http://sourceforge.net/tracker/?func=detail&aid=3574197&group_id=29880&atid=397600

re-done the check_ssh plugin to use libssh - to get the server fingerprint and version

call it with -f a247e883d98bf5c41923470de0bfa826 -r 5.5.0 localhost
plugin makes a warning if version is lower than 5.5.0 and a critical if the host fingerprint is not a247e883d98bf5c41923470de0bfa826


RincewindsHat:
I don't want to annoy anyone, but this PR is 8 years old, not directly mergable and probably has to be redone completely.  For those reasons, I would propose to reject the PR and close it. It would at least reduce the clutter.

PR #28, opened 2012-12-18
check_http: chunked encoding support

abrender:
This modification to check_http.c enables chunked encoding support with the '-U' option. It is disabled by default for backwards compatibility.

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -1042,11 +1042,11 @@ check_http (void)
   page += (size_t) strspn (page, "\r\n");
   header[pos - header] = 0;
 
-  page2 = (char *) calloc(strlen(page)+1, 1);
-  page2_ptr = page2;
-
   if (chunking && check_if_chunked(header))
   {
+    page2 = (char *) malloc(strlen(page)+1);
+    page2_ptr = page2;
+
     while (chunk_length = strtol(page, NULL, 16))
     {
        /* The +2's reference the trailing CRLF at the end of each line */
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -69,6 +69,7 @@ X509 *server_cert;
 #  define my_send(buf, len) send(sd, buf, len, 0)
 #endif /* HAVE_SSL */
 int no_body = FALSE;
+int chunking = FALSE;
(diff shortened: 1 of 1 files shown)


sykosoft:
@weiss @waja We've been patiently waiting for this to be included, as it affects a lot of checks we try to do. Any news on when this might be fixed?

waja:
@sykosoft looking into the troubles introduced by integrating this feature by nagios-plugins into their project, it might be a good idea to have this well tested before integrating this into an office release.

ekeih:
Hi,
we can confirm that this issue still exists.
@weiss could you write some details about your comment from 2013 about several issues with this PR? Maybe this would help other people to understand what we need to do to fix this very old but still annoying issue.
Thank you in advance!

weiss:
> Maybe this would help other people to understand what we need to do to fix this very old but still annoying issue.

I think the proper fix is to switch to [`check_curl`][1] which is currently being developed and tested in the [`feature_check_curl`][2] branch.

[1]: https://github.com/monitoring-plugins/monitoring-plugins/blob/feature_check_curl/plugins/check_curl.c
[2]: https://github.com/monitoring-plugins/monitoring-plugins/tree/feature_check_curl

ekeih:
Thank you for this super fast response! :) We will check out check_curl!

weiss:
> We will check out check_curl!

Cool, if you run into any issues, please report them.  It should be mostly compatible with `check_http` by now, i.e. it's supposed to work as a drop-in replacement except for corner cases.

ekeih:
What is currently needed to merge check_curl into the master branch? Only more testing or anything else?

We will try to run check_curl parallel to check_http for a while. This way we can check if they behave the same way for ~2500 checks.

ekeih:
Okay, not sure if this is a coincidence but here is a PR: https://github.com/monitoring-plugins/monitoring-plugins/pull/1535 :)

waja:
`check_curl`has landed in 2.3.

michaelsmoody:
Wow. Well. I'd honestly lost track of this by now. But, that's excellent news.

RincewindsHat:
So, this is probably resolved? Can this PR be closed?

PR #978, opened 2013-09-24
check_ifoperstatus.pl: Add -N and --namekey search options

moorereason:
Dell's PowerConnect switches store the interface name (ie. Te1/0/1) in ifName
and not in ifDescr.  The new options basically mirror the functionality of the
-d and --desc options but search against the ifName field.

Help text may need to be updated to avoid confusion with -n option.


moorereason:
Abandoning this PR since this repo and the old nagios-plugins repo have diverged.

Issue #1029, opened 2013-09-24
Fixes for check_snmp [sf#3463985]


monitoring-user:
Submitted by None on 2011-12-22 14:34:18

A small patch:

- response should not start with the delimiter
- OID response value taken into account

--- nagios-plugins-1.4.15/plugins/check_snmp.c.rmap     2011-12-22 14:23:26.000000000 +0100
+++ nagios-plugins-1.4.15/plugins/check_snmp.c  2011-12-22 14:27:30.000000000 +0100
@@ -322,6 +322,7 @@
                response = strstr (ptr, delimiter);
                if (response == NULL)
                        break;
+               response = response + 3;

                if (verbose > 2) {
                        printf("Processing oid %i (line %i)\n  oidname: %s\n  response: %s\n", i+1, line+1, oidname, response);
@@ -353,6 +354,9 @@
                else if (strstr (response, "INTEGER: ")) {
                        show = strstr (response, "INTEGER: ") + 9;
                }
+               else if (strstr (response, "OID: ")) {
+                       show = strstr (response, "OID: ") + 5;
+               }
                else if (strstr (response, "STRING: ")) {
                        show = strstr (response, "STRING: ") + 8;
                        conv = "%.10g";

sni:
merged

PR #1173, opened 2013-09-30
check_snmp: small improvement

waja:
Just turning attached patch of github issue #1029 into a push request

"A small patch:

- response should not start with the delimiter
- OID response value taken into account"


sni:
Reviewed and merged. Thanks

Issue #1026, opened 2013-09-24
check_http: Fix for checking http on different virtual port [sf#3442015]


monitoring-user:
Submitted by odenbach on 2011-11-25 12:28:53

Patch against Plugin Version (-V output): 1.4.15
Plugin Name: check_http
Example Plugin Commandline:

with patch:
check_http -H www.uni-paderborn.de:80 -p 5033 -I siva1 -v | head -4
GET / HTTP/1.1
User-Agent: check_http/v1.4.15 (nagios-plugins 1.4.15)
Connection: close
Host: www.uni-paderborn.de

without patch:
check_http -H www.uni-paderborn.de:80 -p 5033 -I siva1 -v | head -4
GET / HTTP/1.1
User-Agent: check_http/v1.4.15 (nagios-plugins 1.4.15)
Connection: close
Host: www.uni-paderborn.de:80:5033

Tested on operating system: Debian Linux, Squeeze
Tested on architecture: amd64
Tested with compiler: x86_64-linux-gnu-gcc

Explanation:

We use a loadbalancer to balance http requests to our two webservers. On the loadbalancer we have defined several IP addresses for different web sites. On the webservers however we use different ports for the different sites. So www.uni-paderborn.de:80 is balanced to siva1.uni-paderborn.de:5033 and siva2.uni-paderborn.de:5033.
Now we want to check both web servers individually to tell if a single host has failed (if one host fails the site is still available over the other one so users do not recognize the error).
In the current version check_http does understand the concept of virtual hosts (using -H and -I), but it does not understand the concept of virtual ports. Let me make this clearer:

If I leave out the virtual port like this:

check_http -H www.uni-paderborn.de -p 5033 -I siva1 -v | head -4 
GET / HTTP/1.1
User-Agent: check_http/v1.4.15 (nagios-plugins 1.4.15)
Connection: close
Host: www.uni-paderborn.de:5033

The request then contains the physical port which is wrong.

If I add the virtual port to the hostname after the -H switch, I get the muddled up result from above.

The patch corrects this behaviour while leaving everything else as it was before. If only one port is given (either in -H or with -p), that one is taken as physical and virtual port.

odenbach:
Hi,

sorry, I had not seen your reaction last year...

It still does not work. Your validation above was invalid (the 'head -4' accidently stripped out the error message):

[...]

PR #1175, opened 2013-09-30
check_http: Fix for checking http on different virtual port

waja:
Just turning attached patch of github issue #1026 into a push request
(with small adjustments).


waja:
Superseeded by #1352

Issue #1007, opened 2013-09-24
check_mysql - allow until condition on a slave [sf#3351505]


monitoring-user:
Submitted by megamaddin on 2011-07-03 16:51:45

Patch against Plugin Version (-V output): check_mysql v1.4.15 (nagios-plugins 1.4.15)
Plugin Name: check_mysql
Example Plugin Commandline: check_mysql -H localhost -S
Tested on operating system: CentOS 5.5
Tested on architecture: AMD64
Tested with compiler: gcc-Version 4.1.2 20080704 (Red Hat 4.1.2-50)

The check_mysql plugin fails with a wrong exit code, if the mysql slave server is started with an UNTIL condition and the sever reaches its desired binlog position. If the binlog position is reached, the slave sql thread stops running, until it's started again, and that is the problem. This patch solves this error by looking if the slave thread is started with an until condition and throws only an error if the Last_Errno variable in mysql is set.

monitoring-user:
Added File: https://www.monitoring-plugins.org/attachments/416883-check_mysql.patch

PR #1179, opened 2013-09-30
check_mysql: allow until condition on a slave

waja:
Just turning attached patch of github issue #1007 into a push request.

"The check_mysql plugin fails with a wrong exit code, if the mysql slave server is started with an UNTIL condition and the sever reaches its desired binlog position. If the binlog position is reached, the slave sql thread stops running, until it's started again, and that is the problem. This patch solves this error by looking if the slave thread is started with an until condition and throws only an error if the Last_Errno variable in mysql is set"


RincewindsHat:
The idea is valid, but the code too far away from that point in time to be applicable.
I am going to reimplement this.

Issue #894, opened 2013-09-24
check_dns: add AAAA,MX,TXT,SRV,CNAME [sf#2741269]


monitoring-user:
Submitted by dhorn2000 on 2009-04-07 19:21:10

Added optional support for the major DNS record types (AAAA, MX, TXT, SRV, CNAME, A, ANY).  Specifically, allow for a new optional --querytype parameter, and a new optional --accept-cname parameter.  Help text has been updated to reflect the changes as well. Patch is against SVN trunk of check_dns.c from April 7,2009.

This change should be completely backwards compatible with existing check_dns usage and syntax.  New features are off by default.

ekohl:
Given check_dig can do this, would it be better to recommend users to use that instead? check_dns is broken anyway because it uses nslookup which is [flawed and shouldn't be used](https://jdebp.eu/FGA/nslookup-flaws.html).

sni:
closing this one, please use check_dig instead.

PR #1185, opened 2013-10-01
check_dns: add AAAA,MX,TXT,SRV,CNAME

waja:
Just turning attached patch of github issue #894 into a push request.

"Added optional support for the major DNS record types (AAAA, MX, TXT, SRV, CNAME, A, ANY). Specifically, allow for a new optional --querytype parameter, and a new optional --accept-cname parameter. Help text has been updated to reflect the changes as well. Patch is against SVN trunk of check_dns.c from April 7,2009.

This change should be completely backwards compatible with existing check_dns usage and syntax. New features are off by default."

### DIFF
--- a/THANKS.in
+++ b/THANKS.in
@@ -302,3 +302,4 @@ Sebastian Schmidt
 Simon Kainz
 Steve Weinreich
 Tim Laszlo
+David Horn
--- a/plugins/check_dns.c
+++ b/plugins/check_dns.c
@@ -45,15 +45,17 @@ int error_scan (char *);
 void print_help (void);
 void print_usage (void);
 
-#define ADDRESS_LENGTH 256
+#define ADDRESS_LENGTH 384 
 char query_address[ADDRESS_LENGTH] = "";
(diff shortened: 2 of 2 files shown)


pbiering:
take care when takeover, looks like the upstream version has a strange upcase issue:

/usr/lib64/nagios/plugins/check_dns -q SOA google.de
DNS OK: 0,023 seconds response time. GOOGle.de returns
ns2.google.com|time=0,022598s;;;0,000000

/usr/lib64/nagios/plugins/check_dns -q MX redhat.com
DNS OK: 0,043 seconds response time. REDHAt.com returns 10 mx2.REDHAt.com.,5 mx1.REDHAt.com.|time=0,042732s;;;0,000000

(amout of upcase chars vary....haven't digged through the code how this could happen...perhaps one can inform upstream)

# /usr/lib64/nagios/plugins/check_dns -V
check_dns v2.0.1 (nagios-plugins 2.0.1)

RincewindsHat:
As was discussed in #894 this should probably not implemented at all and check_dns should probably be deprecated for `check_dig` or something like that.

PR #1195, opened 2013-10-01
check_ldap: add certificate support

### LINKED ISSUE #1067: add check certificate support for LDAP plugin [sf#2430999]
Submitted by guillomovitch on 2008-12-15 14:45:04

Here's a patch adding the certificate expiration check feature to the ldap plugin, based on the http plugin. For some unknown reason, probably due to the way secure connection is established, it doesn't work with the --ssl options, only for the --starttls one.

waja:
Here's a patch adding the certificate expiration check feature to the ldap
plugin, based on the http plugin. For some unknown reason, probably due to
the way secure connection is established, it doesn't work with the `--ssl`
options, only for the `--starttls` one.

Just turning attached patch of github issue #1067 into a push request.
(Closes #1067)


RincewindsHat:
Just did some research on this
I think the idea in itself has a use case IF the ldap server speaks only *STARTTLS* where other plugins who _only_ speak plain TLS can not work (without implementing some ldap there).

This leads me to question whether *STARTTLS* is common enough to justify adding more code. As far as I understand *STARTTLS* was a crutch to implement oportunistic encryption and is discouraged most of the time.

Regardless, this PR is no longer applicable and I do not want to update it to the current code structure, therefor I am closing this.
If someone wants to have that feature, please open an issue.

PR #1201, opened 2013-10-14
check_http: Move time_ssl to end of extended perfdata output

awiddersheim:
When using pnp4nagios as your graphing engine and you don't have the MULTI option turned where each source of information is placed into it's own independent RRD file it is difficult to add a new datasource into the middle of a pre-existing RRD file. Having the data added at the end on the other hand isn't nearly as bad. I'd imagine with other graphing platforms like graphite this isn't really an issue.

This is really an attempt to fix some short comings with how my current setup is configured now but I'm hoping it may help others who are in a similar situation.

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -1197,10 +1197,10 @@ check_http (void)
            perfd_time (elapsed_time),
            perfd_size (page_len),
            perfd_time_connect (elapsed_time_connect),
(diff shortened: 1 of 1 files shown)


RincewindsHat:
No and this is also a very 2013 problem. If this even still exists, it should be fixed in the RRD toolchain.
IMHO we do not and should not give guarantees on the order of performancec data points.

PR #1246, opened 2014-03-13
Add commandline -T switch to check_time and check_ntp_time

langemeijer:
Connection timeout will result in warning status instead of critical

I want to compare my servers to a time server not in my control. When for an unknown reason the time server is not responding for me that is not actionable.

### DIFF
--- a/THANKS.in
+++ b/THANKS.in
@@ -307,3 +307,4 @@ Stéphane Bortzmeyer
 Luca Corti
 Jethro Carr
 Evgeni Golov
+Casper Langemeijer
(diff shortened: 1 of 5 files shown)


langemeijer:
Guys, re-reading this again a year later, I'm still stomped by the process of this pull request.

I put in time, I think I worked well with @skottler and @weiss on this. Explaining the why of my patches, and implementing the suggestions offered by them. I'm not afraid to learn nor explain.

Then comes in @dermoth, first disregarding any of my explanation, then claiming a larger scale, to top it off offer me advice on how to do nagios monitoring.

dermoth:
Hi Casper,

I have to apologize for not looking at your patches like I said I  would...  My point was this should be implemented on a much more generic level.  Like most plugins use macros to set base command switches, I think this patch should be applied to all plugin that use timeouts. The code look OK though so my main concern at this point would be checking if the choice of -T works for at least most plugins with a -t switch, and maybe somehow at making some kind of "reservation", ex by adding comments where it could be used.

@skottler, @weiss, please do not wait on me if you'd like to merge it; I'm not much active in this project since I no longer deal with monitoring at work (I didn't deal a lot with it in back in spring '14, and even less since I've changed job shortly after - last summer).

Regards,

Thomas

PR #1286, opened 2014-08-23
check_http: chunked Transfer-Encoding

waja:
Cloned from https://github.com/nagios-plugins/nagios-plugins/pull/54 provided by @koentje


waja:
There seems to be issues over there at n-p: https://github.com/nagios-plugins/nagios-plugins/issues/98, https://github.com/nagios-plugins/nagios-plugins/issues/103, https://github.com/nagios-plugins/nagios-plugins/issues/104

PR #1303, opened 2014-11-26
check_by_ssh: Add option to force pseudo tty allocation

riskersen:
Because i’ve needed a pseudo tty for running sudo via ssh, i created a small patch for check_by_ssh

It introduces a new option „-z“ which adds 

     -t      Force pseudo-tty allocation.  This can be used to execute arbitrary screen-based programs on a remote machine, which can be very useful, e.g. when implementing menu services.  Multiple -t options
             force tty allocation, even if ssh has no local tty.

While using this new option, you have to set -E because on STDERR appears:
Connection to $IP closed.

Maybe there is a better way to surpress this message but i’m not this deep in C.


riskersen:
Yeah, you're right that seems to be a better solution :)

PR #1305, opened 2014-11-28
check_http: fix Host header if explicitly set with -k

waja:
taken from https://github.com/llicour/nagios-plugins/commit/911c3c34755c65046fe73ce051ec9dc3b76b324a


sni:
merged

PR #1308, opened 2014-12-15
check_load: remove superfluous argument to printf

waja:
The format string expects only one argument.

Signed-off-by: Sebastian Herbszt herbszt@gmx.de


sni:
I'd rather say, we add one more %s format string than stripping the result.

glensc:
agreed

sni:
Fixed with 0dc6537. Thanks.

PR #1319, opened 2015-01-21
check_ircd: Rewrite

lrupp:
I did a rewrite of the check_ircd.pl script to allow the script to 
- allow connections using SSL
- provide performance data

The options are downwards compatible and the script itself is used in our infrastructure since more than a year now without any problems.

An example of a command definition:

``` bash
define command {
    command_name check_ircd_ssl
    command_line $USER1$/check_ircd -H $HOSTADDRESS$ -w $ARG1$ -c $ARG2$ -s -p 6697
}
```

Please note that instead of the perl(Socket) module, the new script needs perl(IO::Socket::INET6).

Left TODO: IPv6 support.

### DIFF
--- a/plugins-scripts/check_ircd.pl
+++ b/plugins-scripts/check_ircd.pl
@@ -176,7 +176,7 @@ MAIN:
     my $hostname;
     Getopt::Long::Configure('bundling');
     GetOptions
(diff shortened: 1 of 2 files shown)


sni (commented):
A change to check_hpjd.c has sliped into this PR. Also a rewrite is probably a good time to add some basic unit tests.

PR #1320, opened 2015-01-22
check_http: Repair HTTP CONNECT function

mfrost8:
seems to be only partially in check_http.c.  That patch was at
http://sourceforge.net/p/nagiosplug/patches/336/ It also seems that the
single-letter variable I added (-K) is now taken by an other option.
I've switched to use the unused -B (the long argument format of
--http-connect is still valid) and re-inserted all the original changes
to allows checks through a proxy server that implements HTTP CONNECT
to work again.

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -129,6 +129,7 @@ char *http_content_type;
 char buffer[MAX_INPUT_BUFFER];
 char *client_cert = NULL;
 char *client_privkey = NULL;
+int http_connect = FALSE;
 
 int process_arguments (int, char **);
 int check_http (void);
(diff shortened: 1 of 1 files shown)


waja:
Thanks @mfrost8,

maybe you can also try to add the needed changes for the (now) failing test into this PR?

lazyfrosch:
@waja This PR should be obsolete since 2.2

waja:
> This PR should be obsolete since 2.2

Cause it was fixed there?

waja:
This should be done with 342dd1b97fd3423cb6368fedbb6c729f3c3b6243

Issue #105, opened 2013-09-24
utils.c TXTBLK is too small! [sf#592262]


monitoring-user:
Submitted by nobody on 2002-08-07 21:36:18

In particular, when trying to use check_snmp with two OIDs, the 
command line used to invoke snmpget is truncated in the middle of 
the second OID resulting in a failure.

i bumped TXTBLK to 
1024 as a quick fix.

PR #1326, opened 2015-03-06
check_disk: reduced compiler warnings

riskersen:
with fixing verbose printfs and fixed printf arguments
   printf line 1046: missing argument, switched llu int to uintmax_t
   printf line 1067: wrong arguments, switched llu int to uintmax_t

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -293,7 +293,7 @@ main (int argc, char **argv)
 
       if (verbose >= 3) {
         printf ("For %s, used_pct=%g free_pct=%g used_units=%g free_units=%g total_units=%g used_inodes_pct=%g free_inodes_pct=%g fsp.fsu_blocksize=%llu mult=%llu\n",
-          me->me_mountdir, path->dused_pct, path->dfree_pct, path->dused_units, path->dfree_units, path->dtotal_units, path->dused_inodes_percent, path->dfree_inodes_percent, fsp.fsu_blocksize, mult);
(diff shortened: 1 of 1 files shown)


waja:
Dunno if we  fix that in 2.1.x or keep for 2.2

riskersen:
To resolve #1331 it should be merged

waja:
Your changes are glibc specific. You could cast the problematic arguments to (long long) to avoid the warning.

riskersen:
Ah okay, didnt know that. I changed uintmax_t to long long and %ju to %llu

result @ Ubuntu 12.04 3.13.0-65-generic #105~precise1-Ubuntu SMP Tue Sep 22 13:22:42 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```
./check_disk -vvv -w 10% -c 5%  -p /
calling stat on /
Thresholds(pct) for / warn: 10,000000 crit 5,000000
calling stat on /
For /, used_pct=82 free_pct=18 used_units=56842 free_units=13248 total_units=73214 used_inodes_pct=11 free_inodes_pct=89 fsp.fsu_blocksize=4096 mult=1048576
Freespace_units result=0
Freespace% result=0
Usedspace_units result=0
Usedspace_percent result=0
Usedinodes_percent result=0
Freeinodes_percent result=0
DISK OK - free space: / 13248 MB (18% inode=89%);| /=56842MB;65892;69553;0;73214
```

weiss:
The arguments should actually be cast to `unsigned long long` (to match the `%llu` format specifiers).

RincewindsHat:
should be fixed in the current master.

PR #1330, opened 2015-03-27
check_ups: Fix incorrect performance data thresholds

sbraz:
Hi, it appears that the performance data thresholds are incorrectly multiplied by 1000 for percentages.
This was probably due to a duplication of the code used for voltage which outputs mV and thus require the multiplier.


weiss:
Applied, thanks.

PR #1335, opened 2015-04-18
check ldap count entries

lausser:
Hi, this patch adds --warn-entries and --crit-entries. As soon as you use one of them, not only the runtime of the ldap query will be checked but also the number of returned entries.
The parameters accept threshold ranges as arguments, e.g.:
--warn-entries 10: --crit-entries 1:
This will result in CRITICAL if the ldap-query was correct, but returned no entry records.


sni:
merged, thanks

PR #1336, opened 2015-04-20
sslutils: optimize output if certificate expires in less then 24h

m4c3:
optimize output if certificate expires in less then 24h
thx to axel schmalowsky for this patch


sni:
I like the idea. After a quick look, i think the code will print wrong information if the time till expire is below one hour. It would then print "minutes" as unit but stil print the hours_left.
Also i am not a fan of cascaded if statements. I would at least break it on several lines.

m4c3:
this looks like a temporary failure with travis or the mozilla test url
can you restart the build?

weiss:
Done (and it helped).

sni:
thanks, PR has been merged.

PR #1337, opened 2015-04-22
sslutils: make time calculations timezone independent (and thus more portable)

ylfingr:
Time calculations have been done locale-dependent which lead to different expiration dates.

### DIFF
--- a/plugins/sslutils.c
+++ b/plugins/sslutils.c
@@ -142,18 +142,21 @@ int np_net_ssl_read(void *buf, int num) {
 
 int np_net_ssl_check_cert(int days_till_exp_warn, int days_till_exp_crit){
 #  ifdef USE_OPENSSL
+        char timestamp[50] = "";
+        char cn[MAX_CN_LENGTH]= "";
+
 	X509 *certificate=NULL;
 	X509_NAME *subj=NULL;
-	char cn[MAX_CN_LENGTH]= "";
+
 	int cnlen =-1;
 	int status=STATE_UNKNOWN;
 
+        char *tz;
 	ASN1_STRING *tm;
 	int offset;
 	struct tm stamp;
 	float time_left;
 	int days_left;
-	char timestamp[50] = "";
 	time_t tm_t;
 
 	certificate=SSL_get_peer_certificate(s);
@@ -210,9 +213,22 @@ int np_net_ssl_check_cert(int days_till_exp_warn, int days_till_exp_crit){
 	stamp.tm_sec = 0;
 	stamp.tm_isdst = -1;
 
+        /* portable version mktime(3) which fixes the date/time representation in the current locale  */
+        tz = getenv("TZ");
+        if (tz)
+                tz = strndup(tz, strlen(tz));
+        setenv("TZ", "UTC", 1);
+        tzset();
+
 	time_left = difftime(timegm(&stamp), time(NULL));
 	days_left = time_left / 86400;
 	tm_t = mktime (&stamp);
+
+        if (tz) {
+                setenv("TZ", tz, 1);
+                free(tz);
+        }
+
 	strftime(timestamp, 50, "%c", localtime(&tm_t));
 
 	if (days_left > 0 && days_left <= days_till_exp_warn) {
--- a/plugins/sslutils.c
+++ b/plugins/sslutils.c
@@ -217,7 +217,7 @@ int np_net_ssl_check_cert(int days_till_exp_warn, int days_till_exp_crit){
         tz = getenv("TZ");
         if (tz)
                 tz = strndup(tz, strlen(tz));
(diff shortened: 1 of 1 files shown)


sni:
How does this make the code more portable? I mean, thats what timezones are for, if you want UTC, just use UTC either by changing your system to UTC or by running the plugin with
TZ="UTC" ..../check_http -C ..
Or just set TZ for your monitoring user, i'd say there are way better solutions than hardcoding UTC in some of the plugins.

ylfingr:
You're right. Hardcoding UTC isn't such a good idea. So I've changed it to clearing TZ.

Not doing so results in an incorrect date/time representation under the user's current locale.

# Example:

## certificate's expiration date
```shell
[user@host] openssl asn1parse -i -in /tmp/github.com.crt -strparse 165
    0:d=0  hl=2 l=  30 cons: SEQUENCE          
    2:d=1  hl=2 l=  13 prim:  UTCTIME           :140408000000Z
   17:d=1  hl=2 l=  13 prim:  UTCTIME           :160412120000Z
```

## current timezone (Europe/Berlin)
```shell
[user@host] date +%z
+0200
```

## current behavior
```shell
[user@host] TZ=UTC  ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 12:00:00 2016.
```
```shell
[user@host] TZ=Europe/Berlin  ./check_http -v -p443 -H github.com -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 **13:00:00** 2016.
```

## new behavior
```shell
[user@host] TZ=UTC  ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 12:00:00 2016.
```
```shell
[user@host] TZ=Europe/Berlin  ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 **14:00:00** 2016.
```

sni:
I am trying to reproduce that here:

%> TZ=UTC  ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue 12 Apr 2016 12:00:00 PM UTC.
%> ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue 12 Apr 2016 01:00:00 PM CEST.
%> TZ=Europe/Berlin  ./check_http -v -p443 -H github.com -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue 12 Apr 2016 01:00:00 PM CEST.
%> ./check_http -V
check_http v2.1.1.51.g08fc (monitoring-plugins 2.1.1)

I'd say, the real question is, why is there no timezone mentioned in your output. The check uses strftime with "%c" fomat. This is "The preferred date and time representation for the current locale.".

So, if i set LANG as well, i get close to your results:
LANG=C TZ=Europe/Berlin  ./check_http -v -p443 -H github.com -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 13:00:00 2016.

Which is not wrong, but you have to know the timezone :-)

In the end, i would leave it like it is. With every ENV variable we set to a fixed value, we just trade reduced flexibility for a specific use case.

ylfingr:
I'm not quite sure why there's no timezone mentioned in my outputs. I've tried various settings (e.g. LANG=de_DE vs LANG=C), but there's never any timezone indicator in the outputs. But this seems to be an issue on my system.

Nevertheless, for the sake of argument, let's assume there's a timezone indicator in the output. As you can see from your outputs, you'll never get the expiration time of **02:00:00 PM CEST** / **14:00:00**, which would be the correct time representation in the case of TZ=Europe/Berlin.

And as you can see in my last patch, the variable TZ is being cleared / unset. Since the check uses "%c" in strftime, the user can still choose his preferred TZ-settings.

At last, I have found a more verbose example:

## current behavior
```shell
TZ=UTC ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 12:00:00 2016.
```
```shell
TZ=PST8PDT ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 13:00:00 2016.
```

## new behavior
```shell
TZ=PST8PDT ./check_http -v -p443 -H github.com  -C 7,3 --sni
OK - Certificate 'github.com' will expire on Tue Apr 12 05:00:00 2016.
```

Since PST8PDT is UTC-8, the second output would be correct.

ylfingr:
any update on this?

sni:
I merged 1396, so this should be ok now.

PR #1340, opened 2015-05-23
check_http: added status code to plugin output perf data

vkhatri:
hello! i am not a C developer, but gave it a shot anyway. i think it is good to have http status code in perf data, for folks like me having status code graph is a great addon.

hope it helps!


awiddersheim:
I like the idea however, I feel like it should be an option that someone has to specify before this data is added because not everyone will want this. Just my 2 cents.

RincewindsHat:
I don't think this should be implemented. A return code is IMHO not the kind of number for which the perfdata thing is designed.
Not that I would not like to have a structured data output for check plugins, I prefer not to implement it like this.

PR #1342, opened 2015-06-16
check_http: Change SNI default behaviour

nohn:
SNI has arrived in the mainstream, so should check_http


nohn:
check_http.t tests fail which should have nothing to do with this PR

RincewindsHat:
This would like break some tests. I would prefer not to do that.

PR #1343, opened 2015-08-13
check_fping: autodetect ipv6 addresses

awiddersheim:
Stole the logic in check_ping that allows it to autodetect whether an
address is ipv6 or not. Now the user does not have to specify -6 when
using check_fping with ipv6 addresses.


awiddersheim:
I didn't add any tests though I did look at them. I can certainly expand upon the basic tests that are there and put in some more tests that do more with ipv6 when it comes to `check_fping`.

sni:
thanks

PR #1344, opened 2015-08-14
check_timex: add new plugin

kerolasa:
This plugin uses ntp_adjtime() function to read and report system time
using struct timex values.  Check is similar to check_ntp_time, but with
a difference being completely local execution.  This has some desirable
effects, such as:

1) This check runs very quickly.

2) The status of clock is reported from solely systems point of view,
i.e., a small network hiccup or public DNS failure will not cause
spurious failures

3) The check works when clock is syncronized with something else than
ntp, such as ptp.

Main difference of the check_timex and check_ntp_time is units used in
--warning and --critical arguments.  This check is using microseconds,
while the check_ntp_time is using seconds.  That means one should not
interchange these checks without additional configuration changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,6 +187,7 @@ NP-VERSION-FILE
 /plugins/check_swap
 /plugins/check_tcp
 /plugins/check_time
+/plugins/check_timex
 /plugins/check_udp
 /plugins/check_ups
 /plugins/check_users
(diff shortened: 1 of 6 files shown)


kerolasa:
Hello. What is going on with getting review to this pull request?

PR #1349, opened 2015-08-27
check_http: Adding support for checking SSL-Websites through Proxies

waja:
Ported the fix from Frank4dd to the latest git version. See also http://nagios.frank4dd.com/howto/monitor-ssl-websites-through-proxy.htm


sni:
Merged. We did use that patch for quite some time already.

PR #1351, opened 2015-09-08
check_geom for freebsd

djzort:
This checks the status of GEOM software raid in FreeBSD.


waja:
As there exists https://github.com/glensc/nagios-plugin-check_raid I don`t believe we will integrate that.

We removed the contrib/ directory with the 1.5 release. The documentation as also some bits about that at https://www.monitoring-plugins.org/doc/guidelines.html#CONTRIBUTEDPLUGINS

djzort:
OK np. I will push the features in to check_raid then.

Issue #1338, opened 2015-05-11
check_http - support TLS v1.2


druchoo:
Please add TLS v1.2 to check_http. Currently only v1 is supported?

```
 -S, --ssl=VERSION
    Connect via SSL. Port defaults to 443. VERSION is optional, and prevents
    auto-negotiation (1 = TLSv1, 2 = SSLv2, 3 = SSLv3).
```

jasonwilk:
Some QSAs are starting to vail PCI scans when TLSv1 is enabled. They've already disallowed SSLv2 and SSLv3. That means that there are no checks that would be available via SSL with this command unless it's extended to support TLSv1.1 and TLSv1.2. This needs to be updated for this plugin to be useful in any environment that has PCI implications.

seccentral:
Yes this is still a problem. and now most https checks are useless. 
When can we expect an update ?

waja:
Just to complete this thread, there is a reply on our development mailing list (https://www.monitoring-plugins.org/archive/devel/2015-August/010055.html):

"It should still connect when you use auto-negotiation."

seccentral:
Don't know about their/his/her setup but i've tried and by default it's auto (right ? manual sais so anyway) 
" -S, --ssl=VERSION
    Connect via SSL. Port defaults to 443. VERSION is optional, and prevents
    auto-negotiation (1 = TLSv1, 2 = SSLv2, 3 = SSLv3)."
so even if i use -S --ssl=1 it still fails. albeit with a different error: (with --ssl=1)
CRITICAL - Cannot make SSL connection.
140318164997776:error:14094438:SSL routines:SSL3_READ_BYTES:tlsv1 alert internal error:s3_pkt.c:1294:SSL alert number 80

seccentral:
Big update on this and solution *nohack* 
So how do you check a host with SNI ? like this :
openssl s_client -connect www.example.com:443 -servername www.example.com 
works perfectly. And this made me think.

now back to the icinga1/2/nagios/etc check_http execution.
this is the correct way to use the command to check a webhost with sni 
check_http -H www.example.com -S --sni 
HTTP OK: HTTP/1.1 200 OK - 13667 bytes in 1.031 second response time |time=1.031097s;;;0.000000 size=13667B;;;0
Now, -H stands for vhost but i guess in the context of sni it's somewhat the same thing tho the documentation should be updated.

Here is a host object definition for such a check 
sequence from hosts.conf 
====
[...]
[...]

PR #1354, opened 2015-09-18
check_http: add tls 1.1 and 1.2 support

keith4:


waja:
This is a proposed patch for #1338

waja:
There is also a suggested solution at https://github.com/monitoring-plugins/monitoring-plugins/issues/1338#issuecomment-141904034

waja:
Might be superseded by #1338

PR #1355, opened 2015-09-22
check_http: Add check_nrpe like behaviour

fredrik-rambris:
With my change you format a web page response like Nagios plugin output, put the numeric exit code in the X-NRPE-Status header, set the Content-Type to text/nagios and the check_http plugin behaves like check_nrpe.

Instead of having to parse the output of your own web applications to suit Nagios, why not have them output the status in that format already.

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -841,59 +841,6 @@ int get_content_length (const char *headers)
   return content_length;
 }
 
-int
-get_old_content_length (const char *headers)
-{
-  const char *s;
-  int content_length = 0;
-
-  s = headers;
-  while (*s) {
-    const char *field = s;
-    const char *value = 0;
-
-    /* Find the end of the header field */
-    while (*s && !isspace(*s) && *s != ':')
-      s++;
-
-    /* Remember the header value, if any. */
-    if (*s == ':')
-      value = ++s;
-
-    /* Skip to the end of the header, including continuation lines. */
-    while (*s && !(*s == '\n' && (s[1] != ' ' && s[1] != '\t')))
-      s++;
-
-    /* Avoid stepping over end-of-string marker */
-    if (*s)
-      s++;
-
-    /* Process this header. */
-    if (value && value > field+2) {
-      char *ff = (char *) malloc (value-field);
-      char *ss = ff;
-      while (field < value-1)
-        *ss++ = tolower(*field++);
-      *ss++ = 0;
-
-      if (!strcmp (ff, "content-length")) {
-        const char *e;
-        while (*value && isspace (*value))
-          value++;
-        for (e = value; *e && *e != '\r' && *e != '\n'; e++)
-          ;
-        ss = (char *) malloc (e - value + 1);
-        strncpy (ss, value, e - value);
-        ss[e - value] = 0;
-        content_length = atoi(ss);
-        free (ss);
-      }
-      free (ff);
-    }
-  }
-  return (content_length);
-}
-
 int get_x_nrpe_status (const char *headers)
 {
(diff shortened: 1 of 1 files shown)


waja:
Hi,

I don't know if this can be merged by us, but it would super cool if you can rebase this PR against our recent master branch, as d73c57aa061a4c408701f980d1f8857b64433488 fixed some problems with our integration tests.

Thanks, Jan.

fredrik-rambris:
Noob warning. Could you quickly describe how do I do that?

waja:
http://lmgtfy.com/?q=git+branch+rebase+master

for example: https://git-scm.com/book/en/v1/Git-Branching-Rebasing or https://www.atlassian.com/git/tutorials/merging-vs-rebasing/conceptual-overview#the-rebase

fredrik-rambris:
All tests passing as of now. Rebased and stuff.

RincewindsHat:
So, this is quite late, but whats the idea of this thing? Having the webserver/webwhatever output a `X-NRPE` status header to signal whether it is ok?

@waja  you were involved. Do you think this is a good idea or can be closed?

fredrik-rambris-bhg:
Yes. You can have the web backend give detailed status in a format that is easily understood by Nagios.

waja:
> @waja you were involved. Do you think this is a good idea or can be closed?

Just about workflow things.

Anyhow, what comes to my mind about this topic:

* Implementing this in `check_http` means this should be also supported by `check_curl`?
* This is not a small addition. Can you please describe, how to setup the nrpe part in the webserver, so we can get a clue, if this might get a significant userbase (I would not like to maintain code, which is not used by a significant userbase). We need a documentation about how to setup the serverside anyhow.

sni:
interesting. But in order to set the header, you always need a server side application. If the status would be read from the actual plugin output, we could also simply put a result file in a web root and let it be served by apache/ngingx or any webserver. So in absence of the header, simply check if the output starts with `<digit>:` and take that as status code. (Only in nrpe mode of course)
And yes, this should be possible with check_curl too then.

fredrik-rambris:
If the functionality is already in check_curl then no need to add it to check_http. The setup in backend is simply logic to query the application internal state and return it in Nagios format with the return code in header. In Java Spring that might be

```java
@GetMapping("/monitor/sessions")
public ResponseEntity<String> getNumberOfSessions() {
  if(this.sessions.count > 500) {
            return ResponseEntity
                .ok()
                .header("X-NRPE-Status", "2")
                .body("CRITICAL Too many sessions");
  } else {
            return ResponseEntity
                .ok()
                .header("X-NRPE-Status", "0")
                .body("OK Sessions under control");
  }
}
```

waja:
Why not just implementing this with http error codes? When everything in your server is fine, answer with 200, which is the default `check_http` and `check_curl` are expecting. When something went wrong, answer with 5xx and a meaningful error message?

fredrik-rambris:
Because HTTP error codes often occur in the HTTP layer of the application stack and to be able to discern them from application state you put the nagios error code somewhere else. The idea with this patch (which is quite old now) was to make the parsing on Nagios side super simple. Just take it like you do with NRPE responses and output it to Nagios. Not needing to go through an intermediate step of parsing say a JSON + HTTP status into something Nagios understand.

waja:
But ... Nagios is just ONE software, which is using `check_http`. What about all those others? Do they care about such an addition?

RincewindsHat:
Ok 3.0.0 is out and `check_http` now fully marked as deprecated (and `check_curl` as stable).
Therefor I am going to close this, since it is not going to be implemented.
Sorry for having this hanging in the air.

Issue #229, opened 2013-09-24
check_ping: Adjusting packet size [sf#781002]


monitoring-user:
Submitted by thedoc31 on 2003-07-31 19:51:35

check_ping doesn't currently support the -s
(packetsize) feature of the ping command.

I have an instance where normal ICMP packets sometimes
don't show packet loss across our WAN but larger
packets will show the packet loss.  It would be handy
to be able to specify the packet size on the
commandline of check_ping so I can make a small packet
check and a large packet check for WAN latency in Nagios.

go8ose:
I'd really like this as well. We had a router replaced last week which was misconfigured, and dropping packets larger than 1468 bytes. We didn't even know we had a problem until users were complaining the next morning. Now I want to setup some monitoring to catch this earlier, which would need this option.

riskersen:
Here's a suggested patch #1356 but the test is failing..maybe someone could have a look on it

RincewindsHat:
check_icmp implements this

PR #1356, opened 2015-09-28
check_ping: added option for packet size

riskersen:
- added option (-s/--packetsize) for setting packet size (limited to 65528 Bytes, according to RFC791)
- added DEFAULT_PACKET_SIZE constant with 56 Bytes
- disabled ping check for UnixWare in autoconf: if condition and with_ping_command had different syntax

This is for issue #229

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1028,6 +1028,7 @@ AC_ARG_WITH(ping_command,
 AC_MSG_CHECKING(for ICMP ping syntax)
 ac_cv_ping_packets_first=no
 ac_cv_ping_has_timeout=no
+ac_cv_ping_has_packetsize=no
 if test -n "$with_ping_command"
 then
 	AC_MSG_RESULT([(command-line) $with_ping_command])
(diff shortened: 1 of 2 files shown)


riskersen:
I'll take a look

riskersen:
Maybe someone have a look on it? Dunno why it keeps failing with a segfault..
```
./t/check_ping.t ......... Got signal 11 for command ./check_ping -H localhost -w 10,100% -c 10,100% -p 1 at ../NPTest.pm line 665.
# Looks like your test exited with 255 before it could output anything.
./t/check_ping.t ......... Dubious, test returned 255 (wstat 65280, 0xff00)
```

RincewindsHat:
Sorry for no answer and for the following: I am voting to reject this. It is bad enough to parse cmdline output and have different `ping` implementations. No need to complicate this.

`check_icmp` has an option to do this and that should be enough IMHO.

@sni @waja  d'accord?

waja:
we lived 7 years without anybody caring about this... so I would say there is not much interest in it

sni:
well, if it works, i'd say why not. But at least this PR cannot be merged easily.

PR #1360, opened 2015-10-02
drop superfluous libexecdir

glensc:
utils.pm locate is already convered by FindBin

i tried to identify if this problem:
https://github.com/nagios-plugins/nagios-plugins/pull/109

is present in monitoring-plugins as well. luckily it was not, but the superfluous use lib libexecdir was. this is cleanup of that.

### DIFF
--- a/plugins-scripts/check_breeze.pl
+++ b/plugins-scripts/check_breeze.pl
@@ -6,7 +6,6 @@ use Getopt::Long;
 use vars qw($opt_V $opt_h $opt_w $opt_c $opt_H $opt_C $PROGNAME);
 use FindBin;
 use lib "$FindBin::Bin";
-use lib '@libexecdir@';
 use utils qw(%ERRORS &print_revision &support &usage);
 
 $PROGNAME = "check_breeze";
--- a/plugins-scripts/check_disk_smb.pl
+++ b/plugins-scripts/check_disk_smb.pl
@@ -26,7 +26,6 @@ use vars qw($opt_P $opt_V $opt_h $opt_H $opt_s $opt_W $opt_u $opt_p $opt_w $opt_
 use vars qw($PROGNAME);
 use FindBin;
 use lib "$FindBin::Bin";
-use lib '@libexecdir@';
 use utils qw($TIMEOUT %ERRORS &print_revision &support &usage);
 
 sub print_help ();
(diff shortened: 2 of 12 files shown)


weiss:
Applied, thanks.

weiss:
Oh is there a reason you didn't update `check_rpc.pl` and `check_wave.pl`?

glensc:
no reason. seems missed them somehow. cherry pick 2d6b205 as well.

ps: why not just do PR merge? as doing cherry pick manually will lose view from ticket that the commit reached master (and you did not include commit ref in your "applied" message as well)

weiss:
> cherry pick 2d6b205 as well.

Done, thanks!

> why not just do PR merge?

Well I have a little `apply 123` shell function that applies PR 123 with `git am`, and then I often amend minor cleanups.  I like this workflow, and I think it's a feature that the committer is recorded without an empty merge commit.  But I know it's not the workflow envisioned by GitHub, and I know that people are unhappy if GitHub is unhappy, so i do consider changing the workflow one day ...

glensc:
yes, without doing merge commits, i cannot figure out from the commit itself, that it
https://github.com/monitoring-plugins/monitoring-plugins/commit/a9b02e7affe6d940e1fb4830156d06e4d816458b
came from this pull request:
https://github.com/monitoring-plugins/monitoring-plugins/pull/1360

contrary sample where it can be seen:
https://github.com/monitoring-plugins/monitoring-plugins/commit/6a682b379f1b6ef920cae79fae71b38b99dff39d

github puts small links to header: master (#1343)

and that is very useful when trying to find pull request based on commit hash from git blame

glensc:
also, each PR contains links below "merge using commandline" -- you could do those commands, except push, and do extra commits with git amend to the "merge" commit. then all features work:

1. you could do your amend, which be visible as diff of the "merge commit"
2. merge commit is there for github
3. original commits from reporter are unaltered

weiss:
> yes, without doing merge commits, i cannot figure out from the commit itself that it a9b02e7 came from this pull request: #1360

That's what I meant above when I said that this is "not the workflow envisioned by GitHub".

> you could [do stuff].

As I said I do consider changing my workflow to make GitHub users happy.

weiss:
Another issue is that we work with two branches, `maint` and `master`.  PRs are always submitted against the "default branch" as configured in GitHub, and we then apply the PR'd commits to either of the two branches as we see fit.  Merging the PR'd branch into the non-default branch won't work, of course ...

PR #1361, opened 2015-10-02
fix bogus use lib

glensc:
besides that it's wrong, it's detected easily with `perl -c` lint:

```
$ perl -c check_ntp.pl 
Bareword "utils" not allowed while "strict subs" in use at check_ntp.pl line 64.
Bareword "pm" not allowed while "strict subs" in use at check_ntp.pl line 64.
BEGIN not safe after errors--compilation aborted at check_ntp.pl line 65.
```


waja:
Sorry ... I did miss your PR anyhow, but I fixed it in the same way with 20d223119eb40bd1ecdd637eb845f5e16c87bd36

PR #1363, opened 2015-10-04
use exit code 3 for version and help

sni:
Plugins which use OK state for help or version information may run
unnoticed in that modes without doing any check. So its good
style to let version and help exit with Unknown state.


riskersen:
Agree on that, typical problem. Happens way too often

PR #1372, opened 2015-10-05
check_procs: Make cgroups aware.

waja:
This patch enables check_proc to monitor processes in PID name-
spaced environments.

Normally, all processes are visible from host - the ones running
directly on host and the ones in containers. This results in
check_procs reporting false-positives or false negatives depending
on whether he sees to many of given processes on to little.

These changes introduce one filter to check_proc - cgroup hierarchy.
Specified on the command line, it allows user to define which cgroup
hierarchy we are interested in and check_proc is ignoring all other.

As mentioned [here](https://github.com/nagios-plugins/nagios-plugins/pull/56#issuecomment-56846776) and [here](https://github.com/nagios-plugins/nagios-plugins/pull/56#issuecomment-67006931), using this option on non cgroup aware systems, the result maybe unexpected. This should be fixed (before aplying).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -746,6 +746,26 @@ dnl 	ac_cv_ps_format=["%*s %d %d %d %d %*d %*d %d %d%*[ 0123456789abcdef]%[OSRZT
 dnl 	ac_cv_ps_cols=8
 dnl 	AC_MSG_RESULT([$ac_cv_ps_command])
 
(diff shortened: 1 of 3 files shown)


glensc:
as a distribution manager, i would like to *compile* plugin with cgroups support, regardless what's the state where the plugin is *installed*. the cgroups availability check should be therefore runtime.

altho seems it's just proper procps option that needs to be passed at compile time

PR #1373, opened 2015-10-05
check_http: Allow for requesting TLSv1.1 and TLSv1.2

weiss:
check_http's `-S`/`--ssl` option now allows for requesting the TLSv1.1 and TLSv1.2 protocols.  Apart from that, a `+` suffix can be appended in oder to also accept newer protocols than the specified version (e.g., `--ssl=1.1+` to accept TLSv1.1 and TLSv1.2).


waja:
+1

sthen:
Thanks, this is sane syntax and a nice approach.

PR #1376, opened 2015-10-07
check_mssql.pl: switched alarm timeout

### LINKED ISSUE #991: Timeout not recognized in check_mssql.pl [sf#3275946]
Submitted by wopfel on 2011-04-05 12:52:37

Hi,
I've just had a look at the source code in the Git repository. It seems that --timeout (-t) command line argument isn't taken into account.

Sub syntax shows that the uppercase TIMEOUT is the default value:
printf("  --timeout -t          Plugin timeout (default:$TIMEOUT)\n");

In sub process_arguments, the value is stored in $timeout:
$timeout = $TIMEOUT unless (defined($timeout));

But the alarm is always called with the uppercase TIMEOUT:
alarm($TIMEOUT);

So I think it would be sufficient to change
alarm($TIMEOUT); --> alarm($timeout);

Since I'm developing something else at the moment, I can't dig deeper and check other files.

Regards
Bernd Arnold

Line 52 as of now, 2011-04-05 12:48 CEST:
http://nagiosplug.git.sourceforge.net/git/gitweb.cgi?p=nagiosplug/nagiosplug;a=blob;f=plugins-scripts/check_mssql.pl;h=9a8fc5117195457ffca41e5534e8b9e02a96f9c5;hb=HEAD#l52

riskersen:
alarm timeout now uses -t option, closes #991

If not set, default TIMEOUT from utils.pm will be used

```
$timeout = $TIMEOUT unless (defined($timeout))
```

### DIFF
--- a/plugins-scripts/check_mssql.pl
+++ b/plugins-scripts/check_mssql.pl
@@ -54,7 +54,7 @@ $SIG{'ALRM'} = sub {
      print ("SQL UNKNOWN: ERROR connection $server (alarm timeout)\n");
      exit $ERRORS{"UNKNOWN"};
 };
-alarm($TIMEOUT);
+alarm($timeout);
 
 unless ($dbh = DBI->connect("dbi:Sybase:server=".uc($server), "$username", "$password")) {
 	printf "SQL CRITICAL: Can't connect to mssql server $DBI::errstr\n";


waja:
Please give contribution here. The SF profile seems to be http://sourceforge.net/u/wopfel/profile/. Using google it seems (https://dev.icinga.org/attachments/download/1899/0001-Added-install-systemd-hint-in-Makefile.patch) he is now: Bernd Arnold <wopfel@gmail.com>

PR #1382, opened 2015-10-08
check_smtp: extended support for expect option

### LINKED ISSUE #1381: check_smtps expect option (-e) only checking first line of output after connect
check_smtps expect option (-e) only checking first line of output after connect

Further reponses after HELO cmd are not checked

riskersen:
expect option (-e) supported only first response, so checking for any other response e.g. 250-xxx would never match. 

This fix stores response of first buffer and compares against

closes #1381

### DIFF
--- a/plugins/check_smtp.c
+++ b/plugins/check_smtp.c
@@ -128,6 +128,7 @@ main (int argc, char **argv)
 	char *cmd_str = NULL;
 	char *helocmd = NULL;
 	char *error_msg = "";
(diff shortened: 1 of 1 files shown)


riskersen:
Hey all,

what do you think about my changes? I think they're safe and will not only be an enhancement, but more a bug fix. Because by now the check is only looking for the first line of response.

BR
Oliver

riskersen:
Another year passed, what do you think about this change?

Issue #1291, opened 2014-09-20
check_icmp and IPv6


jenslink:
looks like check_icmp  does not support IPv6.

nmr50:
Please add IPv6 support into check_icmp :)

bonki:
+1

PR #1384, opened 2015-10-13
check_icmp: Add IPv6 support

LarsMichelsen:
This commit adds IPv6 capabilities to check_icmp. It is now possible to
specify the address family using the arguments -4 (default) or -6.

To make the change possible we had to move the argument parsing previous
to creating the socket to be able to create it with the correct address
family.

This commit also fixes some gcc 4.9.2 compiler warnings. It has been
tested with several current linux distributions (debian, ubuntu, rh,
sles).

This commit fixes monitoring-plugins/monitoring-plugins#1291

### DIFF
--- a/plugins-root/check_icmp.c
+++ b/plugins-root/check_icmp.c
@@ -67,7 +67,9 @@ const char *email = "devel@monitoring-plugins.org";
 #include <netinet/in_systm.h>
 #include <netinet/in.h>
 #include <netinet/ip.h>
+#include <netinet/ip6.h>
 #include <netinet/ip_icmp.h>
+#include <netinet/icmp6.h>
 #include <arpa/inet.h>
 #include <signal.h>
 #include <float.h>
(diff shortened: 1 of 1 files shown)


LaMi-:
Found a bug in parameter handling. The provided -w -c values were not used, sorry. Has been fixed in 2nd commit.

LarsMichelsen:
The tests just failed because of some unrelated test failures. Just re-run the tests and they should be fine.

ghciv6:
i've extended the check to detect ipv6 addresses itself.

here is the patch against your Pull request:
[check_icmp_ipv6_autodetect_patch.txt](https://github.com/monitoring-plugins/monitoring-plugins/files/2109417/check_icmp_ipv6_autodetect_patch.txt)

ghciv6:
i've changed the code to support auto detection of address family.

here's the patch against your pull request:
[check_icmp_ipv6_autodetect_patch.txt](https://github.com/monitoring-plugins/monitoring-plugins/files/2110091/check_icmp_ipv6_autodetect_patch.txt)

waja:
fixed by #1563

PR #1387, opened 2015-10-21
check_flexlm has several design flaws

waja:
that this patch partially addresses (addresses points 2 to 4):
1/ it requires the PATH_TO_LMSTAT variable to be setup in util.pm while this file is nnot a config file. Thus each time the package is upgraded, the config get lost
2/ it doesn't support finding lmstat in PATH
3/ it doesn't support getting lmstat binary path from command line switchs
4/ usage is wrongly documented (needs the -F switch to provide the licence file PATH)
5/ LM_LICENCE_FILE variable not supported
6/ licence directory instead of licence path not supported.

The patch was proposed at https://github.com/nagios-plugins/nagios-plugins/issues/118 and I adjusted it to apply it to our code.

### DIFF
--- a/plugins-scripts/check_flexlm.pl
+++ b/plugins-scripts/check_flexlm.pl
@@ -17,7 +17,7 @@
 #host[anchor]=any host will do;some.address.com;;check-host-alive;3;120;24x7;1;1;1;
 #service[anchor]=yodel;24x7;3;5;5;unix-admin;60;24x7;1;1;1;;check_flexlm!/opt/lic/licfiles/yodel_lic
 #service[anchor]=yeehaw;24x7;3;5;5;unix-admin;60;24x7;1;1;1;;check_flexlm!/opt/lic/licfiles/yeehaw_lic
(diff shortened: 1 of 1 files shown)


waja:
This commit adds a dependency to File::Which. This is the cause Travis-CI (https://travis-ci.org/monitoring-plugins/monitoring-plugins/builds/86646825) is failing. Don't know if we really want to add another dependency or if we could solve this with another solution or just leaving this issue unfixed.

odenbach:
Hi,

some time ago I was also looking for a better solution to monitor our license servers. As the check_flexlm plugin did not provide enough details for me I wrote another plugin called check_lmstat which uses my perl module Flexnet::lmutil to parse the lmstat output. Apart from checking a license server's availability it outputs the number of used licenses as performance data. Together with something like PNP this draws nice graphics of license usage over time.

If anybody wants to give it a try: http://homepages.uni-paderborn.de/odenbach/projects/check_lmstat/

Cheers,

Christopher

PR #1395, opened 2015-11-24
Use GMT timezone in SSL certs

pirtoo:
SSL certs are required to use times in GMT per
https://www.ietf.org/rfc/rfc5280.txt but the mktime() here assumes the
current timezone.

Fix the time_t conversion to be done assuming GMT with timegm() and
only do it once rather than twice.

Display the expiry date and time with ISO format years and give an
offset from GMT and a timezone to be very clear about exactly what time
is being displayed. Time given is correct and now in the machine’s
timezone.


pirtoo:
Need to fix the tests too...

PR #1396, opened 2015-11-24
Use GMT timezone in SSL certs

pirtoo:
SSL certs are required to use times in GMT per
https://www.ietf.org/rfc/rfc5280.txt but the mktime() here assumes the
current timezone.

Fix the time_t conversion to be done assuming GMT with timegm() and
only do it once rather than twice.

Display the expiry date and time with ISO format years and give an
offset from GMT and a timezone to be very clear about exactly what time
is being displayed. Time given is correct and now in the machine’s
timezone.

Fix tests to match.

### DIFF
--- a/plugins/sslutils.c
+++ b/plugins/sslutils.c
@@ -264,10 +264,10 @@ int np_net_ssl_check_cert(int days_till_exp_warn, int days_till_exp_crit){
 		(tm->data[10 + offset] - '0') * 10 + (tm->data[11 + offset] - '0');
 	stamp.tm_isdst = -1;
 
-	time_left = difftime(timegm(&stamp), time(NULL));
+	tm_t = timegm(&stamp);
+	time_left = difftime(tm_t, time(NULL));
 	days_left = time_left / 86400;
-	tm_t = mktime (&stamp);
-	strftime(timestamp, 50, "%c", localtime(&tm_t));
+	strftime(timestamp, 50, "%F %R %z/%Z", localtime(&tm_t));
 
 	if (days_left > 0 && days_left <= days_till_exp_warn) {
 		printf (_("%s - Certificate '%s' expires in %d day(s) (%s).\n"), (days_left>days_till_exp_crit)?"WARNING":"CRITICAL", cn, days_left, timestamp);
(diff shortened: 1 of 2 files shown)


sni:
I'd prefer to keep %c time format but force timezone to GMT. Then people can at least can continue to use their local time format and still get the valid GMT date.

pirtoo:
We've found having the GMT date confusing for people, especially with the current redhat built version we're using where it lists a date/time with no timezone at all. %c on more recent builds can list the wrong timezone, which is doubly incorrect.

The only way I can think of to do what you're saying and not have it be confusing is to force TZ to GMT before the strftime() and put it back to whatever it was before afterwards which feels ugly at best. 

The local time representation of %c is unclear for international users, we have people all around the world but the servers are in the US meaning we get US format ambiguous dates that no one likes and most people can't correctly interpret without a timezone listed. This has caused several near-outages.

pirtoo:
For example, on OSX, if you use %c you get:
pir@shiny> ./plugins/check_http -H www.mozilla.org --onredirect=follow --ssl --sni -C 200000 -4
WARNING - Certificate 'www.mozilla.org' expires in 400 day(s) (Thu 29 Dec 12:00:00 2016).

No timezone listed. You have to know that time is in GMT whereas people generally assume timezone is in the server timezone (which would be PST or other timezones for most of our monitoring servers).

With this code:
	tz = getenv("TZ");
	setenv("TZ", "GMT", 1);
	tzset();
	strftime(timestamp, 50, "%c %z", localtime(&tm_t));
	if (tz)
		setenv("TZ", tz, 1);
	else
		unsetenv("TZ");
	tzset();

I get:
WARNING - Certificate 'www.mozilla.org' expires in 400 day(s) (Thu 29 Dec 12:00:00 2016 +0000).

Which at least lists a time offset from GMT (0 here, since the date is in GMT) and should make it obvious what timezone is and if a timezone is included in %c already the offset is additional information rather than repeating the timezone name.

How does that sound?

m4c3:
Why not use the existing Pull from #1337 ?

pirtoo:
Pull #1337 doesn't fix it doing two conversions into time_t and doesn't do the correct conversion of time_t in the right timezone (it should be using timegm() not mktime()) so I believe calculations can be incorrect for hosts in other timezones.

Basically I don't think it is the right fix.

friesoft:
Any news on this? I guess this superseeds [the original report](https://github.com/nagios-plugins/nagios-plugins/issues/119)

sni:
I merged this PR, thank you very much.

PR #1406, opened 2016-02-25
check for ping -{4,6} from new upstream iputils

eworm-de:


eworm-de:
This is the downstream report from Arch Linux bug tracker:
https://bugs.archlinux.org/task/48338

waja:
#1412 seems related

waja:
@eworm-de if it isn't too much hassle for you, can you please [rebase](http://stackoverflow.com/a/7929499) your PR against master? This might fix the travis checks.

eworm-de:
Already did that before you requested. ;)

waja:
Is this still needed? It's rotting since 5 years and longer but things seems to work without that change.

eworm-de:
For Arch Linux this works as we configure with `--with-ping-command='/usr/bin/ping -4 -n -U -w %d -c %d %s' --with-ping6-command='/usr/bin/ping -6 -n -U -w %d -c %d %s'`...

https://gitlab.archlinux.org/archlinux/packaging/packages/monitoring-plugins/-/blob/main/PKGBUILD?ref_type=heads#L31

waja:
I guess this is be done by every maintainer. In Debian we [do](https://salsa.debian.org/nagios-team/monitoring-plugins/-/blob/master/debian/rules#L54-60):
```
ifeq ($(DEB_HOST_ARCH_OS),linux)
PING_CONFIGURE_ARGS = --with-ping-command="/bin/ping -4 -n -U -w %d -c %d %s" \
                      --with-ping6-command="/bin/ping -6 -n -U -w %d -c %d %s"
else
PING_CONFIGURE_ARGS = --with-ping-command="/bin/ping -n -w %d -c %d %s" \
                      --with-ping6-command="/bin/ping6 -n -w %d -c %d %s"
endif
```

waja:
If nobody is complaining, I would prefer to close this one ...

eworm-de:
Configuring at build time is the way to go.

PR #1407, opened 2016-03-02
check_ifoperstatus.pl: add ifName search key option

moorereason:
Dell's PowerConnect switches store the interface name (ie. Te1/0/1) in ifName
and not in ifDescr, so doing `... -d Te1/0/1` doesn't work.  This commit adds
a new `-N|--namekey` option that allows you to search against the ifName field.


moorereason:
Abandoning this.

PR #1408, opened 2016-03-03
[check_radius.c] prevent segfault on rc_avpair_add.

glensc:
patch from debian (#nagios-plugins_1.4.11-2.diff.gz)

originated from pld linux
https://github.com/pld-linux/monitoring-plugins/blob/auto/th/monitoring-plugins-2.1.2-1/nagios-plugins-check_radius_segfault.patch


waja:
@glensc I dropped the patch via https://github.com/waja/pkg-monitoring-plugins/blob/master/debian/changelog#L529-L531.

Is this problem still reproducible for you?

glensc:
i never reproduced it, but as it was in our tree and still applies, i thought it's valid

PR #1409, opened 2016-03-03
check_ping: ignore "WARNING: failed to install socket filter" messages

glensc:
ignore "WARNING: failed to install socket filter: Protocol not available" message from check_ping

originated from pld-linux:
https://github.com/pld-linux/monitoring-plugins/commits/auto/th/monitoring-plugins-2.1.2-1/nagios-plugins-check_ping-socket-filter-warning.patch

### DIFF
--- a/plugins/check_ping.c
+++ b/plugins/check_ping.c
@@ -487,6 +487,7 @@ run_ping (const char *cmd, const char *addr)
 	while (fgets (buf, MAX_INPUT_BUFFER - 1, child_stderr)) {
 		if (
 			! strstr(buf,"WARNING - no SO_TIMESTAMP support, falling back to SIOCGSTAMP")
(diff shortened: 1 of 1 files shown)


waja:
It seems to be over years in PDL Linux, I think if this would be an issue, this would have been adjusted there. :)

RincewindsHat:
I tried to find that failure but was unable to. Searching hints, that this is either a misconfigured kernel build or some weird network condition.
Decided not to merge it, because I don't know what it does and the problem seems to go away on it's own.

glensc:
> It seems to be over years in PLD Linux, I think if this would be an issue, this would have been adjusted there. :)

this sentence doesn't make sense, what do you mean? it's patch in pld linux, so it is handled there. the goal was to upstream local distro patches.

PR #1411, opened 2016-03-03
check_disk_smb: allow checking 0-sized resource (ex. IPC$)

glensc:
patch by Marek Marczykowski marmarek@staszic.waw.pl

originated from pld-linux
https://github.com/pld-linux/monitoring-plugins/commits/auto/th/monitoring-plugins-2.1.2-1/nagios-plugins-check_disk_smb-zero-cap.patch

### DIFF
--- a/plugins-scripts/check_disk_smb.pl
+++ b/plugins-scripts/check_disk_smb.pl
@@ -212,7 +212,8 @@ if (/\s*(\d*) blocks of size (\d*)\. (\d*) blocks available/) {
 	my ($total_bytes) = $1 * $2;
 	my ($occupied_bytes) = $1 * $2 - $avail_bytes;
 	my ($avail) = $avail_bytes/1024;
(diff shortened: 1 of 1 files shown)


waja:
@glensc can you please tell us, what's the purpose of this change?

waja:
(It seems to be over years in PDL Linux, I think if this would be an issue, this would have been adjusted there. :)

glensc:
I don't have any more info than shared here. It was me carrying over patches from pld linux.

as much I recall when used Windows from the past, `IPC$` may be some share name: `\\hostname\ipc$`

RincewindsHat:
So, to my understanding (which is limited, because this is perl and windows), this prevents a division by zero in that line.
Since the value is only used in the output later it does not disturb other functionality.

RincewindsHat (approved):
Can't hurt I think

PR #1413, opened 2016-03-03
perl taint fixes

glensc:
patch from pld-linux
https://github.com/pld-linux/monitoring-plugins/commits/auto/th/monitoring-plugins-2.1.2-1/nagios-plugins-tainted.patch


RincewindsHat:
Replaced by #1917 , but thank you so much for the original idea

PR #1414, opened 2016-03-10
check_by_ssh: print command output in verbose mode

sni:
right now it is not possible to print the command output of ssh. check_by_ssh
only prints the command itself. This patchs adds printing the output too. This
makes it possible to use ssh with verbose logging which helps debuging any
connection, key or other ssh problems.
Note: you must use -E,--skip-stderr=<high number>, otherwise check_by_ssh would
always exit with unknown state.

Example:

  ./check_by_ssh -H localhost -o LogLevel=DEBUG3 -C "sleep 1" -E 999 -v

Signed-off-by: Sven Nierlein sven@nierlein.de

### DIFF
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -100,6 +100,13 @@ main (int argc, char **argv)
 
 	result = cmd_run_array (commargv, &chld_out, &chld_err, 0);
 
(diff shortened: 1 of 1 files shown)


sni:
merged already

PR #1415, opened 2016-03-15
check_timex: add new plugin

kerolasa:
This plugin uses ntp_adjtime() function to read and report system time
using struct timex values.  Check is similar to check_ntp_time, but with
a difference being completely local execution.  This has some desirable
effects, such as:

1) This check runs very quickly.

2) The status of clock is reported from solely systems point of view,
i.e., a small network hiccup or public DNS failure will not cause
spurious failures

3) The check works when clock is syncronized with something else than
ntp, such as ptp.

Main difference of the check_timex and check_ntp_time is units used in
--warning and --critical arguments.  This check is using microseconds,
while the check_ntp_time is using seconds.  That means one should not
interchange these checks without additional configuration changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,6 +187,7 @@ NP-VERSION-FILE
 /plugins/check_swap
 /plugins/check_tcp
 /plugins/check_time
+/plugins/check_timex
 /plugins/check_udp
 /plugins/check_ups
 /plugins/check_users
--- a/THANKS.in
+++ b/THANKS.in
@@ -357,3 +357,4 @@ Thomas Kurschel
 Yannick Charton
 Nicolai Søborg
 Rolf Eike Beer
+Sami Kerola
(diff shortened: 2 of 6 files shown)


RincewindsHat:
sounds good, gonna take a look at look. Remind me, if I don't :-)
(And sorry for no response the last five years)

RincewindsHat:
ok, had a look at it (after 10 years, I am sorry).
So this is (thankfully) rather simple although I do not understand all the implications.

To my understanding, this asks the (linux) kernel whether the (hardware?) clock is synchronised.

I can not see the value in monitoring this though, if the time differs strongly the NTP application is probably correcting the time which means it is working, if not we still don't know wheter the system time is "correct"(ish) or not since this does not compare it to a different source.

This is a bit ungrateful, but I am rejecting this PR. If I am completely wrong, please don't hesitate to intervene here

kerolasa:
No problem. While this pull request was looked the timex monitoring was added to prometheus: https://github.com/prometheus/node_exporter/blob/master/collector/timex.go

PR #1417, opened 2016-04-26
check_snmp: Add multiply-by and divide-by options

peelman:
Allows simple scaling on numeric results.

~~Useful~~ Necessary when a wrapper script is prohibitive and device outputs scaled results.  Or when you get bits and need bytes (or vice versa).

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -1201,6 +1201,11 @@ print_help (void)
 	printf ("    %s\n", _("Converts rate per second. For example, set to 60 to convert to per minute"));
 	printf (" %s\n", "--offset=OFFSET");
 	printf ("    %s\n", _("Add/substract the specified OFFSET to numeric sensor data"));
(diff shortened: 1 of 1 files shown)


MSteuter:
The plugin works great with single values. Thanks for these useful functions.
I have a netman ups and wanna use  .1.3.6.1.2.1.33.1.4.2.0 (Output Frequency). I get three integers separated by semicolon (500;;). So the result is untouched, still 500 Hertz instead of 50. I think it is necessary to extend the math functions by splitting (";") and then using the math functions value by value. Would you do this or should I try? C is not my preferred language ;-)

peelman:
The math operations will only happen to numbers (via the SNMP return type). The inclusion of a ; means it is returning as a string.

You might try using the [UPS-MIB OIDs](http://www.oidview.com/mibs/0/UPS-MIB.html) and see if the frequency endpoint there returns a cleaner value first.

That's just bad/lazy MIB design that defeats the purpose of SNMP.

PR #1352, opened 2015-09-15
check_http: support the concept of virtual ports

odenbach:
see issue #1026 , rebased against master.

PR #1421, opened 2016-06-04
check_http: Adding support for checking SSL-Websites listening on por…

lausser:
…ts other than 443 through Proxies

Now it is possible to run:
-I squid-ip -p 3128 -H external-webserver:8080
Before this was not possible, check_http tried to connect to squid-ip and port 8080
Now it also handles redirects coming from such external webservers properly.

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -91,6 +91,8 @@ struct timeval tv_temp;
 
 int specify_port = FALSE;
 int server_port = HTTP_PORT;
+char *virtual_server_name = NULL;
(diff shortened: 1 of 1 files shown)


odenbach:
Hi,

I had a very similar problem years ago and also created a patch:

https://github.com/monitoring-plugins/monitoring-plugins/pull/1352

BTW I also had this travis check failure.

Regards,

Christopher

lausser:
This patch is worthless, it breaks non-ssl Proxy requests.

Issue #1420, opened 2016-05-24
check_disk: show only offending paths


ip1981:
In case of warning or crtical output only paths causing that warning or critical. Output all in case of ok.

riskersen:
For this case, we have -e option available, which prints only offending paths.

I'll prepare something for the OK output

PR #1436, opened 2016-11-04
check_disk - show all disks if state is ok and option error only is used

riskersen:
This fix changes output of check_disk in case of --error-only/-e option
is used and state is ok

 - Old output: DISK OK
 - New output: DISK OK - free space: / 159731 MB (83% inode=61%);
   /dev/shm 2926 MB (100% inode=99%); /boot 58 MB (32% inode=99%);

Resolves: #1420

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -168,6 +168,7 @@ main (int argc, char **argv)
   int result = STATE_UNKNOWN;
   int disk_result = STATE_UNKNOWN;
   char *output;
+  char *ko_output;
   char *details;
   char *perf;
   char *preamble;
@@ -187,6 +188,7 @@ main (int argc, char **argv)
 
   preamble = strdup (" - free space:");
   output = strdup ("");
+  ko_output = strdup ("");
   details = strdup ("");
   perf = strdup ("");
   stat_buf = malloc(sizeof *stat_buf);
@@ -351,9 +353,6 @@ main (int argc, char **argv)
                           TRUE, 0,
                           TRUE, path->dtotal_units));
 
-      if (disk_result==STATE_OK && erronly && !verbose)
-        continue;
-
       if(disk_result && verbose >= 1) {
 	xasprintf(&flag_header, " %s [", state_text (disk_result));
       } else {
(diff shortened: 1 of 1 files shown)


waja:
This commit is broken: see https://travis-ci.org/monitoring-plugins/monitoring-plugins#L3768-L3775

sni:
I think a new Option, maybe -E (if not already taken) would be the better solution. I don't think this PR will break anything but it still changes the current behaviour which some people may like.

riskersen:
Yes, you're right, but adding another option won't make the usage easier..?

weiss:
> The tests seem to be okay.

Travis was only happy because the corresponding test wasn't executed there.  Now that @waja [fixed this][1], you can check the Travis output to see [where it's failing][2].

Apart from that, you're [accessing][3]  `flag_header` after it has been [`free()`d][4].

Last not least, the `NEWS` file hasn't been updated with this change (nor with your other changes).

I think we should either sort this out quickly just revert the change for the moment, as we declared the current master branch to be a release candidate and [told people to test it][5] *before* this was merged.

Maybe you could join the `#monitoring-plugins` channel on freenode so we can negotiate these things?  That would be nice :-)

[1]: https://github.com/monitoring-plugins/monitoring-plugins/commit/6d0c1116317ed786971c67a325d67edce37bc746
[2]: https://travis-ci.org/monitoring-plugins/monitoring-plugins/builds/177487042#L3768-L3775
[3]: https://github.com/monitoring-plugins/monitoring-plugins/blob/4430b63b0f6e1388a0f931b597fef64817ce0614/plugins/check_disk.c#L381
[4]: https://github.com/monitoring-plugins/monitoring-plugins/blob/4430b63b0f6e1388a0f931b597fef64817ce0614/plugins/check_disk.c#L369
[5]: https://www.monitoring-plugins.org/archive/devel/2016-November/010375.html

riskersen:
Of course, you're right. I'll revert the changes.

Just one small note, when was the deadline committed? I can't find it in my mails.

Sure, i can join the channel, but unfortunately not now. I'll join later.

weiss:
> Just one small note, when was the deadline committed? I can't find it in my mails.

We discussed this on IRC :-)  There's no email, sorry.  IRC is really the main communication channel of us few maintainers.

> I'll join later.

Cool, thanks.

PR #1440, opened 2016-11-07
Update .travis.yml

glensc:
sudo for `[` seemed wrong, refs 20d8799

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   # multiverse is no on trusty activated (https://github.com/travis-ci/travis-ci/issues/4979)
   - sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty multiverse" && sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty-updates multiverse"
   # /etc/hosts has IPv6 hosts (https://github.com/travis-ci/travis-ci/issues/4978)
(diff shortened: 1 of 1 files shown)


sni:
I'd say both works. Your solution is nicer because it doesn't run grep and wc with sudo. But we can just remove sudo completly, because `ip addr show` works with user permissions as well and we just need to keep the sudo before the `sed`.

sni:
And it seems like yaml does not want a string to start with [. So i'd say we leave things as they are.

ckujau:
Assuming that ```/etc/hosts``` is only modified on the TravisCI system, what about:

``ip addr show | grep -q "inet6 ::1" && sudo sed -i '/^::1/d' /etc/hosts```

glensc:
@sni you should understand how shell parser works. wc and grep are not ran under sudo, they are next command in pipe chain, if *want* them to run under sudo, you must prepend sudo to each of them

why it works is likely "ip" command does not require root access, and sudo will just include `/sbin` in path so the program is visible. solution: use abs path.

however i kept sudo, because depending on kernel hardening `ip` via `sudo` may give more information than under normal user.

glensc:
if travis does not like `[` then the line can be replaced with:

```
- test $(sudo ip addr show | grep "inet6 ::1" | wc -l) -lt "1" && sudo sed -i '/^::1/d' /etc/hosts
```

qha:
You might also consider replacing
  `grep "inet6 ::1" | wc -l`
with 
  `grep -c "inet6 ::1"`

PR #1443, opened 2016-11-13
add openssl 1.1 support

sni:
changes:
  - CRYPTO_lock detection replaced in configure.ac. We don't use that
    function anywhere, so just replace it with the suggested one from
    https://wiki.openssl.org/index.php/Library_Initialization#Autoconf
  - OPENSSL_NO_SSL2 is no longer defined while ssl2 is not included.
    Set it ourself using the suggested openssl 1.1 version check from
    https://wiki.openssl.org/index.php/1.1_API_Changes#Backward_compatibility
  - openssl 1.1 sends a sigpipe if the connection is still open when
     calling SSL_shutdown(), so move the close before the shutdown.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -493,15 +493,15 @@ if ! test x"$with_openssl" = x"no"; then
 	dnl Check for crypto lib
 	_SAVEDLIBS="$LIBS"
 	LIBS="-L${with_openssl}/lib"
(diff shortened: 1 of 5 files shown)


waja:
Manpages for 1.1.0 can be found at https://www.openssl.org/docs/man1.1.0/

PR #1444, opened 2016-11-16
check_http: add check for virtual port usage

odenbach:
Hi,

I just added a test for virtual port usage:

./check_http -H 127.0.0.1 -p 50637 -u /virtual_port

will fail, while

./check_http -H 127.0.0.1:80 -p 50637 -u /virtual_port

will succeed.

Any Comments?

Thanks,

Christopher

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -1041,6 +1041,10 @@ check_http (void)
       microsec_firstbyte = deltime (tv_temp);
       elapsed_time_firstbyte = (double)microsec_firstbyte / 1.0e6;
     }
+    while (pos = memchr(buffer, '\0', i)) {
+      /* replace nul character with a blank */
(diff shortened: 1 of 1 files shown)


waja:
Hi @odenbach,

thanks for sharing. A small side note: 

Please do not make changes to the master branch of your for but create a own branch for every fix/feature/contribution. Otherwise you/we/all will end up into a mess.

Thanks and cheers, Jan.

sni:
Also please do not change existing tests, like the "run_common_tests" unless neccessary. I think its better to add new tests aling the existing ones.
Also i added some tests last week with https://github.com/monitoring-plugins/monitoring-plugins/commit/9e02d0c3e6cc4c2cf08140dcff50fb280315bfc8. Do these tests add different checks?

PR #1445, opened 2016-11-17
check_http: Add test for virtual port usage

odenbach:
Hi,

next try with a dev branch, hope I did it right this time (I am still on my first steps with git).

The test should run for the non-SSL and SSL-server, so the natural way to implement both tests was in the run_common_tests routine. To make the test aware of the used port I had to add the virtual port to the hostname (127.0.0.1:80 instead of 127.0.0.1). This does not make any difference for the already implemented tests and works fine for my new one.

What I did:

- change number of tests
- add new test URL to the http daemons
- add two new tests to the run_common_tests routine
- add the virtual port to the hostname

If you do not like the idea of adding the virtual port for all other checks as well (which are not aware of the virtual port), I would suggest passing hostname, real port and virtual port as individual parameters and have the routine use them as needed. I can implement that if you prefer it.

Any more comments?

Thanks,

Christopher


odenbach:
What is wrong with mariadb in the Travis Environment?

waja:
> What is wrong with mariadb in the Travis Environment?

Looks like an issue on Travis-CI: https://www.google.de/#q=Unable+to+correct+problems,+you+have+held+broken+packages.&tbs=qdr:d

odenbach:
Hi,

next try to add a test for my check_http changes which got integrated.

Cheers,

Christopher

odenbach:
Hi,

could you please have a look at my latest changes? Any comments?

sni:
Could you please rebase your PR on the latest master? It's quite hard to tell what you actually changed.

odenbach:
Like this?

sni:
not really, this PR still contains all other commits since then. I meant a "rebase". Like
`git rebase master` instead of a `git merge`. You might have to force push the changes then, since you changed the history of your branch then. (Make a backup of you changes before you do that :-)
If the rebase doesn't work for some reasons, just delete the branch or create a new one from the fresh master and apply your changes there.

odenbach:
So please have a look at #1474.

sni:
great, this looks much cleaner. I will close this one.

PR #1448, opened 2016-11-21
check_smtp: Add lmtp support

waja:
Simple extension of check_smtp to support lmtp as well.


RincewindsHat:
Was duplictated (and merged) with #1715. Ops

PR #1449, opened 2016-11-21
check_http: Support http redirect

waja:
Referenced redirect of the format //www.server.com/folder would result
in check_http trying to contact http://hostname:80//www.server.com/folder
instead of http://www.server.com/folder. Referenced redirect of this format
is listed in rfc3986 ( https://tools.ietf.org/html/rfc3986 ). It should
work as expected now.

See also https://github.com/nagios-plugins/nagios-plugins/pull/148 and https://github.com/nagios-plugins/nagios-plugins/commit/d5fb1b067ffe309c0a3fd20028633e7e7cac47eb


RincewindsHat:
hm, @sni  or someone who knows http? :-)

sni (approved):
looks fine

PR #1455, opened 2016-12-19
check_dns: allow for IPv6 RDNS

prauscher:
small change to allow the use of check_dns for IPv6-RDNS. Currently only IPv4 is supported.


sni:
looks good to me.

sni:
thanks

PR #1456, opened 2016-12-21
check_dns: Accept CIDR

NicolaiSoeborg:
Improve `check_dns` by allowing the _expected address_ to be specified by CIDR notation (IPv4 only).


sni:
Great,, thank you very much. Looks good to me.

PR #1430, opened 2016-09-01
check_apt: add option to list package names

mafgh:
List names of pending package upgrades:

```
plugins/check_apt --input-file /tmp/postgres -l
APT CRITICAL: 5 packages available for upgrade (2 critical updates) postgresql-client-common postgresql-client-9.1 postgresql-common postgresql-9.1 postgresql . |available_upgrades=5;;;0 critical_updates=2;;;0
```

Input file contents:

```
Inst postgresql-client-common (134wheezy4 Debian:7.11/oldstable [all])
Inst postgresql-client-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Inst postgresql-common (134wheezy4 Debian:7.11/oldstable [all])
Inst postgresql-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Inst postgresql (9.1+134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-client-common (134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-client-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Conf postgresql-common (134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
```

PR #1427, opened 2016-08-25
check_apt: Add -l/--list option to print packages

vvidic:
Closes issue #1256

Issue #1256, opened 2014-05-13
check_apt: Please output  package names to update


wAmpIre:
Just a list of packages which cause WARN(updates)/CRIT(security updates) status as multiline output could be very helpful.

THX

PR #1458, opened 2017-01-02
check_apt: add option to list package names

mafgh:
Updated version of #1430:

List names of pending package upgrades:

```
plugins/check_apt --input-file /tmp/foo -l
APT CRITICAL: 5 packages available for upgrade (2 critical updates) postgresql-client-common postgresql-client-9.1 postgresql-common postgresql-9.1 postgresql. |available_upgrades=5;;;0 critical_updates=2;;;0
```
Input file contents:
```
Inst postgresql-client-common (134wheezy4 Debian:7.11/oldstable [all])
Inst postgresql-client-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Inst postgresql-common (134wheezy4 Debian:7.11/oldstable [all])
Inst postgresql-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Inst postgresql (9.1+134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-client-common (134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-client-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])
Conf postgresql-common (134wheezy4 Debian:7.11/oldstable [all])
Conf postgresql-9.1 (9.1.23-0+deb7u1 Debian-Security:7.0/oldstable [i386])

```

### DIFF
--- a/plugins/check_apt.c
+++ b/plugins/check_apt.c
@@ -66,7 +66,7 @@ char* construct_cmdline(upgrade_type u, const char *opts);
 /* run an apt-get update */
 int run_update(void);
 /* run an apt-get upgrade */
-int run_upgrade(int *pkgcount, int *secpkgcount);
+int run_upgrade(int *pkgcount, int *secpkgcount, char *packages, size_t len);
 /* add another clause to a regexp */
 char* add_to_regexp(char *expr, const char *next);
 
@@ -81,12 +81,16 @@ static char *do_include = NULL;  /* regexp to only include certain packages */
 static char *do_exclude = NULL;  /* regexp to only exclude certain packages */
 static char *do_critical = NULL;  /* regexp specifying critical packages */
 static char *input_filename = NULL; /* input filename for testing */
+static int do_list = 0; /* whether to list package names */
 
 /* other global variables */
 static int stderr_warning = 0;   /* if a cmd issued output on stderr */
 static int exec_warning = 0;     /* if a cmd exited non-zero */
 
 int main (int argc, char **argv) {
+	const unsigned int BUF_LEN = 1000;
+	char packages[BUF_LEN];
+
 	int result=STATE_UNKNOWN, packages_available=0, sec_count=0;
 
 	/* Parse extra opts if any */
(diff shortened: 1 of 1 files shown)


waja:
This seems to be a possible fix for #1256. An alternative might be #1427.

waja:
I would prefer #1427 over this one. But what do I know. 😄

plugins/check_apt.c
```diff
@@ -148,22 +153,23 @@ int process_arguments (int argc, char **argv) {
...
-		c = getopt_long(argc, argv, "hVvt:u::U::d::ni:e:c:", longopts, NULL);
+		c = getopt_long(argc, argv, "hVvt:u::U::d::ni:e:c:l", longopts, NULL);
 
 		if(c == -1 || c == EOF || c == 1) break;
 
 		switch(c) {
 		case 'h':
 			print_help();
-			exit(STATE_UNKNOWN);
+			exit(STATE_OK);
```
sni: Please never do that, this might lead to plugins which only print the help or version and you will never notice because they exit OK. See https://www.monitoring-plugins.org/doc/guidelines.html#AEN78

plugins/check_apt.c
```diff
@@ -148,22 +153,23 @@ int process_arguments (int argc, char **argv) {
...
 		case 'V':
 			print_revision(progname, NP_VERSION);
-			exit(STATE_UNKNOWN);
+			exit(STATE_OK);
```
sni: same here

plugins/check_apt.c
mafgh: thanks, good catch.

mafgh:
I implemented @sni's changes and rebased onto master.

Please note, meanwhile master introduced the --only-critical option with bf18dc4, however my change will still list all pending updates. I am not sure if we should filter out non security updates when --only-critical is set.

waja:
> I am not sure if we should filter out non security updates when --only-critical is set.

I think it would make sense to filter them, when this option is set. At least I think this is what users would expect.

PR #1459, opened 2017-01-07
Several fixes shellcheck complaining about

waja:
[Shellcheck ](https://www.shellcheck.net/) complains about ```plugins-scripts/*.sh```.

This fixes some of this complains:

* SC2006 (some are unfixed)
* SC2086
* SC2009
* SC2143
* SC2162
* SC2166

There are still issues open:

* SC1090
* SC2006
* SC2034
* SC2123
* SC2154

The output of shellcheck after appying this fixes is attached.

[shellcheck.txt](https://github.com/monitoring-plugins/monitoring-plugins/files/691436/shellcheck.txt)

### DIFF
--- a/plugins-scripts/check_log.sh
+++ b/plugins-scripts/check_log.sh
@@ -60,8 +60,8 @@
 
 PATH="@TRUSTED_PATH@"
 export PATH
-PROGNAME=`basename $0`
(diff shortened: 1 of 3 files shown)


waja:
> Is it really required to do something like exit "$STATE_OK" when states are only numbers

Does it hurt? :o)

PR #1460, opened 2017-01-09
check_dns: ignore CNAMEs in RDNS-mode

prauscher:
Suppose your provider uses RFC 2317 (or RFC 4183 for that matter). The output of nslookup could look like the following:
```
2.120.22.172.in-addr.arpa	canonical name = 2.0/26.120.22.172.in-addr.arpa.
2.0/26.120.22.172.in-addr.arpa	name = rajesh.intern.prauscher.de.
```
Without this filter, check_dns would check the string against "2.0/26.120.22.172.in-addr.arpa., rajesh.intern.prauscher.de." which will fail for obvious reasons.
In forward DNS, this is achieved in line 165, as nslookup will print the address separate from the CNAME-resolving.


prauscher:
Looks like the CI-Check failed, but this is attributed to clang C at check_ping, which i did not change - probably this is just a glitch?
https://travis-ci.org/monitoring-plugins/monitoring-plugins/jobs/190322786

PR #1461, opened 2017-01-26
Enable check_file_age to test for maximum file size

philipowen:
Useful when a regular process overwrites a file showing the current status. In my use case, the error output from a regular rsync so I can check there were no errors

### DIFF
--- a/plugins-scripts/check_file_age.pl
+++ b/plugins-scripts/check_file_age.pl
@@ -57,7 +57,7 @@ GetOptions(
 	"C=s" => \$opt_C, "critical-size=s" => \$opt_C);
 
 if ($opt_V) {
-	print_revision($PROGNAME, '2.2.22.g0d73b');
+	print_revision($PROGNAME, '@NP_VERSION@');
 	exit $ERRORS{'UNKNOWN'};
 }
 
@@ -152,9 +152,9 @@ sub print_help () {
 	print "  If any of the warning and critical arguments are in range syntax (not just bare numbers)\n";
 	print "  then all warning and critical arguments will be interpreted as ranges.\n";
 	print "  To use range processing the perl module Monitoring::Plugin must be installed\n";
-	print "  For range syntax see https://nagios-plugins.org/doc/guidelines.html#THRESHOLDFORMAT\n";
+	print "  For range syntax see https://www.monitoring-plugins.org/doc/guidelines.html#THRESHOLDFORMAT\n";
 	print "  It is strongly recommended when using range syntax that all four of -w, -W, -c and -C are specified\n";
-	print "  otherwise it is unlikely that the size test will be doint wat is desired\n";
+	print "  otherwise it is unlikely that the size test will be doing what is desired\n";
 	print "\n";
 	support();
 }
--- a/plugins-scripts/check_file_age.pl
+++ b/plugins-scripts/check_file_age.pl
@@ -43,8 +43,6 @@ $ENV{'ENV'}='';
 
 $opt_w = 240;
 $opt_c = 600;
-$opt_W = 0;
-$opt_C = 0;
 $opt_f = "";
 
 Getopt::Long::Configure('bundling');
@@ -53,13 +51,13 @@ GetOptions(
 	"h"   => \$opt_h, "help"	=> \$opt_h,
 	"i"   => \$opt_i, "ignore-missing"	=> \$opt_i,
 	"f=s" => \$opt_f, "file"	=> \$opt_f,
-	"w=f" => \$opt_w, "warning-age=f" => \$opt_w,
-	"W=f" => \$opt_W, "warning-size=f" => \$opt_W,
-	"c=f" => \$opt_c, "critical-age=f" => \$opt_c,
-	"C=f" => \$opt_C, "critical-size=f" => \$opt_C);
+	"w=s" => \$opt_w, "warning-age=s" => \$opt_w,
+	"W=s" => \$opt_W, "warning-size=s" => \$opt_W,
+	"c=s" => \$opt_c, "critical-age=s" => \$opt_c,
+	"C=s" => \$opt_C, "critical-size=s" => \$opt_C);
 
 if ($opt_V) {
-	print_revision($PROGNAME, '@NP_VERSION@');
+	print_revision($PROGNAME, '2.2.22.g0d73b');
 	exit $ERRORS{'UNKNOWN'};
 }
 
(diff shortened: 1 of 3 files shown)


sni:
Thanks and bonus points for adding tests :-) But wouldn't it be better to just support ranges in the current thresholds.

philipowen:
Now you mention it, yes it probably would. Perhaps I should have opened an issue to discuss it beforehand! I'll have a look at one of the other plugins that uses ranges to see how it is currently implemented.

sni:
I just peeked in the script and it would be super easy if the check would use the Monitoring::Plugin perl module with its threshold functions. But it doesn't. And right now we don't use the Monitoring::Plugin module in any of our perl plugins so i guess we cannot assume that the module is installed at all.

So to keep this plugin backwards compatible, we could check the arguments and if its not a number (assuming its a threshold then) require the M::P module and use that threshold functions

waja:
> So to keep this plugin backwards compatible, we could check the arguments and if its not a number (assuming its a threshold then) require the M::P module and use that threshold functions

Just in case we make use of M::P, we need to inform our users that dependencies has changed in a appropriate way.

philipowen:
I'm not a seasoned Perl guru but I'll try and put something together as suggested with a conditional dependency and post it for review.

sni:
the idea is to make the "require" optional if somebody uses the new threshold syntax. In that case we would not break existing checks.

philipowen:
I was looking at using the following logic:
if (any_arg_contains_non_digit) {
  require Monitoring::Plugin::Range;
  # use range processing
}
else {
  # use existing logic
}

The problem with that is that people could be using the range shorthand syntax e.g. '10' to mean between zero and 10 to be valid. That is fine for the age check, but the size check is looking to alert at less than a given threshold. So if the argument is '10' how can I tell whether this is existing logic (>=11 is okay) or range shorthand (0..10 is okay)?

Do we need a new argument to enable range check interpretation?

sni:
right, thats exactly how i meant it. I don't see the need for a new argument. If somebody uses a simple number, we fallback to the old logic anyway?

philipowen:
I was (perhaps) over-complicating it and concerned about people wanting to use range processing but expecting to be able to use only the single number shorthand for 0..x but I'll put something in the usage output that explicitly says what the behavior is. e.g. in the case of -w 10 -c 50 we get existing behavior and for -w 10 and -c 50:60 the 10 will be processed as a range and hence is 0..10

philipowen:
I have an implementation as discussed. I wanted to sanity check something before submitting it. It is a subtlety created because of the fact that the file size test behavior is a "less than" test in the existing code, and default to zero. For example, if I have a file foo of age 10 seconds: 
check_file_age -f foo -w 300 -c 600
FILE_AGE OK: foo is 10 seconds old and 30 bytes | age=10s;300;600 size=30B;0;0;0
But if I start using ranges for the age, the invisible default value of "0" for the size gets interpreted as a range, and we are outside the range 0:0:
FILE_AGE CRITICAL: foo is 10 seconds old and 30 bytes | age=10s;0:300;0:600 size=30B;0;0;0
At the moment I have put some text in the usage notes to strongly suggest that when using ranges, all 4 arguments are specified. Is that sufficient? Or should we turn off the size checking if no arguments for it are specified (regardless of range processing)?

philipowen:
Implementation pushed for review. I decided my previous question was really "should I tread more heavily on the code to make the default behavior more sensible?" to which the answer was clearly "yes".

philipowen:
Please can I get a review of this PR?

Thanks.

plugins-scripts/check_file_age.pl
```diff
@@ -53,13 +51,13 @@
...
-	"c=f" => \$opt_c, "critical-age=f" => \$opt_c,
-	"C=f" => \$opt_C, "critical-size=f" => \$opt_C);
+	"w=s" => \$opt_w, "warning-age=s" => \$opt_w,
+	"W=s" => \$opt_W, "warning-size=s" => \$opt_W,
+	"c=s" => \$opt_c, "critical-age=s" => \$opt_c,
+	"C=s" => \$opt_C, "critical-size=s" => \$opt_C);
 
 if ($opt_V) {
-	print_revision($PROGNAME, '@NP_VERSION@');
+	print_revision($PROGNAME, '2.2.22.g0d73b');
```
sni: Please don't change this.

plugins-scripts/check_file_age.pl
```diff
@@ -120,7 +147,15 @@ ()
 	print "\n";
 	print "  -i | --ignore-missing :  return OK if the file does not exist\n";
 	print "  <secs>  File must be no more than this many seconds old (default: warn 240 secs, crit 600)\n";
-	print "  <size>  File must be at least this many bytes long (default: crit 0 bytes)\n";
+	print "  <size>  File must be at least this many bytes long (default: crit 0 bytes)\n\n";
+	print "  Both <secs> and <size> can specify a range using the standard plugin syntax\n";
+	print "  If any of the warning and critical arguments are in range syntax (not just bare numbers)\n";
+	print "  then all warning and critical arguments will be interpreted as ranges.\n";
+	print "  To use range processing the perl module Monitoring::Plugin must be installed\n";
+	print "  For range syntax see https://nagios-plugins.org/doc/guidelines.html#THRESHOLDFORMAT\n";
```
sni: please link to: https://www.monitoring-plugins.org/doc/guidelines.html#THRESHOLDFORMAT

plugins-scripts/check_file_age.pl
```diff
@@ -120,7 +147,15 @@ ()
...
+	print "  It is strongly recommended when using range syntax that all four of -w, -W, -c and -C are specified\n";
+	print "  otherwise it is unlikely that the size test will be doint wat is desired\n";
```
sni: typo: "doing what"

plugins-scripts/t/check_file_age.t
```diff
@@ -48,22 +48,44 @@ cmp_ok( $result->return_code, '==', 1, "Warning for file over 1 second old" );
 like  ( $result->output, $warningOutput, "Output for warning correct" );
 
 $result = NPTest->testCmd(
+	"./check_file_age -f $temp_file -w 0:1"
```
sni: The tests will fail if Monitoring::Plugins::Range is not installed. Could you add a `apt-get install libmonitoring-plugin-perl` to the .travis.yml file?

philipowen:
Thanks @sni  

All changes made. Apologies for the schoolboy mistakes I should have looked harder at the diff.

I notice the Travis build is still failing. It seems like the package libmonitoring-plugin-perl is only available in Ubuntu yakkety. What's the best approach here to get it installed, cpan?

sni:
We could make the tests optional and skip over the range tests if the module is not available. Or maybe there is a backport anywhere or we use cpan to install the missing module. Usually i would avoid installing perl modules by cpan, but its only a test virtual machine...

sni:
@waja was so kind to add libmonitoring-plugin-perl to his trusty backports, so the travis tests are fine now. So no further change requests from my side.

philipowen:
Fantastic. Thanks to you both.

sni:
Merged, thanks.

Issue #585, opened 2013-09-24
Wrong compile Options for openSSL [sf#1291115]


monitoring-user:
Submitted by fichi on 2005-09-14 17:45:34

I tried to compile nagios-plugins-1.4.1.tar.gz on a
debian 3.1r0a system.
I installed openssl-0.9.8 successfully into
/usr/local/ssl which is default for this release.
I run ./configure --with-openssl=/usr/local/ssl

The make aborted with the following error message.

gcc  -g -O2  -L. -R/usr/local/ssl/lib
-L/usr/local/ssl/lib -o check_http  check_http.o
netutils.o utils.o ../lib/libnagiosplug.a
../lib/libcoreutils.a  -lnsl -lresolv -lssl -lcrypto
-I/usr/local/ssl/include
gcc: unrecognized option `-R/usr/local/ssl/lib'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x45):
In function`dlfcn_load':
: undefined reference to `dlopen'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0xc4):
In function`dlfcn_load':
: undefined reference to `dlclose'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x102):
In function`dlfcn_load':
: undefined reference to `dlerror'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x1ad):
In function`dlfcn_bind_var':
: undefined reference to `dlsym'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x1f2):
In function`dlfcn_bind_var':
: undefined reference to `dlerror'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x2ed):
In function`dlfcn_bind_func':
: undefined reference to `dlsym'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x337):
In function`dlfcn_bind_func':
: undefined reference to `dlerror'
/usr/local/ssl/lib/libcrypto.a(dso_dlfcn.o)(.text+0x68a):
In function`dlfcn_unload':
: undefined reference to `dlclose'
collect2: ld returned 1 exit status
make[2]: *** [check_http] Error 1
make[2]: Leaving directory
`/root/src/nagios-plugins-1.4.1/plugins'
[...]

Issue #584, opened 2013-09-24
check_dig needs better check in if statement [sf#1155565]


monitoring-user:
Submitted by mishikal on 2005-03-03 04:38:16

This patch helps compensate for different DNS returns:

--- nagios-plugins-1.3.1/plugins/check_dig.c.orig  
 2003-01-13 04:15:15.000001000 -0800
+++ nagios-plugins-1.3.1/plugins/check_dig.c  
2003-12-12 07:48:30.000001000 -0800
@@ -76,11 +76,12 @@
                printf ("Could not open stderr for
%s\n", command_line);

```
    while (fgets (input_buffer, MAX_INPUT_BUFFER -
```

1, child_process)) {

```
            /* the server is responding, we just
```

got the host name... */
-               if (strstr (input_buffer, ";; ANSWER
  SECTION:")) {
-               if (strstr (input_buffer, ";; ANSWER
  SECTION:") ||
- ```
         strstr (input_buffer, ";; ANSWERS:")) {
  
                  /* get the host address */
                  if (!fgets (input_buffer,
  ```
  
  MAX_INPUT_BUFFER - 1, child_process))
                              break;

PR #1462, opened 2017-01-27
Fixed logic in 'plugins/check_nt.c' UPTIME at warning/critical return_code.

theAdmin10:
Reversed the logic in the '`plugins/check_nt.c`' `UPTIME` check to behave properly.

### DIFF
--- a/plugins/check_nt.c
+++ b/plugins/check_nt.c
@@ -224,9 +224,9 @@ int main(int argc, char **argv){
 
 			xasprintf(&output_message,_("System Uptime - %u day(s) %u hour(s) %u minute(s) |uptime=%lu"),updays, uphours, upminutes, uptime);
 
-			if (check_critical_value==TRUE && uptime <= critical_value)
+			if (check_critical_value==TRUE && uptime >= critical_value)
 				return_code=STATE_CRITICAL;
-			else if (check_warning_value==TRUE && uptime <= warning_value)
+			else if (check_warning_value==TRUE && uptime >= warning_value)
 				return_code=STATE_WARNING;
 			else
 				return_code=STATE_OK;


technopak:
Sorry.  Hit the wrong button, and closed it.  Re-opened it.  At least it will get checked twice by Travis. :-P

technopak:
It passed the Travis CI build in build [#584](https://travis-ci.org/monitoring-plugins/monitoring-plugins/builds/195737953)., but then after I closed the pull request and opened it again, the second build failed.  But it looks that the failure in [#585.2](https://travis-ci.org/monitoring-plugins/monitoring-plugins/jobs/195747714). is related to check_https, and not check_nt.
Is there a way to trigger another Travis CI build test, with out closing and re-opening the Pull request?

sni:
i restarted the ci test. they are not very reliable unfortunatly.

sni:
You could argue that this is the expected behaviour. People maybe might want to detect reboots.

riskersen:
I would say both is useful. We're doing that for some of our not so reliable appliances..detecting reboots can be helpful to detect watchdog triggers and so on.

So with that change you will change the default behavior, people may get confused.
To avoid that, I would propose to add your change as an addition, e.g. "min uptime":"max uptime" as parameter.

technopak:
@riskersen : I have some other thoughts:  

- I understand not wanting to change the "default" behaviour from what it's been, however when I look at some of the other checks, i.e. USEDDISKSPACE, and MEMUSE, the "normal" logic behaviour is as it is in the proposed patch.  In another check i.e FILEAGE the logic is "normal" as above, but reversed if the critical_value is less than the warning_value.  I could make this a similar change, in that the logic would reverse, based on whether or not the critical_value was greater than the warning_value. However, the "inverse thresholds" only works if you have specified, both warning and critical values are set, and critical is < warning.

- Another, option is to add another switch/argument (i.e. -r or -i ) for "reverse logic" or "inverse logic".  This could provide the ability to reverse or invert the logic on the UPTIME check and, if wanted on other checks as well.  (NOTE: from the above point, as in the FILEAGE check, the "inverse threshold" only works, if both warning and critical values are set, and critical is < warning.)

Question: when you or others are utilizing the "default" behaviour, are you using both critical and warning values, or just critical, or just warning?

sni:
Adding a reverse option doesn't make sense, you could just use the `negate` plugin.

PR #1463, opened 2017-01-27
check_ircd.pl: use pack_sockaddr_in rather than hand-rolled

sthen:
On some OS, sockaddr structs include a length field. Perl's pack_sockaddr_in
takes this into account; the hand-rolled "pack('S n a4 x8'..." doesn't do so,
resulting in connection failures.

Problem reported on OpenBSD by Alf Schlichting


RincewindsHat (approved):
looks useful and is definitely better readable

RincewindsHat:
Integrated with #1932 . Thank you very much.

Issue #1466, opened 2017-02-08
check_snmp.c - UOM in perfdata not set correctly when using warn/crit


riskersen:
With current version, the UOM isn't correctly in perfdata output when using warn/crit values:

`SNMP OK - 536 | iso.3.6.1.4.1.2620.1.1.25.3.0=536;15000;20000c;15000;20000;`

So, basically UOM / type was added at the very end of the perfdata strncat processing.

The problem is on line 591, this one has to be moved above the warn/crit strncat processing.

riskersen:
I already prepared a fix for that, which is of course quite simple.

riskersen:
Will be solved by #1465

PR #1467, opened 2017-02-08
check_snmp - append UOM correctly

riskersen:
When using warn and crit thresholds, UOM value was added at the
  end of the string, which was warn or crit threshold, depending
  what would be the last value in the row.
  This fix will append UOM correctly to perfdata value and solves #1466


riskersen:
This is madness..this issue is there since years and just yesterday there was PR #1465 which does the same.

PR #1468, opened 2017-02-24
check_snmp: add multiplier option for simple math on numeric values

mitch-geht-ab:
I've added a simple multiplier to scale numeric values.
The difference to #1417 is, that only one multiplier is needed. If division is needed, multiply with 0.x and for negation multiply with -x.
It only works when thresholds are set.

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -118,6 +118,7 @@ size_t oids_size = 0;
 char *label;
 char *units;
 char *port;
+double value_multiplier = 1;
 char *snmpcmd;
 char string_value[MAX_INPUT_BUFFER] = "";
 int  invert_search=0;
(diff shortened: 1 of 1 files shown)


mitch-geht-ab:
Hmmm... I don't know why Travis Build Job 590.2 failed on check_ping. I haven't changed anything on this check. Curiously both Jobs are successful for my check_snmp :)

swapper82:
This function would be very nice to get, i have tried the mitch-geht-ab version (compiled) and it works but there is one problem i can see. When scaling values that are below 0 (like 0.4) the plugin shows 0 (omits the decimals) and i cant find a parameter to set number of decimals to show.
Would also be good if it where not an requirement to have a warning or critical value set (if a service is only for performance data)

RincewindsHat:
Replaced by #1813

PR #1476, opened 2017-03-29
Allow user to specify HTTP method after proxy CONNECT

### LINKED ISSUE #1475: check_http Using CONNECT proxy - http_method GET is hardcoded
When I'm using `-j CONNECT` to make a connection via an HTTP CONNECT proxy, the request method itself is hardcoded and can't be changed.

E.g. when you try to POST something against a service.

Problem can be found here: https://github.com/monitoring-plugins/monitoring-plugins/blob/dbb92c347e8a19b3fbe33a9160776b7a16ea588b/plugins/check_http.c#L953

``` C
asprintf (&buf, "%s %s %s\r\n%s\r\n", "GET", server_url, host_name ? "HTTP/1.1" : "HTTP/1.0", user_agent);
```

lazyfrosch:
Simple format, avoids refactoring the CONNECT feature.

fixes #1475

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ install:
   - sudo apt-get install -qq --no-install-recommends autoconf automake
   - sudo apt-get install -qq --no-install-recommends faketime
   - sudo apt-get install -qq --no-install-recommends libmonitoring-plugin-perl
+  - sudo apt-get install -qq --no-install-recommends squid3
   # Trusty related dependencies (not yet provided)
   - test "$(dpkg -l | grep -E "mysql-(client|server)-[0-9].[0-9]" | grep -c ^ii)" -gt 0 || sudo apt-get install -qq --no-install-recommends mariadb-client mariadb-server
   # enable ssl apache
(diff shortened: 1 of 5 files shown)


lazyfrosch:
Tested against Charles Proxy

sni:
We should get the check_curl ready as soon as possible: https://github.com/monitoring-plugins/monitoring-plugins/tree/feature_check_curl
This should fix most of the issues with check_http...

lazyfrosch:
+1 for `check_curl`

Rebased my commit, this is a simple change, please review+merge @sni @waja @weiss :wink:

weiss:
Why do you [remove][1] the `strdup(optarg)` call and [modify][2] `optarg` directly if it contains a colon?

[1]: https://github.com/monitoring-plugins/monitoring-plugins/pull/1476/files#diff-458353164a55a2e140dfc531b6ac2c38L448
[2]: https://github.com/monitoring-plugins/monitoring-plugins/pull/1476/files#diff-458353164a55a2e140dfc531b6ac2c38R451

lazyfrosch:
I adopted an even older patch, and this uses a simple in-memory split of the string.

Would you prefer `strtok()`?

weiss:
I'm fine with in-memory splitting of the string (the way you do it), but without your patch, the string was a copy of `optarg`, and now you're operating on `optarg` directly.  Just keep the `http_method = strdup (optarg);` line and then split the `http_method` copy.

lazyfrosch:
Updated as suggested.

I also added tests with a local Squid proxy.

PR #1481, opened 2017-04-16
New general output functions for checks

sritd:
Dear Monitoring Team,

I had a look at your checks and they are working great. I will introduce a new software, which the monitoring plugins will be one part of it. I have written some test cases for my own software and got some troubles. I have implemented my interpretation of the result to the general output specification: SERVICE STATUS: First line of output|First part of performance data|Second part of performance data
or
SERVICE STATUS: First line of output|First part of performance data|Second part of performance data
Third part of performance data|Fourth part of performance data ...

Some of the checks I have tested, does not take care about this behaviour. So I checked the code and tried to find the problem. I was successful and had a great look into the code from you.

As a general solution for this problem, I have done the following steps:
- Introduced general output functions for this use case
 - print_singleline (int service_state, const char * msg, ...)
   Will print the output and returns to caller
 - print_singleline_return (int service_state, const char * msg, ...)
   Will print the output, returns the service_state as a return value and returns to caller
 - print_singleline_exit (int service_state, const char * msg, ...)
   Will print the output and exits program with service_state
- Declaration and implementation is done in utils.h and utils.c
- Changed sslutils.c and netutils.c to use new functions
- Removed lines with additional debug output, which will break the output format in sslutils.c and netutils.c
- Changed check_ssh.c to use new functions
- Changed check_ssh.c to use new functions

In the same step I applied code styles to my changed files
- Replaced ugly tabs with spaces
- Reformatted to have the same look and feel

My used settings for coding styles:
 - set encoding=utf-8
 - set tabstop=8
 - set softtabstop=4
 - set shiftwidth=4
 - set expandtab

If you think, that this is a great work and I should continue with applying my changes to the rest of the checks, let me know. I will test my functions with some more checks and will commit my changes to my fork. I can create further pull requests, if you are interested.

Best regards
Sven

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -384,7 +384,7 @@ main (int argc, char **argv)
     xasprintf (&output, "%s%s", output, details);
 
 
-  printf ("DISK %s%s%s|%s\n", state_text (result), (erronly && result==STATE_OK) ? "" : preamble, output, perf);
+  printf ("DISK %s%s%s%s\n", state_text (result), (erronly && result==STATE_OK) ? "" : preamble, output, perf);
   return result;
 }
 
--- a/plugins/check_ssh.c
+++ b/plugins/check_ssh.c
@@ -244,13 +244,13 @@ int ssh_connect (char *haddr, int hport, char *remote_version, char *remote_prot
 
         if (remote_version && strcmp(remote_version, ssh_server)) {
             close(sd);
-            print_singleline_exit (STATE_CRITICAL, _("%s (protocol %s) version mismatch, expected '%s'\n"),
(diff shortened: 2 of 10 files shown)


sritd:
Forget this pull request. I will create another one, with smaller changes and fixes to get the tests running.

sni:
thanks for your effort, please separate style changes from functional changes. This makes it a lot easier to review the pull request.

sritd:
Thank you for your reply. That was one reason why I have cancelled my pull request. Also I did not know, how your tests are working. I do not want to break them in total. So I just created a branch, were I will only introduce this function. I will only implement this function and do not take care about the perfdata. This will be done in a later step.

PR #1483, opened 2017-04-18
New general output functions for checks

sven-ruess:
Dear Monitoring Team,

I had a look at your checks and they are working great. I will introduce a new software, which the monitoring plugins will be one part of it. I have written some test cases for my own software and got some troubles. I have implemented my interpretation of the result to the general output specification: SERVICE STATUS: First line of output|First part of performance data|Second part of performance data
or
SERVICE STATUS: First line of output|First part of performance data|Second part of performance data
Third part of performance data|Fourth part of performance data ...

Some of the checks I have tested, does not take care about this behaviour. So I checked the code and tried to find the problem. I was successful and had a great look into the code from you.

As a general solution for this problem, I have done the following steps:
- Introduced general output functions for this use case
- print_singleline (int service_state, const char * msg, ...)
  Will print the output and returns to caller
- print_singleline_return (int service_state, const char * msg, ...)
  Will print the output, returns the service_state as a return value and returns to caller
- print_singleline_exit (int service_state, const char * msg, ...)
  Will print the output and exits program with service_state
- Declaration and implementation is done in utils.h and utils.c

The first step is to get this new function accepted by you.

After that, my next steps will be the following:
- Adapt sshutils.c and netutils.c to use this new functions
- Adapt some tests, to take care about the new output results
- Step by step change the checks to use this new functions and replace the usage* functions
- Take care about the tests, that they will be successfully tested

Best regards
Sven

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -166,7 +166,7 @@ main (int argc, char **argv)
   argv=np_extra_opts (&argc, argv, progname);
 
   if (process_arguments (argc, argv) == ERROR)
-    usage4 (_("Could not parse arguments"));
+    print_singleline_exit (STATE_UNKNOWN, _("Could not parse arguments"));
 
   if (display_html == TRUE)
     printf ("<A HREF=\"%s://%s:%d%s\" target=\"_blank\">",
@@ -188,7 +188,7 @@ test_file (char *path)
 {
   if (access(path, R_OK) == 0)
     return;
-  usage2 (_("file does not exist or is not readable"), path);
+  print_singleline_exit (STATE_UNKNOWN, _("File does not exist or is not readable %s"), path);
 }
 
 /* process command-line arguments */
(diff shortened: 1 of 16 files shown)


sritd:
I do not know, why the tests for ping failed. I did not modify anything, which will be related to this issue. Perhaps you can trigger a restart for this test. I should be finished successful.

sni:
I triggered a recheck of the travis build.

sritd:
Thank you. Now the tests are all successfully finished. Now you can have a look on the code. On the weekend I will adapt netutils and sslutils to use this new functions. For that I will create new branches from this current branch.

sritd:
It looks like, that your tests have some timing issues. I only optimized my code a little bit and the tests fail again. Can you give me a short introduction, how I can get your tests running locally? I followed the steps of developer and user guide but most of the tests are failing. Do I have to setup a special environment or services?

RincewindsHat:
this might be a little late, but it sounds interesting and I might take a look at that in the future.

PR #1485, opened 2017-04-20
plugins: check_dig: Fix spelling

kbabioch:
This fixes a minor spelling issue. The localization files need to be adjusted accordingly, but I couldn't find any documentation on how to update the pot file and didn't want to do this by hand.


sni:
merged, thank you

PR #1486, opened 2017-04-26
check_pgsql: Add extra output

philipowen:
This adds the ability to display extra output resulting from the execution of a custom query. When a second column is present in the resultset that column will be appended to the plugin output. This can be used to give greater information to the user. For example, if the check was looking for all overdue orders, the second column could be a list of their IDs so that recipients of the alert don't have to dig into the database to find that information.


waja:
> /rebase

too bad. :)

waja:
doesn't look so invasive ... maybe we give it a try?

PR #1491, opened 2017-05-31
check_swap: repaired "-n" behaviour

odenbach:
If run with "-n ok" a host which ran completely out of swap space would return "ok" which is
not desired. It should only return "ok" if there is no swap space configured at all.

### DIFF
--- a/plugins/check_swap.c
+++ b/plugins/check_swap.c
@@ -51,7 +51,7 @@ const char *email = "devel@monitoring-plugins.org";
 # define SWAP_CONVERSION 1
 #endif
 
-int check_swap (int usp, float free_swap_mb);
+int check_swap (int usp, float free_swap_mb, float total_swap_mb);
 int process_arguments (int argc, char **argv);
 int validate_arguments (void);
 void print_usage (void);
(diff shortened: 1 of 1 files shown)


waja:
Hi @odenbach,

a test for this issue would be cool.

odenbach:
The problem is that check_swap reads /proc/meminfo. This path is defined in config.h. Is there any way to tweak this path in the testsuite?

odenbach:
Or do you want a setup how to reproduce the bug? That's easy: Just edit config.h and replace the path of /proc/meminfo with something writable (e.g. /tmp/proc/meminfo). Then copy /proc/meminfo to this new path and edit the values for SwapTotal and SwapFree. Then compile check_swap and run with "-n ok".

Expected behaviour: OK for SwapTotal == 0, warning/critical for SwapTotal > 0 and SwapFree == 0

odenbach:
Should I provide anything else?

odenbach:
Any chance to get this upstream?

sni:
looks fine, thanks. And sorry for the delay.

PR #1493, opened 2017-06-07
Fix regexp for nullmailer "mailq" output

darksoul42:
As it currently is, the regular expression does not match mailq command
output on a Debian Jessie setup.

Three erroneous behaviors fixed :
- Meaningful lines do not end with the "bytes" word
- There might be one or more space before the byte count, not 2
- Time match was completely broken, it only accounted for 0-29 minutes
and 0-29 seconds.


darksoul42:
Sorry for the duplicate pull request, please ignore PR #1492, I tried pushing it from an old fork of this repository.

RincewindsHat (approved):
Works for me :-)

PR #1494, opened 2017-06-10
check_disk: return UNKNOWN instead of CRITICAL where…

Al2Klimov:
… a disk's usage isn't available

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -220,7 +220,7 @@ main (int argc, char **argv)
 
   while (temp_list) {
     if (! temp_list->best_match) {
-      die (STATE_CRITICAL, _("DISK %s: %s not found\n"), _("CRITICAL"), temp_list->name);
+      die (STATE_UNKNOWN, _("DISK %s: %s not found\n"), _("UNKNOWN"), temp_list->name);
     }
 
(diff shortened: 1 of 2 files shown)


Al2Klimov:
* [x] unit tests
* [x] master rebase
* [x] actual email

RincewindsHat (approved):
Looks sane.

RincewindsHat:
I think this behaviour would be ok, but there might be different opinions on that topic.
Someone might be surprised by finding that a suddenly vanished file system is not a reason for `CRITICAL`. Maybe add a `--ignore-missing` ?

sni:
i'd say you might argue about this. If we merge this, someone will come soon and want to change it back. For both variants there are valid scenarios. So i'd vote for a switch to change the behaviour but keep the default as is.

KriSchu:
Hi, I don't like the idea to return UNKNOWN when the filesystem is missing without an option to have a critical alert.
There are definitely scenarios where one wants an alert for missing filesystems, because they actually should be mounted on the server and a missing FS is a problem state.

I hope the PR #1801 is a feasible solution for the requirement not to get alerts for unavailable FS.
But please, don't merge this code change, as it would be a breaking change for us.

PR #1495, opened 2017-06-12
check_mailq: add support for opensmtpd

Subito:
Tested on FreeBSD, should be OS-independent, though.


RincewindsHat:
hm, tried to test that, but opensmtpd is called `smtpd` on debian. Not sure how to handle that.
@weiss, @waja : Any ideas?

waja:
> hm, tried to test that, but opensmtpd is called `smtpd` on debian. Not sure how to handle that.

What's the question about the "name" of opensmtpd in debian? You need a running opensmtpd for tesing?

RincewindsHat:
Sorry, that was in its shortness rather confusing. The binary itself is `smtp` not `opensmtp`. So, this is probably not OS independent. What would be the best way to circumvent that?

waja:
looking into the [documentation](https://www.opensmtpd.org/manual.html), it indicates that the binary is `smptd`. In Debian for example there is only one software with `smtpd` in the search path, opensmtpd. See https://packages.debian.org/search?mode=path&suite=bullseye&section=all&arch=any&searchon=contents&keywords=bin%2Fsmtpd

Searching on the net around the binary seems even there `smtpd`:

- https://calomel.org/opensmtpd.html
- https://freebsd.pkgs.org/13/freebsd-aarch64/opensmtpd-6.8.0,1.pkg.html

So maybe the option should be better "smtpd" instead of "opensmtpd"? It just sounds a bit to generic, that @Subito used "opensmtpd" as option for `-M (--mailserver)` but didn't had in mind, that this is autodetected if this option is not used?

RincewindsHat:
hm, I should have taken a look at FreeBSD too then. I guess, with the change from `opensmtpd` to `smtpd` we should be fine, then.

waja:
hopefully. I think we should somewhere (prominent) document, that `smtpd` stands explicitly for OpenSMTPD.

RincewindsHat:
replaced by #2330

PR #1496, opened 2017-06-19
Added option for null zero length string exit codes

FracKenA:
When using a large distributed network with the same group of checks used against a large number of devices, occasionally there are missing cards in a few devices that are present in other devices. Rather than having a large number of unknown results, disable active checking on those large number of result or having to create a unique check configuration for those devices.

This option allows you to select an OK, WARNING, CRITICAL or UNKNOWN status while still retaining the default behavior when not present. This also allows a for the check to immediately start checks as intended should the hardware be added that the check is looking for.

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -113,7 +113,7 @@ char *authproto = NULL;
 char *privproto = NULL;
 char *authpasswd = NULL;
 char *privpasswd = NULL;
-int nulloid = 3;
+int nulloid = STATE_UNKNOWN;
 char **oids = NULL;
 size_t oids_size = 0;
 char *label;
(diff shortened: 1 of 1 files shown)


KenD-OP5:
Submitting for merge into the master branch.

KenD-OP5:
any chance of accomplishing this merge at this time?

plugins/check_snmp.c
```diff
@@ -113,6 +113,7 @@ char *authproto = NULL;
 char *privproto = NULL;
 char *authpasswd = NULL;
 char *privpasswd = NULL;
+int nulloid = 3;
```
sni: i would set this to STATE_UNKNOWN to make it more clear

PR #1500, opened 2017-07-19
Use size_t instead of int when calling sysctl(3).

grayed:
Otherwise, it writes sizeof(size_t) bytes to &oldlen, smashing the stack.

Found on OpenBSD, but applies to all BSDs (oldlen sysctl(3) argument is size_t everywhere).


grayed:
ping

grayed:
pong

grayed:
fhtagn?

sni:
Thanks for you patience.

PR #1503, opened 2017-08-18
Add support to Jitter, MOS and Score

dirtyren:
Add support to monitor Jitter, MOS and Score.


dirtyren:
Better later than never... ;)

PR #1505, opened 2017-09-04
Output plugin's stderr to stderr

m-kraus:
Output plugin's stderr to stderr and do  not exit with WARNING in that case


sni:
Looks good to me as well.

PR #1506, opened 2017-09-06
Adding packages-warning option to check_apt plugin

sanchezfauste:
This option specifies the minumum number of packages available for upgrade to return WARNING status. Default is 1 package.


sni:
thanks, merged

PR #1507, opened 2017-09-07
Adding print top consuming processes option to check_load

sanchezfauste:
```
 -W, --print-top-warning
    Print top consuming processes on WARNING status
 -C, --print-top-critical
    Print top consuming processes on CRITICAL status
 -n, --procs-to-show=NUMBER_OF_PROCS
    Number of processes to show when printing top consuming
    processes. Not useful without -W or -C. Default value is 5
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1016,6 +1016,10 @@ if test -n "$ac_cv_ps_varlist" ; then
 		AC_DEFINE(PS_USES_PROCETIME,"yes",
 		          [Whether the ps utility uses the "procetime" field])
 	fi
+	if echo "$ac_cv_ps_varlist" | grep "procpcpu" >/dev/null; then
+		AC_DEFINE(PS_USES_PROCPCPU,"yes",
+		          [Whether the ps utility uses the "procpcpu" field])
+	fi
 fi
 
 AC_PATH_PROG(PATH_TO_PING,ping)
--- a/plugins/check_load.c
+++ b/plugins/check_load.c
@@ -33,6 +33,7 @@ const char *copyright = "1999-2007";
 const char *email = "devel@monitoring-plugins.org";
 
 #include "common.h"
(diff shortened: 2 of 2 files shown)


sni:
I like the idea but i would make it less complicated. I would just add one additional option `-n` and print the top most processes if that number is greater than zero. I don't its neccessary to set different options for warnings and criticals.

sanchezfauste:
@sni I agree with you, so I change it to:

```
 -n, --procs-to-show=NUMBER_OF_PROCS
    Number of processes to show when printing the top consuming processes.
    NUMBER_OF_PROCS=0 disables this feature. Default value is 0
```

sni:
I am also a bit concered about the hard-coded `/bin/ps`. We have a `PS_COMMAND` constant which can be customized by the `./configure` script. And i assume this is for a reason. The `--sort=-pcpu` is probably not very portable across the various unix/linux systems.

```
%> ./configure --help 2>&1 | grep ps-
  --with-ps-command=PATH  Verbatim command to execute for ps
  --with-ps-format=FORMAT Format string for scanning ps output
  --with-ps-cols=NUM      Number of columns in ps command
  --with-ps-varlist=LIST  Variable list for sscanf of 'ps' output
```

I guess there is no easy fix for this.

sanchezfauste:
I will fix the hard coded `/bin/ps`. What do you suggest for the `--sort=-pcpu`? Sort the output by hand using `qsort()`?

sanchezfauste:
Hi @sni,

I have tried to re-implement using the macros defined for the `ps` command and sorting the procs list by CPU usage using `qsort` function. If procpcpu is not present on PS_VARLIST the procs are printed without sorting.

I hope this implementation solves the two problems that you have mentioned before.

sni:
sorry this took a bit longer, finally merged

PR #1512, opened 2017-10-04
Fix check_procs not detecting zombies on Solaris

duncs:
When using:

```
check_procs -s Z -w :5 -c:10
```

zombie processes are not actually detected on Solaris due to pst3 ignoring
those it could not get the info for.

Zombie processes were created for testing by using:

```
nohup perl -e "if (fork()>0) {while (1) {sleep 100*100;};};"&
```


RincewindsHat:
Can someone reproduce this? Dont have a solaris around.

PR #1513, opened 2017-10-07
Added check_mysql_slave

ip1981:
Supports MariaDB multi-source replication

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -364,7 +364,7 @@ if test $with_mysql = "no" ; then
   AC_MSG_WARN([Skipping mysql plugin])
   AC_MSG_WARN([install mysql client libs to compile this plugin (see REQUIREMENTS).])
 else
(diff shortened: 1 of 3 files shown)


RincewindsHat:
Can this be integrated in `check_mysql`? may related to #1716

GerMalaz:
> Can this be integrated in `check_mysql`? may related to #1716

As author of #1716:
The goal of both is the same (coincidentally, they both use -N for the replica name).
It modifies the existing check_mysql rather than adding a new plugin.
It supports both MariaDB and MySQL named replicas, rather than just MariaDB.

waja:
As I was involved into the work removing `contrib/` I'm not a fan of adding new plugins (maybe nobody is caring).

RincewindsHat:
Sorry for the wait. Integrating the test into an existing plugins makes more sense to me.

PR #1520, opened 2017-10-27
Define our own default MySQL port in check_mysql*

orlitzky:
This fixes compilation with newer versions of MariaDB, the commit messages should explain things.

### DIFF
--- a/plugins/check_mysql.c
+++ b/plugins/check_mysql.c
@@ -1,33 +1,33 @@
 /*****************************************************************************
-* 
+*
 * Monitoring check_mysql plugin
-* 
+*
 * License: GPL
 * Copyright (c) 1999 Didi Rieder (adrieder@sbox.tu-graz.ac.at)
 * Copyright (c) 2000 Karl DeBisschop (kdebisschop@users.sourceforge.net)
 * Copyright (c) 1999-2017 Monitoring Plugins Development Team
-* 
+*
 * Description:
-* 
+*
 * This file contains the check_mysql plugin
-* 
+*
 * This program tests connections to a mysql server
-* 
-* 
+*
+*
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
-* 
+*
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
-* 
-* 
+*
+*
 *****************************************************************************/
 
 const char *progname = "check_mysql";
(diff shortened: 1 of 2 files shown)


waja:
I would opt for #1522

orlitzky:
@waja I alluded to this in my commit message, but using `MYSQL_PORT` from the header file was never a great choice: when you're monitoring various MySQL servers, the compiled-in port for the _local_ MySQL server (if you're even monitoring it!) is generally the wrong port to check. The "default default": of 3306 is more likely to be correct.

The fix in https://github.com/monitoring-plugins/monitoring-plugins/pull/1522 still has that downside, namely that the default port will be wrong if compiled on a weird machine. Moreover, it changes the behavior of the plugin (once again) depending on where it is compiled: if it's compiled on a machine with a new MariaDB, it will do something different than if compiled on a machine with an older version. That's a maintenance headache for little benefit in my opinion.

plugins/check_mysql.c
```diff
@@ -58,7 +62,7 @@ char *ciphers = NULL;
 bool ssl = false;
 char *opt_file = NULL;
 char *opt_group = NULL;
-unsigned int db_port = MYSQL_PORT;
+unsigned int db_port = CHECK_PORT_DEFAULT;
```
ofalk: My suggestion here would be
#ifndef MYSQL_PORT
#define MYSQL_PORT 3306
#endif 
to retain the same behaviour on old installs and systems with MySQL, instead of MariaDB.

orlitzky: But why go out of our way to do the wrong thing sometimes, when we could do the right thing unconditionally?

Maybe I'm asking the wrong question... let's make sure we're on the same foot: does anyone think that "whatever the default MySQL port was on the machine where monitoring-plugins was compiled" is a useful default for end-users of monitoring-plugins? (Consider that most users of monitoring-plugins have no idea where it was compiled.)

ofalk: For distributions that do ship packages in binary (eg. Fedora, CentOS, RHEL, SLES), it shouldn't matter, since they do compile 'em in an isolated environment.
That's why I'd suggest to use ifndef, since if someone really _intentionally_ want's to compile the check plugin differently, it's easier to define MYSQL_PORT somewhere, instead of changing the source and use a different _static_ value.

orlitzky: Is it easier to add `#define MYSQL_PORT 8675309` than it is to edit the existing line to say `#define CHECK_PORT_DEFAULT 8675309` instead of `#define CHECK_PORT_DEFAULT 3306`? I think they're roughly the same difficulty.

And besides, there are exactly zero people who want to do that. If you need a specific port, you can specify it on the command-line.

With all else being equal, the simpler shorter code that acts predictably is more fun to maintain.

ofalk: Sure it's the same "difficulty", no matter if it's called MYSQL_PORT or CHECK_PORT_DEFAULT. My intention/idea was to keep the same behaviour as it was (use port from MariaDB/MySQL header files), but allow override (eg. another MySQL fork comes along and uses another port...). And I'm a huge fan of reusing existing static values, if reasonable.
But it's your source and for sure, defining your own static value will probably increase predictability.

ofalk: Ping also from my side. Is there any intention to make this work for everybody?

PR #1521, opened 2017-11-02
check_ping: check PING_COMMAND existence

### LINKED ISSUE #1509: check_ping /check_icmp dosn't work in docker container
I maintain a icinga2 docker container.
When i will use the `check_ping` command i become these result:
```
/etc/icinga2 # /usr/lib/monitoring-plugins/check_ping -4  -H 8.8.8.8 -w '100,5%' -c '200,15%'
CRITICAL - You need more args!!!
Could not open pipe:
```

as alternative, i will use `check_icmp` that doesn't work either:
```
/etc/icinga2 # /usr/lib/monitoring-plugins/check_icmp -H 8.8.8.8  -s eth0 -v
ttl set to 64
Setting alarm timeout to 10 seconds
packets: 5, targets: 1
target_interval: 0.000, pkt_interval 80.000
crit.rta: 500.000
max_completion_time: 3400.000
crit = {500000, 80%}, warn = {200000, 40%}
pkt_interval: 80000  target_interval: 0  retry_interval: 0
icmp_pkt_size: 76  timeout: 10
Illegal instruction
```
The simple `ping` works fine:
```
/etc/icinga2 # ping -4 -c 1 8.8.8.8
PING 8.8.8.8 (8.8.8.8): 56 data bytes
64 bytes from 8.8.8.8: seq=0 ttl=45 time=15.523 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 15.523/15.523/15.523 ms
```

Now, i'm perplexed and need help.

Is this a limitation of docker, or an bug into both checks?

guessi:
`PING_COMMAND` and `PING6_COMMAND` should be checked before compile

Fixed #1509

### DIFF
--- a/REQUIREMENTS
+++ b/REQUIREMENTS
@@ -86,6 +86,10 @@ check_ups:
 check_ide_smart:
 	- Uses the Linux specific SMART interface [http://smartlinux.sourceforge.net/smart/index.php]. 
 
+check_ide_smart:
+	- Requires iputils or iputils-ping package
+	  https://github.com/iputils/iputils
+
 OS Specific Issues
 ------------------
 
--- a/REQUIREMENTS
+++ b/REQUIREMENTS
@@ -86,7 +86,7 @@ check_ups:
 check_ide_smart:
 	- Uses the Linux specific SMART interface [http://smartlinux.sourceforge.net/smart/index.php]. 
 
-check_ide_smart:
+check_ping:
 	- Requires iputils or iputils-ping package
 	  https://github.com/iputils/iputils
 
--- a/configure.ac
+++ b/configure.ac
@@ -1029,7 +1029,7 @@ AC_ARG_WITH(ping_command,
 AC_MSG_CHECKING(for ICMP ping syntax)
 ac_cv_ping_packets_first=no
 ac_cv_ping_has_timeout=no
-if test -n "$with_ping_command"
+if test -n "$PATH_TO_PING" && test -n "$with_ping_command"
 then
 	AC_MSG_RESULT([(command-line) $with_ping_command])
 	if echo "$with_ping_command" | grep '%d.*%d.*%s' >/dev/null
@@ -1121,7 +1121,7 @@ then
 	AC_MSG_RESULT([$with_ping_command])
 
 else
-	AC_MSG_WARN([unable to find usable ping syntax])
+	AC_MSG_WARN([unable to find usable ping command or syntax])
 fi
 
 AC_DEFINE_UNQUOTED(PING_COMMAND,"$with_ping_command",
@@ -1147,7 +1147,7 @@ AC_ARG_WITH(ping6_command,
 if test x"$with_ipv6" != xno ; then
 AC_MSG_CHECKING(for ICMPv6 ping syntax)
 ac_cv_ping6_packets_first=no
-if test -n "$with_ping6_command"
+if test -n "$PATH_TO_PING6" && test -n "$with_ping6_command"
 then
 	AC_MSG_RESULT([(command-line) $with_ping6_command])
 	if echo "$with_ping6_command" | grep '%d.*%d.*%s' >/dev/null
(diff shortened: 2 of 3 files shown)


guessi:
@waja could you please verify the commit again, thanks

waja:
As #1509 is closed, is this still relevant?

guessi:
@waja it have been 4 years... almost forget I have created this PR.

I just review the issue once again, #1509 is resolving an issue of package build under Alpine Linux, which was [resolved](https://github.com/monitoring-plugins/monitoring-plugins/issues/1509#issuecomment-478840268) already. And #1521 (this PR) is aim to ensure required `PING_COMMAND` and `PING6_COMMAND` have correctly installed at build time. I'm not sure if it is still an issue right now, but I think they are two separated topics.

waja:
The question is: Is `ping` or `ping6` really needed on compile time or is it sufficient when it's available at runtime?

sni:
The configure script tests available ping options. Thats why it requires the ping command to be available. So i guess this is a good idea.

waja:
> The configure script tests available ping options.

but you can use something like [this](https://salsa.debian.org/nagios-team/pkg-monitoring-plugins/-/blob/master/debian/rules#L54-60):
```
ifeq ($(DEB_HOST_ARCH_OS),linux)
PING_CONFIGURE_ARGS = --with-ping-command="/bin/ping -4 -n -U -w %d -c %d %s" \
                      --with-ping6-command="/bin/ping -6 -n -U -w %d -c %d %s"
else
PING_CONFIGURE_ARGS = --with-ping-command="/bin/ping -n -w %d -c %d %s" \
                      --with-ping6-command="/bin/ping6 -n -w %d -c %d %s"
endif
```

In this case, is technically `ping` still needed at compile time?

sni:
strictly speaking, ping is not needed at compile (make) time, but required at "./configure" time.
I guess it would be nicer to have the options as configure options instead of running ping and having to guess the options.
Basically the configure trys to detect the different output formats and sets the options: ac_cv_ping_packets_first, ac_cv_ping6_packets_first and ac_cv_ping_has_timeout. Which then leads to the macros:
PING_COMMAND, PING_HAS_TIMEOUT, PING_PACKETS_FIRST, PING6_COMMAND and PING6_PACKETS_FIRST.
If all those variables could be set from configure options, then no ping would be required during configure and make.

I guess in the end its easier to simply install ping.

sni:
why not add it as `Build-Depends` to debian/control or the dsc file then?

waja:
Yes, this is what it's done. But `ping` can behave differently. Depending on the version of `ping`, the architecture and depending on the [implementation](https://packages.debian.org/search?searchon=contents&keywords=bin%2Fping&mode=path&suite=stable&arch=any).

guessi:
Initially, the goal of the PR is quite simple. To tell user that `ping[6] command from iputils-ping` is required for running `check_ping.c` with `ping[6]` command. But there are so many distributions to support, it is hard to ensure the package could be run on all platforms, so I choose the most simple way to archive that goal at that time.

But still having some issues,
- If we only check command existence before run, how do we know "which package" provide X command?
- If we put package X as dependency (spec file, deb file), need to take care of compatibilities as well.

To be more strictly,
- Version of X ( breaking change ? )
- Platform compatibility ( inetutils-utils, iputils-ping ? )
- $PATH issue ( if the X command really X => security issue )
- and more...

Maybe it could be simpler by adding a new section to [FAQ](https://www.monitoring-plugins.org/doc/faq/index.html) or have some update to outdated [REQUIREMENTS](https://github.com/monitoring-plugins/monitoring-plugins/blob/master/REQUIREMENTS).

---

_Please feel free to close this PR or re-create with another topic._

waja:
I would leave distro maintainers to configure it at build-time, depending on available ping(6) implementations.

PR #1527, opened 2018-01-19
check_snmp: make calcualtion of timeout value in help output more clear

m-kraus:
Previously it was unclear that when setting a timeout using -t it is not the final timeout interval in check_snmp.

The final timeout interval is calculated using "timeout_interval * retries + 5" it seems.

This PR makes this more clear in the help output.


m-kraus:
@sni What's your opinion about this?

PR #1531, opened 2018-04-19
check_ping: pass -4 to the ping command, fixes #898

### LINKED ISSUE #898: check_ping ignores -4 option [sf#2832817]
Submitted by phalenor on 2009-08-05 22:15:37

check_ping v1991 (nagios-plugins 1.4.13)
Solaris 10u5 sparc

check_ping ignores -4 option when the host passed with -H is a hostname which has both A and AAAA records associated with it. Expected behavior is that -4 forces check_ping to use an IPv4 address returned from an A record lookup of the host if a hostname is provided. This is the behavior seen with other plugins such as check_ssh, which correctly query over only IPv4 or IPv6 with -4 or -6 are passed, respectively.

sbraz:
When trying to ping a host which has A and AAAA records, we need to pass
'-4' to the ping command.

### DIFF
--- a/plugins/check_ping.c
+++ b/plugins/check_ping.c
@@ -113,7 +113,7 @@ main (int argc, char **argv)
 		if (address_family != AF_INET && is_inet6_addr(addresses[i]))
 			rawcmd = strdup(PING6_COMMAND);
 		else
-			rawcmd = strdup(PING_COMMAND);
+			xasprintf (&rawcmd, "%s %s", PING_COMMAND, "-4");
 #else
 		rawcmd = strdup(PING_COMMAND);
 #endif


sni:
since we already distinguish between PING_COMMAND and PING6_COMMAND, it would be better to detect the -4 option during configure time and add it there. Whats the reason of having a configure option for the ping command if we change it later anyway.

sbraz:
@sni any kind of fix works for me. Feel free to open another PR and I'll close this one.

waja:
> @sni any kind of fix works for me. Feel free to open another PR and I'll close this one.

what about this?

```
$ ./configure --help | grep ping
  --with-ping-command=SYNTAX sets syntax for ICMP ping
  --with-ping6-command=SYNTAX sets syntax for ICMPv6 ping
  --with-fping-command=PATH Path to fping command
  --with-fping6-command=PATH Path to fping6 command
```

sbraz:
@waja OK I see how this makes sense. I noticed that my check_ping [was built with](https://github.com/gentoo/gentoo/blob/dbc12eca2a307d8c5a36133d39f1f87bf27b5c11/net-analyzer/monitoring-plugins/monitoring-plugins-2.2-r3.ebuild#L73):
`--with-ping-command="/bin/ping -n -U -w %d -c %d %s"`.
I'm going to get it fixed downstream, however, it would be nice to have the configure script add the `-4` option to ping_command automatically.

waja:
> it would be nice to have the configure script add the `-4` option to ping_command automatically.

This depends much on the implementation of `ping` used in your used environment. So it does not make much sense to implement this in our project. This should be done by distribution maintainers or people compiling it depending on their use case.

sbraz:
I see, I'll close this PR then.

PR #1535, opened 2018-05-28
Feature check curl

sni:
This PR adds check_curl as optional and experimental new plugin. Options and behaviour is mostly like check_http.
To enable this plugin, you need to configure plugins with `./configure --enable-check-curl`


ekeih:
Note: I had to use `./configure --enable-check-curl --with-libcurl`.

ekeih:
I deployed the check for ~1000 checks as a first test. Most of the checks look promising but until now I found two issues:

```
HTTP CRITICAL HTTP/1.x 0 unknown - Unparseable status line in 0.0286 seconds response time
```

and

```
*** Error in `/usr/lib/nagios/plugins/check_curl': free(): invalid pointer: 0x000000000041910e ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x2af2a6b427e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x2af2a6b4b37a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x2af2a6b4f53c]
/usr/lib/nagios/plugins/check_curl[0x407eda]
/usr/lib/nagios/plugins/check_curl[0x4074a9]
/usr/lib/nagios/plugins/check_curl[0x4026f5]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x2af2a6aeb830]
/usr/lib/nagios/plugins/check_curl[0x402749]
======= Memory map: ========
...
```

I am aware that this output alone is not very helpful yet. I am out of office for a few days, but afterwards I will test this again and I will try to post more details.

sni:
oh great, that would be quite helpful. For the first one, a check_curl -vvv with at least the headers would be interesting. The second one run by gdb with a backtrace would be nice. Thanks.

ekeih:
About the unparsable status line: The connection seems to die after the headers. It happens for several checks. As far as I currently know all affected setups have a nginx revers proxy and a java application as backend (e.g. custom application in tomcat or a jira instance). But it does not fail for all java backends (e.g. a confluence instance works just fine)!

# check_curl

```
/usr/lib/nagios/plugins/check_curl -I 10.x.x.x -H foo.bar.de -vvv
* Socket timeout set to 10 seconds
* SSL library string is : OpenSSL/1.0.2g OpenSSL (1)
* Added foo.bar.de:80:10.x.x.x to DNS cache
* Hostname foo.bar.de was found in DNS cache
*   Trying 10.x.x.x...
* Connected to foo.bar.de (10.x.x.x) port 80 (#0)
> GET / HTTP/1.1
Host: foo.bar.de
User-Agent: check_curl/v2.2.124.gedba (monitoring-plugins 2.2)
Accept: */*
Connection: close

< HTTP/1.1 200
< Server: nginx
< Date: Thu, 31 May 2018 09:56:42 GMT
< Content-Type: text/html;charset=UTF-8
< Transfer-Encoding: chunked
< Connection: close
< X-AREQUESTID: 716x2463604x1
< X-ASEN: SEN-2188512
< Set-Cookie: xxxSomeCookiexxx
< X-AUSERNAME: anonymous
< X-Content-Type-Options: nosniff
< Set-Cookie: xxxSomeCookiexxx
< X-UPSTREAM: 10.x.x.x:x
<
* Closing connection 0
HTTP CRITICAL HTTP/1.x 0 unknown - Unparseable status line in 0.0293 seconds response time|time=0.029325s;0;0;; size=46765B;;;

```

# check_http

```
/usr/lib/nagios/plugins/check_http -I 10.x.x.x -H foo.bar.de -vvv
GET / HTTP/1.1
User-Agent: check_http/v2.1.2 (monitoring-plugins 2.1.2)
Connection: close
Host: foo.bar.de


http://10.x.x.x:80/ is 47295 characters
STATUS: HTTP/1.1 200
**** HEADER ****
Server: nginx
Date: Thu, 31 May 2018 09:56:54 GMT
Content-Type: text/html;charset=UTF-8
Transfer-Encoding: chunked
Connection: close
X-AREQUESTID: 716x2463607x1
X-ASEN: SEN-2188512
Set-Cookie: xxxSomeCookiexxx
X-AUSERNAME: anonymous
X-Content-Type-Options: nosniff
Set-Cookie: xxxSomeCookiexxx
X-UPSTREAM: 10.x.x.x:8081
**** CONTENT ****
1905

... the full body ...

0


HTTP OK: HTTP/1.1 200 - 47295 bytes in 1.192 second response time |time=1.192152s;;;0.000000 size=47295B;;;0
```

ekeih:
About the pointer issue:

I think it is easy to reproduce :) Just use `-f follow|sticky|stickyport` with a server that answers with a redirect.
For example:

```
/usr/lib/nagios/plugins/check_curl -H heise.de -vvv -f sticky
```

ekeih:
Another topic:

I told @hartwork that I found his code (uriparser) in your PR. Of course he liked the fact that other people use his software :) But he asked a few questions:

1. Why do you bundle the source of uriparser in your source? The library should be available in all linux distros by now.
2. Your uriparser version is not the latest. Is this on purpose?
3. He was also surprised that you have 83 commits (including several merge commits) in your PR.

As I was not able to answer any of this questions I decided to forward them to you. Maybe they are helpful for you :)

ekeih:
The check fails when I use `-C`:

```
# /usr/lib/nagios/plugins/check_curl -H google.de -C 31
CRITICAL - Cannot retrieve certificate subject.

# /usr/lib/nagios/plugins/check_curl -H google.de -S
HTTP OK: HTTP/1.1 301 Moved Permanently - 710 bytes in 0.104 second response time |time=0.103541s;0;0;; size=219B;;;
```

andreasbaumann:
I fixed the free of server_url in case it's still initialized with '/' (i.e. when you don't specify the URL parameter).

About uriparser: it was the newest version when I was originally working on the plugin. :-)
I'll update it. I integrated it because distributions tend to bundle even older versions, for instance
Centos 7: uriparser-0.7.5-9.el7.x86_64. Centos 6 has nop uriparser (unless maybe in EPEL).

Thanks a lot for testing and for providing feedback. :-)

andreasbaumann:
We could add a --with-uriparser option to take the system version and bundle a local version which can be used with --with-local-uriparser, or something similar?

andreasbaumann:
About the issue:

```
./check_curl -H google.de -C 31
CRITICAL - Cannot retrieve certificate subject.
```
When I do the same on Archlinux I get:

```
OK - Certificate '*.google.de' will expire on Tue 07 Aug 2018 07:54:00 PM GMT +0000.
```
On which Linux distribution are you running check_curl? Curl itself could be too old to understand
wildcards hosts in the subject line..

andreasbaumann:
I'm thinking: if somebody calls ./configure with --enable-check-curl maybe we should enable --with-curl (without directory) automatically?

ekeih:
> I fixed the free of server_url in case it's still initialized with '/' (i.e. when you don't specify the URL parameter).

Thanks for the fix. I can confirm that it works for us :)

> I'm thinking: if somebody calls ./configure with --enable-check-curl maybe we should enable --with-curl (without directory) automatically?

Yes

> On which Linux distribution are you running check_curl? Curl itself could be too old to understand
wildcards hosts in the subject line..

It's an Ubuntu 16.04.4 LTS server with current updates installed. I have the suspicion that I screwed up the SSL support:

```
# ./configure --enable-check-curl --with-libcurl
...
configure: WARNING: OpenSSL or GnuTLS libs could not be found or were disabled
...
```

```
# dpkg -l | grep -i curl
ii  curl                           7.47.0-1ubuntu2.8                          amd64        command line tool for transferring data with URL syntax
ii  libcurl3:amd64                 7.47.0-1ubuntu2.8                          amd64        easy-to-use client-side URL transfer library (OpenSSL flavour)
ii  libcurl3-gnutls:amd64          7.47.0-1ubuntu2.8                          amd64        easy-to-use client-side URL transfer library (GnuTLS flavour)
ii  libcurl4-openssl-dev:amd64     7.47.0-1ubuntu2.8                          amd64        development files and documentation for libcurl (OpenSSL flavour)
ii  python3-pycurl                 7.43.0-1ubuntu1                            amd64        Python bindings to libcurl (Python 3)

# dpkg -l | grep -i openssl
ii  libcurl3:amd64                 7.47.0-1ubuntu2.8                          amd64        easy-to-use client-side URL transfer library (OpenSSL flavour)
ii  libcurl4-openssl-dev:amd64     7.47.0-1ubuntu2.8                          amd64        development files and documentation for libcurl (OpenSSL flavour)
ii  libgnutls-openssl27:amd64      3.4.10-4ubuntu1.4                          amd64        GNU TLS library - OpenSSL wrapper
ii  openssl                        1.0.2g-1ubuntu4.12                         amd64        Secure Sockets Layer toolkit - cryptographic utility
```

Any advise what I am missing?

andreasbaumann:
I think you miss a libssl-dev. :-)

andreasbaumann:
I just compiled on Ubuntu 16.04 LTS myself. I get:

```
./check_curl -H google.de -S -C 30
OK - Certificate '*.google.de' will expire on Tue Aug  7 19:54:00 2018 +0000.
```

ekeih:
> I think you miss a libssl-dev. :-)

Yes, I did! Thank you :)

> I just compiled on Ubuntu 16.04 LTS myself. I get:

Yes, after compiling with libssl-dev it works.


To summarize the different issues of this growing ticket:
- `The check fails when I use -C` -> fixed by compiling with SSL
- `free of server_url` -> fixed
- `unparsable status line` -> still open

Let me know when I can do something to help to fix the last one :)

weiss:
> I integrated it because distributions tend to bundle even older versions

And those won't work?

> We could add a --with-uriparser option to take the system version and bundle a local version which can be used with --with-local-uriparser, or something similar?

I'd rather avoid bundling libraries altogether, to be honest.  If old versions won't work, I'd let `./configure` check for the required version and leave the rest to package maintainers (or users, if they really compile our plugins from source).

andreasbaumann:
The last one: I don't know if the unparsable status line is the root cause or the connection is closed thus causing the unparsable status line mesage by accident.
I tested with another nginx proxy, there it worked.
Ideally a publicly available URL demononstrating the issue would be an idea.

ekeih:
>  Ideally a publicly available URL demononstrating the issue would be an idea.

I will try.

andreasbaumann:
uriparser: I upgraded to the newest version for now.
There is also the picohttpparser which is currently bundled.
To be frank: I would rather bundle libcurl and all crypto libraries as they tend to work very inconsistently
across Linux distros (OpenSSL, LibreSSL, GnuTLS, NSS, etc.)

weiss:
> There is also the picohttpparser library which is currently bundled (not even on Archlinux its a package,
you have to build it from the AUR - correction: not even that).

That's not an uncommon situation when adding dependencies.  Distribution maintainers will most probably package them up if we require them.

> I would rather bundle libcurl and all crypto libraries as they tend to work very inconsistently
across Linux distros (OpenSSL, LibreSSL, GnuTLS, NSS, etc.). ;-)

The main target audience of our source tarball isn't users but distribution maintainers.  They won't use any versions we bundle anyway (and will be annoyed if they have to take any action to ignore the bundled stuff).  I think the fraction of users building our plugins from source is negligible, but those who do build from source will be used to coping with dependencies themselves.

The main problem I see with bundling software is that we become responsible for the bundled code.  We'll have to track each bundled project for (security) flaws and publish a new Monitoring Plugins release whenever there's a serious issue in one of the bundled dependencies.  One other problem I see is that you more easily run into licensing issues when bundling stuff.

So unless there's a reason we *must* bundle a library, I'm opposed to doing this.

hartwork:
Hi!

> To be frank: I would rather bundle libcurl and all crypto libraries as they tend to work very inconsistently across Linux distros (OpenSSL, LibreSSL, GnuTLS, NSS, etc.). ;-)

Please don't. You will not have the time to track and apply security updates within all of these. Hence they will all turn vulnerable at some point.

I would like to share this article:
https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies

> We could add a --with-uriparser option to take the system version and bundle a local version which can be used with --with-local-uriparser, or something similar?

If you decide to continue to bundle, that switch would help, yes. Please make compilation against the system-wide library the default then and also ensure that if there is autodetection it can be bypassed forcefully by people packaging.

Thanks!

andreasbaumann:
You misread my winking face in the note about curl. :-)

You fully convinced me to remove the local uriparser library. It has a stable ABI and releasing.

picohttpparser I would actually keep as local copy for now. I asked the author whether he could
provide at least a relase, see https://github.com/h2o/picohttpparser/issues/42.

To justify the use of so many third-party libraries. There are two extremes: check_http which 
implements everything on its own, but has the benefit to be simple and compile without tons of
third-party libraries (sort of a busybox/tinybox version). check_curl is more in the corner of
prividing most features, but it gets all nice stuff from third-party libraries. It would have felt
bad to use libcurl, but to do URL and HTTP header parsing by hand again.

weiss:
> picohttpparser I would actually keep as local copy for now.

Ah, I wasn't aware there's no release yet.  That's a good point I guess.  I think we should still set things up to use the bundled version only as a fallback.

> I asked the author whether he could provide at least a release, see h2o/picohttpparser#42.

Thanks!

> it gets all nice stuff from third-party libraries.

Yes, I'm all for re-using existing libraries.

Many thanks for your work on `check_curl`.

plugins/check_curl.c
```diff
@@ -0,0 +1,2242 @@
...
+      (int)body_buf.buflen);
+  }
+
+  /* return a CRITICAL status if we couldn't read any data */
+  if (strlen(header_buf.buf) == 0 && strlen(body_buf.buf) == 0)
+    die (STATE_CRITICAL, _("HTTP CRITICAL - No header received from host\n"));
+
+  /* get status line of answer, check sanity of HTTP code */
+  if (curlhelp_parse_statusline (header_buf.buf, &status_line) < 0) {
+    snprintf (msg, DEFAULT_BUFFER_SIZE, "Unparseable status line in %.3g seconds response time|%s\n",
```
ekeih: Unparsable

andreasbaumann:
Aha: uriparser on Ubuntu for Travis builds. This could be an issue. I could add a launchpad repo in the prepare script for Travis..

weiss:
> Aha: uriparser on Ubuntu for Travis builds. This could be an issue. I could add a launchpad repo in the prepare script for Travis..

Yes if one exists and it's easy to use, that would be an option.  Or we bundle it temporarily just for Travis without including it with the release tarball.

ekeih:
### unparsable status line

Unfortunately I am not allowed to share the affected customer setups. Instead I learned to use gdb and with the support of a coworker we figured it out. Or at least we think we did.

The `curlhelp_parse_statusline` function returns in https://github.com/monitoring-plugins/monitoring-plugins/blob/25a2205f7ab4037c5b63f3af3b1533e0b472327f/plugins/check_curl.c#L1925

It works when the status line looks like this: `HTTP/1.1 200 OK`
If fails when the status line looks like this: `HTTP/1.1 302`
`check_http` works for both.

From https://www.ietf.org/rfc/rfc2616.txt

```

6 Response

   After receiving and interpreting a request message, a server responds
   with an HTTP response message.

       Response      = Status-Line               ; Section 6.1
                       *(( general-header        ; Section 4.5
                        | response-header        ; Section 6.2
                        | entity-header ) CRLF)  ; Section 7.1
                       CRLF
                       [ message-body ]          ; Section 7.2

6.1 Status-Line

   The first line of a Response message is the Status-Line, consisting
   of the protocol version followed by a numeric status code and its
   associated textual phrase, with each element separated by SP
   characters. No CR or LF is allowed except in the final CRLF sequence.

       Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF

6.1.1 Status Code and Reason Phrase

   The Status-Code element is a 3-digit integer result code of the
   attempt to understand and satisfy the request. These codes are fully
   defined in section 10. The Reason-Phrase is intended to give a short
   textual description of the Status-Code. The Status-Code is intended
   for use by automata and the Reason-Phrase is intended for the human
   user. The client is not required to examine or display the Reason-
   Phrase.
```

So according to the RFC the unparsable status line is missing a `Reason-Phrase`. So the question is: Should `check_curl` enforce the RFC or should it be as compatible as possible to `check_http`? Personally I would prefer the latter because it makes it easier to use `check_curl` as a drop-in replacement.

ekeih:
I am trying to compile the new version with the system version of uriparser. 

I thought it would work like this:

```
# dpkg -l | grep uriparser
ii  liburiparser-dev               0.8.4-1                                    amd64        development files for uriparser
ii  liburiparser1:amd64            0.8.4-1                                    amd64        URI parsing library compliant with RFC 3986

# ./configure --enable-check-curl --with-libcurl --with-uriparser
```

But I always get:

```
configure: WARNING: Skipping curl plugin
configure: WARNING: install the uriparser library to compile this plugin (see REQUIREMENTS).
```

I guess it's pretty obvious that I don't know much about C development ;) So I would be grateful for some hint what I am missing.

ekeih:
FYI: https://github.com/monitoring-plugins/monitoring-plugins/blob/25a2205f7ab4037c5b63f3af3b1533e0b472327f/plugins/check_curl.c#L1016 is affecting some of our checks.

andreasbaumann:
Hi, first you need pkg-config (I should improve the test for pkg-config uriparser a little bit).
Failures during configure are in config.log:

...
pkg-config: no
configure:16190: gcc -o conftest -g -O2    -L. conftest.c   >&5
conftest.c:70:27: fatal error: uriparser/Uri.h: No such file or directory
```

I'll improve this test, most likely by just simply croaking in configure if pkg-config is not installed.

**Comment by andreasbaumann** [comment]
```
 die (STATE_UNKNOWN, _("HTTP UNKNOWN - non-absolute location, not implemented yet!\n")); 
```

Yes, that's one of the todos, so I'll prioritize this one.. :-)

**Comment by ekeih** [comment]
> first you need pkg-config

Thanks, it works now :)

> Yes, that's one of the todos, so I'll prioritize this one.. :-)

Thank you for pushing 5e46233e2ea5c0d6bcbbd59fb75b07b33f6ce6cc. Our affected checks work fine now.

**Comment by ekeih** [comment]
We stumbled about the fact, that check_curl uses another user agent than check_http. As a drop-in replacement we could consider to use the same.
Personally I think that it is very questinable to have checks that rely on the user agent ;-) I just thougth I would let you know that this could result in issues for some users.

**Comment by andreasbaumann** [comment]
```
  const char *progname = "check_curl";

  snprintf( user_agent, DEFAULT_BUFFER_SIZE, "%s/v%s (monitoring-plugins %s)",
    progname, NP_VERSION, VERSION);
```

So it's just about changing progname from check_curl to check_http.

On the other hand we may want to be able to distinguish error reports from people using check_http
or check_curl as long as the plugins run in parallel.

Thoughs?

**Comment by ekeih** [comment]
>`snprintf( user_agent, DEFAULT_BUFFER_SIZE, "%s/v%s (monitoring-plugins %s) (curl)",
    progname, NP_VERSION, VERSION);`

I changed the user agent in a similar way to make it work for us.

I am happy to inform you that currently check_curl and check_http deliver the same results for all of the ~2000 HTTP services in our setup.

**Comment by odenbach** [comment]
I also did some checks. While ordinary checks work really well, check_curl fails with advanced virtual host and/or virtual port setups (which work correctly with check_http). Please have a look at the following table:

```
-I      -p      -H      v_host  v_port  s_host  s_port  result
--------------------------------------------------------------
A                       A       80      A       80      pass
A       8080            A       8080    A       8080    pass
                A       A       80      A       80      pass
                A:8080  A       8080    A       8080    FAIL
A               B       B       80      A       80      FAIL
A       8080    B       B       8080    A       8080    FAIL
A       8080    B:80    B       80      A       8080    FAIL
```
The first three columns show the used parameters, v_host and v_port are the resulting virtual host name and virtual port (which must be used in the HTTP request), s_host and s_port are the resulting server host name and server port (which must be used to connect to).

Some of these checks are already implemented in the test suite, so `make test` should mention them as failed.

I am currently looking through the code to see if this behaviour can be corrected easily but it takes some time.

**Comment by odenbach** [comment]
Please have a look at #1540.

**Comment by theftf** [comment]
I am unable to build this with SSL Support. Any tips? Tried it with openssl and gnutls on Debian and CentOS :(

**Comment by ekeih** [comment]
@theftf Please post what you tried and what errors you got. It is a lot easier to help when we see your output.

**Comment by theftf** [comment]
On Debian I had:
`            --with-apt-get-command: /usr/bin/apt-get
              --with-ping6-command: 
               --with-ping-command: /bin/ping -n -U -w %d -c %d %s
                       --with-ipv6: yes
                      --with-mysql: no
                    --with-openssl: no
                     --with-gnutls: yes
               --enable-extra-opts: yes
                       --with-perl: /usr/bin/perl
             --enable-perl-modules: no
                     --with-cgiurl: /nagios/cgi-bin
               --with-trusted-path: /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
                   --enable-libtap: no
               --enable-check-curl: yes
                    --with-libcurl: /usr/include/x86_64-linux-gnu/curl/
                  --with-uriparser: /usr/include/uriparser/
`
and:
`            --with-apt-get-command: /usr/bin/apt-get
              --with-ping6-command: 
               --with-ping-command: /bin/ping -n -U -w %d -c %d %s
                       --with-ipv6: yes
                      --with-mysql: no
                    --with-openssl: yes
                     --with-gnutls: no
               --enable-extra-opts: yes
                       --with-perl: /usr/bin/perl
             --enable-perl-modules: no
                     --with-cgiurl: /nagios/cgi-bin
               --with-trusted-path: /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
                   --enable-libtap: no
               --enable-check-curl: yes
                    --with-libcurl: /usr/include/x86_64-linux-gnu/curl/
                  --with-uriparser: /usr/include/uriparser/
`

On CentOS I build with:
`            --with-apt-get-command: 
              --with-ping6-command: 
               --with-ping-command: /usr/bin/ping -n -U -w %d -c %d %s
                       --with-ipv6: yes
                      --with-mysql: no
                    --with-openssl: yes
                     --with-gnutls: no
               --enable-extra-opts: yes
                       --with-perl: /usr/bin/perl
             --enable-perl-modules: no
                     --with-cgiurl: /nagios/cgi-bin
               --with-trusted-path: /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
                   --enable-libtap: no
               --enable-check-curl: yes
                    --with-libcurl: /usr/include/curl
                  --with-uriparser: /usr/include/uriparser
`

The build runs smooth every time but the resulting binary gives me:
`root@s031025:~/monitoring-plugins# ./plugins/check_curl -S heise.de
check_http: Invalid option - SSL is not available
`
check_http from the same build works with SSL so the ssl libs should work in theory.

**Comment by andreasbaumann** [comment]
Did you try with --with-openssl respectively --with-gnutls? Do you have the corresponding
development packages installed?

**Comment by theftf** [comment]
@andreasbaumann this way it worked :)
--with-openssl=/usr instead of just --with-openssl did the trick. the rest of monitoring plugins seen fine with just --with-openssl.

thanks a lot so i can try out this check

**Comment by andreasbaumann** [comment]
Mmh. `--with-openssl` leads to lib/yes errors, so something is wrong in those library checks.
Having to specify `--with-openssl=/usr` feels akward.

**Comment by theftf** [comment]
In found a little Bug:
`
HTTP OK: HTTP/1.1 200 OK - 321079 bytes in 0.296 second response time |time=0.296314s;0;0;; size=320692B;;;
`
I have two different values for the contend size.

And the time values differ from the old check_http:
`
$ ./check_curl -S -H www.heise.de -u /newsticker/;./check_http -S -H www.heise.de -u /newsticker/
HTTP OK: HTTP/1.1 200 OK - 321079 bytes in 0.296 second response time |time=0.296314s;0;0;; size=320692B;;;
HTTP OK: HTTP/1.1 200 OK - 321079 bytes in 0.126 second response time |time=0.126370s;;;0.000000;10.000000 size=321079B;;;0
`
But there I don't know which time is measured there. The response time should not be different. If the running time of the program is somehow measured this could be expected.

**Comment by andreasbaumann** [comment]
The second size seems to be content+header, so we should subtract the length of the header.
We should also check, that we get the execution time from curl including DNS lookups and such. Maybe this explains the difference.

**Comment by winter1967** [comment]
Hi,
i'm not able to configure with enabling check-curl:

configure: WARNING: unrecognized options: --enable-check-curl

**Comment by theftf** [comment]
@winter1967 did you chekout the correct branch?

**Comment by sni** [comment]
warning: i rebased this hole thing onto latest master...

Then i fixed the difference in the reported page size. And aligned performance data with the ones from check_http.

@ekeih: that should do the trick, when having `liburiparser-dev` installed, checkout this feature branch and `./configure --enable-check-curl` should be sufficient.

**Comment by theftf** [comment]
@sni I built your new version. now the performance data delivered by check_curl is consistent in itself :)

great work :+1: 

it still differs from the data delivered by check_http. but i think that's not so important.

**Comment by sni** [comment]
so, ports and host headers should be fine now. If anyone wants to test, now would be a good time to do so.
The reported response times differ slightly from check_http, around 4% in my tests, i guess we can't really do much about that.

**Comment by ekeih** [comment]
Thank you for your work and the progress :)
Unfortunately I will only be able to test this in November. So it will take some time until you get feedback from me.

**Comment by andreasbaumann** [comment]
Tested on 32-bit/64-bitLooking good on Archlinux (32-bit and 64-bit Intel, as well as ARMv6, ARMv7),
everything compiles and tests are passing. I upgraded some monitoring installations and all seems fine
too.

**Comment by andreasbaumann** [comment]
Built on Centos 6, 64-bit. 

https://centos.pkgs.org/6/repoforge-x86_64
uriparser-devel-0.7.5-1.el6.rf.x86_64.rpm.html
(also needs cpptest, same repo)

Cannot retrieve certificates (libcurl linked with SSL library 'NSS' is too old)
=> ok, not much we can do about that but rebuilt libcurl on Centos 6.

**Comment by sni** [comment]
Thanks for testing.
I noticed something on debian 9, which ships libssl version 1.1 and 1.0. 1.1 beeing the default is used to build the plugins, but libcurl is linked again 1.0 which leads to:
```
ldd plugins/check_curl | grep ssl
        libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007f3095aac000)
        libssl.so.1.0.2 => /usr/lib/x86_64-linux-gnu/libssl.so.1.0.2 (0x00007f3093f35000)
```
and this consequently segfaults on checking certificates. But i guess there is nothing we can do about it.

**Comment by odenbach** [comment]
Hi,

just checked our wicked port-rewriting setup, and it works - apart from some funny warning message:

$ ./check_curl -H www.uni-paderborn.de:443 -I lyon1.uni-paderborn.de -p 6033 -S -v
* curl CURLOPT_URL: https://lyon1.uni-paderborn.de:6033/
* Address in 'www.uni-paderborn.de:6033:lyon1.uni-paderborn.de' found illegal!
*   Trying 131.234.189.49...
* TCP_NODELAY set
* Connected to lyon1.uni-paderborn.de (131.234.189.49) port 6033 (#0)
[...]
*  SSL certificate verify ok.
> GET / HTTP/1.1
Host: www.uni-paderborn.de
[...]
HTTP OK: HTTP/1.1 200 OK - 151539 bytes in 0,034 second response time |time=0,033852s;;;0,000000;10,000000 size=151539B;;;0

So looks good to me.

Thanks,

Christopher

**Comment by sni** [comment]
thanks for testing, i removed that part now since it wasn't neccessary anyway.

**Comment by odenbach** [comment]
Rechecked again - warning has vanished, perfect!

**Comment by andreasbaumann** [comment]
I see some changed behaviour when checking certificate validity compared to earlier versions:

```
check_curl -v -H www.openssl.org -C 14,7 -S -j CONNECT 
* curl CURLOPT_URL: https://www.openssl.org:443/
* <url> malformed
* Closing connection -1
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 3 - URL using bad/illegal format or missing URL
```

This is on Centos 6 with a curl 7.65 (self-compiled)

**Comment by sni** [comment]
@andreasbaumann that seems to be related to the "CONNECT" method which enables proxy mode which then requires `-u` option to be set. I will have a look.

**Comment by andreasbaumann** [comment]
check_curl -v -H www.openssl.org -C 14,7 -S -j CONNECT -u /

I'll tested also on Archlinux, same behaviour.

**Comment by sni** [comment]
i can reproduce it on debian as well, i don't think its related to the os or the libcurl version.

**Comment by andreasbaumann** [comment]
In libcurl lib/url.c: 

```
rc = sscanf(data->change.url, "%[^\n/?#]%[^\n]", conn->host.name, path);

failf(data, "<url> malformed");
          return CURLE_URL_MALFORMAT;
```
Most likely it doesn't like the :443, as this is implicitely the default port of https.

**Comment by andreasbaumann** [comment]
I added some code around the 'CONNECT handling:

```
    if (verbose>=1)
      printf ("* curl CURLOPT_URL (CONNECT): %s\n", server_url);
    handle_curl_option_return_code (curl_easy_setopt (curl, CURLOPT_URL, server_url), "CURLOPT_URL");
```

then I get:

```
./check_curl -vv -H www.openssl.org -C 14,7 -S -j CONNECT     
* Set SSL/TLS version to 0
* Set SSL/TLS version to 0
* Socket timeout set to 10 seconds
* curl CURLOPT_URL: https://www.openssl.org:443/
* curl CURLOPT_PROXY: www.openssl.org:443
* curl CURLOPT_URL (CONNECT): /
* SSL library string is : OpenSSL/1.1.1 OpenSSL (1)
* Closing connection -1
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 3 - URL using bad/illegal format or missing URL
```

So, server_url is the relative path, url is the full URL. When I put 'url', it makes more sense. But then,
why is it treated specially in the 'CONNECT' case at all?

**Comment by andreasbaumann** [comment]
I found some other problems around certificate checking and -ffollow not working all the time.. but I have to test more..

**Comment by andreasbaumann** [comment]
I found some other problems around certificate checking and -ffollow not working all the time.. but I have to test more..

**Comment by andreasbaumann** [comment]
About the redirection: -ffollow doens't do what -fcurl does and not what it did before:

```
./check_curl -v -fcurl -w 10 -c 30 -I www.zh.ch -H www.zh.ch -u "/content/internet/de/service/suche.html?_charset_=utf-8&keywords=tq92eurospiderzhinternetyc8+bu32eurospiderzhinternetp93" -r "tq92eurospiderzhinternetyc8"
```
works:

```
HTTP OK: HTTP/1.1 200 OK - 49352 bytes in 0.678 second response time |time=0.678427s;10.000000;30.000000;0.000000;31.000000 size=49352B;;;0
```

and

```
./check_curl -v -ffollow -w 10 -c 30 -I www.zh.ch -H www.zh.ch -u "/content/internet/de/service/suche.html?_charset_=utf-8&keywords=tq92eurospiderzhinternetyc8+bu32eurospiderzhinternetp93" -r "tq92eurospiderzhinternetyc8"
```

results in:

```
HTTP CRITICAL: HTTP/1.1 200 OK - pattern not found - 42264 bytes in 0.022 second response time |time=0.021760s;10.000000;30.000000;0.000000;31.000000 size=42264B;;;0
```

**Comment by andreasbaumann** [comment]
Another different behaviour (where I'm not quite sure, what the correct behaviour should be):

```
./check_curl -fcurl -v -S -I xs02.eurospider.ch -H "www.eurospider.com" -p 443 -u "/" -r "Eurospider"
```

gives me:

```
HTTP CRITICAL: HTTP/1.1 400 Bad Request - pattern not found - 493 bytes in 0.188 second response time |time=0.187818s;;;0.000000;10.000000 size=493B;;;0
```

the error on server side is:

```
[Mon Nov 12 15:58:04 2018] [error] Hostname xs02.eurospider.ch provided via SNI and hostname www.eurospider.com provided via HTTP are different
```

So, this worked before with a hostname not matching the certificate, I get the feeling that either
xs02.eurospider.ch should be part of the certificate when -I xs02.eurospider.ch is used or that
-I and -H must both be www.eurospider.com.

This could be, because I changed the certificates to Truecrypt.

**Comment by sni** [comment]
`CONNECT` is handled because thats the legacy way to make proxy requests. Tbh, i fail to see how

    ./check_http -v -H www.openssl.org -C 14,7 -S -j CONNECT -u /

makes sense at all. The certificate check bails out after fetching the certificates, so no url is involved at all. Unless you want to check certificates over a proxy? Is thats what you want? But then still / makes no sense but something like:

  ./check_http -I 127.0.0.1 -p 3128 -u https://www.verisign.com/ -S -j CONNECT -H www.verisign.com -C 1,5

**Comment by andreasbaumann** [comment]
You're right, this example doesn't make any sense. :-)
Turns out this weas actually a simple certificate check in the end.

**Comment by sni** [comment]
This breaks the `-I` option. Seems like we have to bring back `CURLOPT_CONNECT_TO` or `CURLOPT_RESOLVE` option. I had hoped we don't need them, since the raise the libcurl version requirement.

**Comment by andreasbaumann** [comment]
Just got an Ngynx throwing the following at me:

```
HTTP CRITICAL HTTP/1.x 0 unknown - Unparsable status line in 0.725 seconds response time|time=0.725461s;;;0.000000;10.000000 size=2377B;;;0
```

last time I checked the RFC HTTP status line versioning had to use DIGIT and only DIGIT.

I have also seen with curl using HTTP/2:

```
HTTP/2 200
```

So, what should be the strategy here? Be far more tolerant when parsing the first
HTTP header line? Or have a switch?
I'm away HTTP/2 support is not there at all yet.. :-)

**Comment by winter1967** [comment]
Sorry, with the newest master, the configure doesn't know about --enable-check-curl:

`./configure --prefix=/usr/lib64/nagios/monitoring_plugins --enable-check-curl`

```
config.status: creating po/POTFILES
config.status: creating po/Makefile
configure: WARNING: unrecognized options: --enable-check-curl
```

**Comment by sni** [comment]
You need to checkout the `feature_check_curl` branch, its not in the master branch yet.

**Comment by GUI** [comment]
Thanks for working on this, this is great to have a checker backed by curl!

Just to note a few issues I ran into:

1. On the latest `feature_check_curl` branch (`34cc70c`), when running `./configure --enable-check-curl`, I still received a message saying `configure: WARNING: unrecognized options: --enable-check-curl`. It actually did enable and build `check_curl`, though, so it seems like this warning was a false positive.
2. `check_curl --version` reports the application being "check_http," which I wasn't sure was intentional and perhaps a bit confusing. The User-Agent remaining "check_http" I think is useful for compatibility purposes, but I wasn't sure if this `--version` check also needed to report that for any type of compatibility:
    ```
    $ check_curl --version
    check_http v2.2.git (monitoring-plugins 2.2)
    libcurl/7.64.0 OpenSSL/1.1.1c zlib/1.2.11 libidn2/2.0.5 libpsl/0.20.2 (+libidn2/2.0.5) libssh2/1.8.0 nghttp2/1.36.0 librtmp/2.3
    ```
3. When built against the version of curl in Debian Bionic, HTTP2 seems to be used by default where supported, and all those checks fail. HTTP/2 issues have already been noted here, but to provide a simple example against google.com:
    ```
    $ check_curl -H www.google.com -I www.google.com -S -v
    *   Trying 172.217.12.228...
    * TCP_NODELAY set
    * Expire in 4997 ms for 3 (transfer 0x55dd7cabea40)
    * Expire in 200 ms for 4 (transfer 0x55dd7cabea40)
    * Connected to www.google.com (172.217.12.228) port 443 (#0)
    * ALPN, offering h2
    * ALPN, offering http/1.1
    * successfully set certificate verify locations:
    *   CAfile: none
      CApath: /etc/ssl/certs
    * SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
    * ALPN, server accepted to use h2
    * Server certificate:
    *  subject: C=US; ST=California; L=Mountain View; O=Google LLC; CN=www.google.com
    *  start date: Aug 23 10:22:28 2019 GMT
    *  expire date: Nov 21 10:22:28 2019 GMT
    *  issuer: C=US; O=Google Trust Services; CN=GTS CA 1O1
    *  SSL certificate verify ok.
    * Using HTTP2, server supports multi-use
    * Connection state changed (HTTP/2 confirmed)
    * Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
    * Using Stream ID: 1 (easy handle 0x55dd7cabea40)
    > GET / HTTP/2
    Host: www.google.com
    User-Agent: check_http/v2.2.git (monitoring-plugins 2.2, libcurl/7.64.0 OpenSSL/1.1.1c zlib/1.2.11 libidn2/2.0.5 libpsl/0.20.2 (+libidn2/2.0.5) libssh2/1.8.0 nghttp2/1.36.0 librtmp/2.3)
    Accept: */*
    Connection: close

    * old SSL session ID is stale, removing
    * Connection state changed (MAX_CONCURRENT_STREAMS == 100)!
    < HTTP/2 200 
    < date: Fri, 06 Sep 2019 19:29:55 GMT
    < expires: -1
    < cache-control: private, max-age=0
    < content-type: text/html; charset=ISO-8859-1
    < p3p: CP="This is not a P3P policy! See g.co/p3phelp for more info."
    < server: gws
    < x-xss-protection: 0
    < x-frame-options: SAMEORIGIN
    < set-cookie: 1P_JAR=2019-09-06-19; expires=Sun, 06-Oct-2019 19:29:55 GMT; path=/; domain=.google.com; SameSite=none
    < set-cookie: NID=188=a-_qtpY5xysNPQtlD9IX3BNggagRFWJCltcWRXIr-iN-8fDzh8chbkRQFqYFbPPOUOPAH2T8kkYMqAOQu3_V3T7sbVHSGEY-cK5S73uIiIjMEAc71aq-S1IXlipRpiKLuVFZ7_XILb5Lwk5RLCqxxSv-xUavhA-saEQRZTTR8pU; expires=Sat, 07-Mar-2020 19:29:55 GMT; path=/; domain=.google.com; HttpOnly
    < alt-svc: quic=":443"; ma=2592000; v="46,43,39"
    < accept-ranges: none
    < vary: Accept-Encoding
    < 
    * Connection #0 to host www.google.com left intact
    HTTP CRITICAL HTTP/1.x 0 unknown - Unparsable status line in 0.301 seconds response time|time=0.300801s;;;0.000000;10.000000 size=12387B;;;0
    ```
    I'm not sure what the best solution here is, but it seems like adjusting the parsing of the status line would be needed to deal with these `HTTP/2 XXX` responses.

Thanks again for everyone's work on this!

**Comment by andreasbaumann** [comment]
1. --enable-check-curl

Is no longer required, as the plugin is always built now.
--enable-check-curl was eliminated and building the plugin is now triggered after checking
whether libcurl is available and has the right version (IIRC).

Somehow the check fails and check_curl is not building when I try it on Archlinux.
I'll have to have a look at this.

2. --version 

It should maybe report itself as 'check_curl' while it is coexisting with 'check_http'. Maybe
the idea was to replace 'check_http' completely.

3. Debian HTTP/2 per default

This is weird and needs investigation..

**Comment by andreasbaumann** [comment]
There is a: 

```
AM_CONDITIONAL([WITH_CHECK_CURL], [test "$_can_enable_check_curl" = "yes"])
```

check_curl is in EXTRA_PROGRAMS and might be set to an automake conditional?

which is then used nowhere?

When uriparser is not installed, then check_curl is silently not built (that was my problem on
Archlinux).

**Comment by andreasbaumann** [comment]
I spotted some problems in `curlhelp_parse_statusline` not being able to parse HTTP/2 (which for
some reasons is quite different than HTTP/1.x). As libcurl supports HTTP/2 we should support it.
There might be a need to provide a version switch (like for -S for SSL/TLS) to force a certain version
of the HTTP protocol (force 1.0, 1.1 or 2, nobody expected 0.9 to work anymore, I hope, and 3 is not
out of the door yet IIRC).

**Comment by andreasbaumann** [comment]
Fixed.

I'm not partiullarly happy with my now solution to set 

```
+#define HTTP_EXPECT "HTTP/"
```

to cover both HTTP/1.x and HTTP/2

**Comment by andreasbaumann** [comment]
f546041722b279d947f75523c52bd779ca6344dc also raises the questions:
- what should be the default HTTP version (CURL_HTTP_VERSION_NONE is the default and
  picks the "best possible")?
- Should we force HTTP/1.1 as default for backwards compatibility?
- Should we use HTTP/1.1 for HTTP and HTTP/2 for HTTPS for default?
- Should we allow auto-negotiation?

See also https://curl.haxx.se/libcurl/c/CURLOPT_HTTP_VERSION.html

**Comment by sni** [comment]
>     * what should be the default HTTP version (CURL_HTTP_VERSION_NONE is the default and
picks the "best possible")?
I'd go with that one. The hole idea of using curl was to outsource some of the work and logic to libcurl. So if there is code already which chooses the best one, i'd use that.
A switch to explicitly set some other value would be nice of course.

**Comment by andreasbaumann** [comment]
Good, that's what I implemented. :-)

**Comment by GUI** [comment]
Thanks so much! These new changes are working great for HTTP 2 now. (And just to clarify a typo in my original report, I was running into the HTTP/2 issues in Debian Buster, not "Bionic." But it sounds like you were able to make sense of everything and reproduce the issues.)

The only subtle issue I've run into now might be outside of the scope of `check_curl`, but since I encountered it in the context of the new HTTP 2 handling, I thought I'd mention it: The `--header-string` option checks for strings in the response header, but if you're using this to check for HTTP headers, the HTTP header names are checked in a case-sensitive fashion. This is consistent with how `check_http` behaves, but I'm wondering if this should be considered a bug, since HTTP field names should be compared in case insensitive fashion in all HTTP versions. See https://http2.github.io/http2-spec/#rfc.section.8.1.2:

> Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed

But as noted here, HTTP/2 now requires lowercase field names in the HTTP response. Since HTTP/2 proxies (like AWS CloudFront) lowercase any HTTP headers returned by the underlying server, I'm wondering if some of this case sensitivity might cause more issues now if you're trying to reuse checks across different HTTP versions.

For example, I had a checking using `--header-string "Location: https://example.com/"` for monitoring redirect behavior of some underlying servers that only support HTTP/1.1. I had previously been re-using that same check against the public load balanced URL that supports HTTP/2 (due to the load balancer proxy supporting HTTP/2). However, now that HTTP/2 is being used, that check will fail against the load balanced HTTP/2 version of the response, since it needs to be rewritten as `--header-string "location: https://example.com/"` due to `--header-string`'s case sensitivity. So then I need separate checks for the underlying server which is returning camel-cased headers and the HTTP/2 proxied endpoint which is returning lowercase headers.

Again, I'm not totally sure if this is within scope, since `--header-string` has historically been case sensitive in `check_http`. It also seems more complicated to try and detect the HTTP header syntax within the string and handle that according (since only the HTTP header field name should be compared case insensitively, while the HTTP header value should still be case sensitive). But since I imagine checking for HTTP headers is how `--header-string` is predominantly used, it does seem like comparing the header field names case-insensitively (as the HTTP spec indicates) might be ideal. It's certainly possible to work around this with more specific checks, but just thought I'd see if anyone else had thoughts on how this should be handled.

Thanks again!

**Comment by feuerrot** [comment]
Hi, we're currently evaluating check_curl as a replacement for check_http and found two "bugs", where check_curl behaves differently than check_http using the same command line options. Both of them disappear, if other options are used, so it's more a compatibility problem than a real bug.

# HEAD requests
`./check_http -H "mirror.example.org" -I "192.0.2.77" -e 200 -j "HEAD" -p 80 -u /RPM-GPG-KEY-EPEL`
We've got some checks where we use a HEAD request. The server transmits a content-size header but no content and check_curl outputs `HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 18 - Transferred a partial file` due to the missing content.
This probably happens because we do not use the option `-N` together with `-j HEAD`, so `-j HEAD` should probably imply `-N`?

# redirect to a relative URL
`./check_curl -H "cms.example.org" -I "192.0.2.235" -c 20 -f "sticky" -p 8444 -s "Anmeldung" -u /example/login -w 10`
check_curl and check_http behave differently if there is a redirect, `-f sticky` is used and the redirect target is a relative URL:
check_http [reuses](https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/check_http.c#L1417) the port, check_curl doesn't and uses the default one instead.
I suppose a relative URL as redirect target should always cause the reuse of scheme, host and port of the previous request.

**Comment by andreasbaumann** [comment]
Yes, agreed, on HEAD we should definitely not wait for a body (set curl option CURLOPT_NOBODY).
I fixed that accordingly.

**Comment by feuerrot** [comment]
I found another bug today :)

# Hostname resolving with a fixed IP and TLS
```
./check_curl -I 192.0.2.150 -H "example.org" -S -c 20 -e 200 -p 19420 -s "teststring" -u / -w 10 -vvv
* Set SSL/TLS version to 0
* Socket timeout set to 21 seconds
* curl CURLOPT_URL: https://example.org:19420/
* SSL library string is : NSS/3.36 NSS (4)
* About to connect() to example.org port 19420 (#0)
*   Trying 93.184.216.34...
```
check_http resolves (or tries to resolve) the hostname even if we supply an IP address. This does not happen without TLS:
```
./check_curl -I 192.0.2.150 -H "example.org" -c 20 -e 200 -p 19420 -s "teststring" -u / -w 10 -vvv
* Socket timeout set to 21 seconds
* curl CURLOPT_URL: http://192.0.2.150:19420/
* SSL library string is : NSS/3.36 NSS (4)
* About to connect() to 192.0.2.150 port 19420 (#0)
*   Trying 192.0.2.150...
* Connected to 192.0.2.150 (192.0.2.150) port 19420 (#0)
```

If I revert [this](https://github.com/monitoring-plugins/monitoring-plugins/commit/faea5899ba3264581bf75649e4b399d0b69bd125) commit, check_curl connects to the supplied IP again. Maybe [CURLOPT_RESOLVE](https://curl.haxx.se/libcurl/c/CURLOPT_RESOLVE.html) can be used to fix the original problem with SNI?

**Comment by andreasbaumann** [comment]
What about CURLOPT_SNI_HOSTNAME? CURLOPT_RESOLVE is more for a pre-populated DNS cache IMHO. What I want to make sure is that the client in the TLS hello and the virtual hostname are
correctly set.

**Comment by feuerrot** [comment]
According to https://github.com/curl/curl/issues/1775#issuecomment-322352214 `CURLOPT_SNI_HOSTNAME` was abandoned and at least in 2015 Daniel Stenberg suggested using `CURLOPT_RESOLVE` for this usecase [here](https://curl.haxx.se/mail/archive-2015-01/0042.html).

**Comment by andreasbaumann** [comment]
I fear, that's not our use case. As I see it, we need:
- a host/ip to connect to (this can be hepled with  CURLOPT_RESOLVE, which is sort of a /etc/hosts
  file)
- a virtual host, where the web server decides what server-side config to pick (the Host: header)
- a SNI hostname in the TLS hello, which in my understand is needed to pick the right
  configuration (for instance certificate) before going into virtual hosting.

I agree that just setting the hostname in the URL (as in my commit) is wrong.

**Comment by feuerrot** [comment]
Are there actually use cases, where the virtual host is different to the SNI hostname? If there are none, `CURLOPT_RESOLVE` would be sufficient:
* If the user specifies only a hostname xor an IP: Use it in `CURLOPT_URL`
* If the user specifies a hostname and an IP: Use the hostname in `CURLOPT_URL` and add `CURLOPT_RESOLVE`

If there is an use case, Daniel gives another example [in his blog](https://daniel.haxx.se/blog/2018/04/05/curl-another-host/):
> Connect to a HTTPS host running on localhost, use the correct name for SNI and certificate verification, but then still ask for a separate host in the Host: header? Sure, no problem:
> `curl --resolve example.com:443:127.0.0.1 https://example.com/ --header "Host: diff.example.com"`

The `--libcurl` parameter generates code which uses `CURLOPT_URL` for the specified hostname (`example.com`), `CURLOPT_HTTPHEADER` for the virtual hostname (`diff.example.com`) and `CURLOPT_RESOLVE` for the IP (`127.0.0.1`) to connect to.

**Comment by andreasbaumann** [comment]
Reverting:

```
 use_ssl ? host_name : server_address,
```
results in my brain more to something like:

```
 curl --resolve example.com:443:45.33.24.55 https://45.33.24.55/ --header "Host: example.com"
```

The certificate is tested against 45.33.24.55, so I get:

```
curl: (60) SSL: no alternative certificate subject name matches target host name '45.33.24.55,
```

**Comment by GUI** [comment]
I think I've hit some bugs with the `--onredirect=follow` implementation. However, I did discover the new `--onredirect=curl` option, which seems to work properly, but just thought I'd report the issues with the legacy `follow` option.

Here's a demonstration against https://httpbin.org/relative-redirect/1, which redirects to https://httpbin.org/get, where "User-Agent" should be in the response body:

```sh
check_curl -H httpbin.org -u '/relative-redirect/1' --ssl --onredirect=follow -s 'User-Agent'
# HTTP CRITICAL: HTTP/1.1 200 OK - string 'User-Agent' not found on 'http://httpbin.org:80' - 9827 bytes in 0.090 second response time |time=0.089793s;;;0.000000;10.000000 size=9827B;;;0
```

Note that it thinks the response was a 200 OK, but the URL doesn't seem correct and it didn't find the expected string in the response body. Instead of redirecting to `/get`, it appears to be redirecting to `/` (which response body checks seem to verify, since it is able to find content in the body that's present on that root page).

A second issue that might be separate is that if you try to run this request with `--verbose` to debug things, it results in a segmentation fault:

```sh
check_curl -H httpbin.org -u '/relative-redirect/1' --ssl --onredirect=follow -s 'User-Agent' --verbose
# ...
# * Connection #0 to host httpbin.org left intact
# Segmentation fault
```

Running with Valgrind indicates maybe this line is triggering the segfault in verbose mode: https://github.com/monitoring-plugins/monitoring-plugins/blob/adb958849491ea7e0bba4a15896ebd41a0fbff2d/plugins/check_curl.c#L1000

```
==14046== Process terminating with default action of signal 11 (SIGSEGV)
==14046==  Access not within mapped region at address 0x3
==14046==    at 0x4839418: strncmp (vg_replace_strmem.c:648)
==14046==    by 0x110FBE: redir (check_curl.c:1000)
==14046==    by 0x110756: check_http (check_curl.c:787)
==14046==    by 0x10B97C: main (check_curl.c:272)
```

But again, the redirect behavior does seem to be working well with `--onredirect=curl` option, so this just seems specific to the legacy implementation in the `redir` function.

andreasbaumann:
Hi, thanks for testing.

On the first issue: check_curl does a switch to HTTP/2, hence the 'Location' gets a 'location'
and 'User-Agent' a 'user-agent' (who ever had this brilliant idea to define HTTP/2 to be
lowercase).

The following request works as expected:

```./check_curl --http-version=1.1 -H httpbin.org -u '/relative-redirect/1' --ssl --onredirect=follow -s 'User-Agent'
HTTP OK: HTTP/1.1 200 OK - 619 bytes in 0.550 second response time |time=0.549536s;;;0.000000;10.000000 size=619B;;;0
```

This makes me think that:

1. We should maybe make HTTP/1.1 the default as it fits closer to the behaviour of chec_http
2. We should provide means to do checks for header fields based on the HTTP-version used
   (not only 'Location' vs. 'location' will be an issue I'm afraid)
3. String and regex compares on HTTP header fields should either automatically switch based
   on the HTTP version or there should be a way to do case-sensitive and case-insensitive
  header checks (-d and -D analog to -r and -R)

sni:
i added the CURLOPT_RESOLVE entry to solve the SNI issue. Looks promising so far.

sni:
Tests are green. Any objections to merge this into master? 
It should not have any sideeffects, since it is a new and separate plugin.

weiss:
I'm all for merging this. Many thanks for all the work that went into it.

sni:
Finally merged it.
Thanks to all involved in this PR.

PR #1538, opened 2018-06-03
Uptime plugin

wopfel:
Hi,
as mentioned some days ago, I've created a new plugin. And here it is - my check plugin for the uptime of a system. Would be great if you could add this to the monitoring-plugins package.

Basic checks can be done with -w and -c parameters, generating a warning or critical state if the uptime is higher than the -w or -c value (usual stuff). Both parameters are read as seconds, but you can suffix them with m (minutes), h (hours), ...
Besides that, you can get a nice formatted output if you provide the -f (Running for ...) or -s (Running since ...) switch.

I've added a test file, too. Hopefully the system where the tests are executed is running longer than 2 seconds and less than about 190 years :)

Feedback is much appreciated.

Kind regards,
Bernd

## Examples:
```
# /usr/local/bin/check_uptime -w 10 -c 11      
CRITICAL: uptime is 5005860 seconds. |uptime=5005860;10;11;0
# /usr/local/bin/check_uptime -w 10 -c 11 -f
CRITICAL: uptime is 5005862 seconds. Running for 8 weeks, 1 day, 22 hours, 31 minutes and 2 seconds. |uptime=5005862;10;11;0
# /usr/local/bin/check_uptime -w 10 -c 11 -s
CRITICAL: uptime is 5005864 seconds. Running since 2018-03-30 14:58:53. |uptime=5005864;10;11;0
# /usr/local/bin/check_uptime -w 10 -c 11 -s -f
CRITICAL: uptime is 5005871 seconds. Running for 8 weeks, 1 day, 22 hours, 31 minutes and 11 seconds. Running since 2018-03-30 14:58:53. |uptime=5005871;10;11;0
```


plugins-scripts/check_uptime.pl
```diff
@@ -0,0 +1,256 @@
+#!/usr/bin/perl -w
```
sni: please use: `#!@PERL@ -w`

plugins-scripts/check_uptime.pl
```diff
@@ -0,0 +1,256 @@
...
+# Replace last occurence of comma with "and"
+$pretty_uptime =~ s/, $/ and /;
+# Always print the seconds (though it may be 0 seconds)
+$pretty_uptime .= sprintf( "%d second%s", $secs, $secs == 1 ? "" : "s" );
+
+
+# Default to catch errors in program
+my $state_str = "UNKNOWN";
+
+# Check values
```
sni: It would be nice to have optional range support like in https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins-scripts/check_file_age.pl#L98

plugins-scripts/check_uptime.pl
wopfel: Thanks for the hint. I've changed that line by accident in the previous commit.

wopfel:
Hi Sven, thanks for your remarks. I've thought about ranges when writing this plugin and came to the conclusion that other checks might step in when they fail during the reboot. I thought there's no need for ranges.
Anyhow, the range parameter was no big deal. And there's no need for an additional Perl module to have range support.
Your opinion so far? If my changes fit, then I'll modify the help text, too.

sni:
I made two minor changes:
 - adding the script to the makefile, so it will be build properly
 - replacing a weird space character with an actual space.

Looks good to me.

wopfel:
Hi Sven, I think the feature is complete now. What do you think? Thanks for your feedback and your commits so far!
Kind regards,
Bernd

PR #1542, opened 2018-07-07
check_snmp: add 'multiplier' to modify current value

wnieder:
Sometimes the value returned by check_snmp should be modified to represent the actual value (e.g. an SNMP value of 1920 representing a temperature value of 19,2 °C). To avoid doing the modification in the next step I added an option (-M multiplier) to specify a float number. The current value in output and performance data is multiplied by this value which defaults to 1.0. Specifying values 0 < n < 1 will act as a divisor.

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -1135,7 +1135,11 @@ multiply (char *str)
 {
 	double val = strtod (str, NULL);
 	val *= multiplier;
-	sprintf(str, "%f", val);
+	if (val == (int)val) {
+		sprintf(str, "%d", val);
+	} else {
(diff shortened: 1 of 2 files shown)


sni:
I like the idea, but right now it turns all integers into floats. So i'd suggest to only apply the multiplier if its not 1. Also it would be nice to have a test case.

canhastrytocode:
@wnieder Please finish this, i beg you :>

sni:
could you please rebase onto latest head, travis tests should be much more stable now.

RincewindsHat:
was replaced  by #1813

PR #1543, opened 2018-07-08
timeout_alarm_handler(): return UNKNOWN on timeout

Al2Klimov:


Al2Klimov:
* [x] unit tests
* [x] master rebase

sni:
while i agree with you, that a timeout should return unknown. I prefer a broader approach which would make the exit code configurable, like ex.: `./check_... -t 2:UNKNOWN` or something like that.

Al2Klimov:
Hello @sni and thank you for your feedback!

If you agree, please approve this. IMO the configurability (if any) should be another PR.

Best,
AK

PS: Please see also #1494.

sni:
I agreed that it should be unknown but i don't think its a good idea to change that globally now. If i would approve anything in that direction, it would be the solution mentioned above with the configurable exit code.

waja:
@Al2Klimov are you still interested into get this integrated? If yes, @sni would be satisfied for shure if you change this accordingly https://github.com/monitoring-plugins/monitoring-plugins/pull/1543#issuecomment-403495471

Al2Klimov:
I'm afraid:

PR #1545, opened 2018-07-25
check_dns: improve support for checking multiple addresses

DerDakon:
Improve the handling of multiple IPs in check_dns: my use case is a set of hosts with 2 addresses each (IPv4 + IPv6), which I don't want to split into "IPv4 is smaller in string compare than IPv6" and "the others". And I want to check that the exactly the addresses I have configured are returned, no more, no less.

This more or less intentionally breaks the previously working "pass addresses as comma separated string and compare that". The combination of -L and multiple -a does the same without breaking randomly on every IP change. I don't care. YMMV.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,7 @@ This file documents the major additions and syntax changes between releases.
 	check_dns: allow 'expected address' (-a) to be specified in CIDR notation
 	  (IPv4 only).
 	check_dns: allow for IPv6 RDNS
+	check_dns: allow unsorted addresses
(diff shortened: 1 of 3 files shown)


DerDakon:
Ping?

DerDakon:
Ping?

DerDakon:
Ping?

DerDakon:
ping?

DerDakon:
ping?

sni:
thanks and merged. Sorry, we are all somehow quite busy.

PR #1547, opened 2018-09-12
Add ignored -H/--host flag to negate

RaymiiOrg:
LibreNMS services automatically add -H <hostname> to every service check, no option to ignore this. My HTTP check returns a 401, but that is an OK status. Adding this -H parameter to be ignored allows the
command to succeed with this parameter.


sni:
this should be fixed in librenms imho then

PR #1548, opened 2018-09-20
check_load: reports wrong load average using -r option

Teszko:
Hello,
check_load -r divides the load average by the number of cpu cores. The scaled values are then used in the output, which leads to interesting notices such as: 
> WARNING - load average: 0.08, 0.10, 0.06

While in fact the load was 72 time higher. 

I got this warning from Nagios when using the default nrpe config on ubuntu
> command[check_load]=@pluginsdir@/check_load -r -w .15,.10,.05 -c .30,.25,.20

The first commit fixes this problem, the second makes the file have uniform indents instead of mixed tabs and spaces. 
Have a nice day!

### DIFF
--- a/plugins/check_load.c
+++ b/plugins/check_load.c
@@ -155,16 +155,16 @@ main (int argc, char **argv)
 		printf (_("Could not open stderr for %s\n"), PATH_TO_UPTIME);
 	}
 	fgets (input_buffer, MAX_INPUT_BUFFER - 1, child_process);
-    if(strstr(input_buffer, "load average:")) {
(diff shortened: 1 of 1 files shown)


Teszko:
the failed test seems to be for an unrelated plugin.

RincewindsHat:
ok, the changes seem sensible, but I would like the question, if the output should show the unscaled load, when the `-r` option is used. This seems to be confusing, to be honest.

RincewindsHat:
Implemented by  #1778

PR #1560, opened 2018-11-09
Add new flag --show-body/-B to print body

### LINKED ISSUE #1559: check_http: Optionally return content
When using check_http on an API with a healthz endpoint, this is basically useless:

image

I'd like to see the content that the HTTP request response is returning. It's actually stating that an API on the internet that it is using is down. So I can't do anything and can ignore this problem.

towolf:
This should help with figuring out ia problem at a glance when enabled for
healthz endpoints on web APIs, for example.

The content of the body can point to what the problem is and help with
diagnostics.

Fixes #1559

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -120,6 +120,7 @@ int use_ssl = FALSE;
 int use_sni = FALSE;
 int verbose = FALSE;
 int show_extended_perfdata = FALSE;
+int show_body = FALSE;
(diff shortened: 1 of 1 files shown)


towolf:
@sni @lazyfrosch PTAL?

rossnick:
This PR seems to have been merged, but never released. Current versions of the check_http don't have this option.

PR #1563, opened 2018-11-29
check_icmp: Add IPv6 support

### LINKED ISSUE #1291: check_icmp and IPv6

jacobbaungard:
https://github.com/monitoring-plugins/monitoring-plugins/pull/1384 got stuck some while ago, and needed to be rebased on the latest master to pass the tests. The original branch for the PR is gone, so had to create a new one.

All credits to @LarsMichelsen 

> This commit adds IPv6 capabilities to check_icmp. It is now possible to
> specify the address family using the arguments -4 (default) or -6.
> 
> To make the change possible we had to move the argument parsing previous
> to creating the socket to be able to create it with the correct address
> family.
> 
> This commit also fixes some gcc 4.9.2 compiler warnings. It has been
> tested with several current linux distributions (debian, ubuntu, rh,
> sles).
> 
> This commit fixes #1291

### DIFF
--- a/plugins-root/check_icmp.c
+++ b/plugins-root/check_icmp.c
@@ -410,6 +410,7 @@ main(int argc, char **argv)
 #ifdef SO_TIMESTAMP
 	int on = 1;
 #endif
+	char * opts_str = "vhVw:c:n:p:t:H:s:i:b:I:l:m:64";
 
 	setlocale (LC_ALL, "");
 	bindtextdomain (PACKAGE, LOCALEDIR);
@@ -462,7 +463,7 @@ main(int argc, char **argv)
 
 	/* Parse protocol arguments first */
 	for(i = 1; i < argc; i++) {
-		while((arg = getopt(argc, argv, "vhVw:c:n:p:t:H:s:i:b:I:l:m:64")) != EOF) {
+		while((arg = getopt(argc, argv, opts_str)) != EOF) {
 			unsigned short size;
 			switch(arg) {
 			case '4':
@@ -484,7 +485,7 @@ main(int argc, char **argv)
 
 	/* parse the arguments */
 	for(i = 1; i < argc; i++) {
-		while((arg = getopt(argc, argv, "vhVw:c:n:p:t:H:s:i:b:I:l:m:64")) != EOF) {
+		while((arg = getopt(argc, argv, opts_str)) != EOF) {
 			unsigned short size;
 			switch(arg) {
 			case 'v':
--- a/plugins-root/check_icmp.c
+++ b/plugins-root/check_icmp.c
@@ -67,7 +67,9 @@ const char *email = "devel@monitoring-plugins.org";
 #include <netinet/in_systm.h>
 #include <netinet/in.h>
 #include <netinet/ip.h>
+#include <netinet/ip6.h>
 #include <netinet/ip_icmp.h>
+#include <netinet/icmp6.h>
 #include <arpa/inet.h>
 #include <signal.h>
 #include <float.h>
@@ -113,8 +115,8 @@ typedef struct rta_host {
 	unsigned short id;           /* id in **table, and icmp pkts */
 	char *name;                  /* arg used for adding this host */
 	char *msg;                   /* icmp error message, if any */
-	struct sockaddr_in saddr_in; /* the address of this host */
-	struct in_addr error_addr;   /* stores address of error replies */
+	struct sockaddr_storage saddr_in;     /* the address of this host */
+	struct sockaddr_storage error_addr;   /* stores address of error replies */
 	unsigned long long time_waited; /* total time waited, in usecs */
 	unsigned int icmp_sent, icmp_recv, icmp_lost; /* counters */
 	unsigned char icmp_type, icmp_code; /* type and code from errors */
@@ -139,6 +141,18 @@ typedef struct icmp_ping_data {
 	struct timeval stime;	/* timestamp (saved in protocol struct as well) */
 	unsigned short ping_id;
 } icmp_ping_data;
+
+typedef union ip_hdr {
+	struct ip ip;
+	struct ip6_hdr ip6;
+} ip_hdr;
+
+typedef union icmp_packet {
+	void *buf;
+	struct icmp *icp;
+	struct icmp6_hdr *icp6;
+	u_short *cksum_in;
+} icmp_packet;
 
 /* the different modes of this program are as follows:
  * MODE_RTA: send all packets no matter what (mimic check_icmp and check_ping)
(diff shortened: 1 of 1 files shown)


plugins-root/check_icmp.c
```diff
@@ -1160,31 +1318,58 @@ add_target_ip(char *arg, struct in_addr *in)
...
+
+	switch (address_family) {
+	case -1:
+		// -4 and -6 are not specified on cmdline
+		address_family = AF_INET;
+		sin = (struct sockaddr_in *)&ip;
+		result = inet_pton(address_family, arg, &sin->sin_addr);
+#ifdef USE_IPV6
+		if( result != 1 ){
+			address_family = AF_INET6;
```
sni: address_family should be a local copy here. Otherwise address resolving might fail later. Consider this example:
`./check_icmp -H localhost -4`
address_family is -1 when entering add_target() but will be set to AF_INET6 during testing with inet_pton(). Then later getaddrinfo() uses AF_INET6 to resolve the name but fails because its AF_INET.

sni:
I guess we should first iterate the arguments to set the adress_familiy and then parse the rest of the arguments. Althought this would make it impossible to use both ipv4 and ipv6 at the same time, but i doubt its possible right now.
Something like
`./check_icmp -4 -H ipv4host -6 -H ipv6host` compared to
`./check_icmp -H ipv4host -4`

sni:
seems like travis does not run tests from the plugins-root folder. If i run them locally, they fail.
Could you have a look at the t/check_icmp.t please until #1565 is merged.

```
Testing:  ./check_icmp -H localhost -H 10.99.0.1 -w 10000ms,100% -c 10000ms,100% -n 1 -m 1
not ok 13 - One of two host nonresponsive - one required
#   Failed test 'One of two host nonresponsive - one required'
#   at t/check_icmp.t line 77.
#          got: '3'
#     expected: '0'
not ok 14 - Output OK
#   Failed test 'Output OK'
#   at t/check_icmp.t line 78.
#                   'check_icmp: Failed to resolve 10.99.0.1: Address family for hostname not supported'
#     doesn't match '/OK - .*?: rta (?:[\d\.]+ms)|(?:nan), lost \d+%/'
Testing:  ./check_icmp -H localhost -H 10.99.0.1 -w 10000ms,100% -c 10000ms,100% -n 1 -m 2
not ok 15 - One of two host nonresponsive - two required
#   Failed test 'One of two host nonresponsive - two required'
#   at t/check_icmp.t line 83.
#          got: '3'
#     expected: '2'
not ok 16 - Output OK
#   Failed test 'Output OK'
#   at t/check_icmp.t line 84.
#                   'check_icmp: Failed to resolve 10.99.0.1: Address family for hostname not supported'
#     doesn't match '/(WARNING|CRITICAL) - .*?: rta [\d\.]+ms, lost \d%/'
# Looks like you failed 4 tests of 16.

```

jacobbaungard:
Those should be fixed now, but I guess we should have a few more tests to test IPv6 functionality, and the -4 / -6 flags.

jacobbaungard:
Rebased on latest master to run the tests.

sni:
thanks, meanwhile #1565 has been merged, so at least we get some travis tests as well. I noticed one thing.

This should work:
```
sudo ./check_icmp -H localhost -H 192.168.123.253 -n 1 -m 1 
CRITICAL - localhost: rta nan, lost 100% :: 192.168.123.253: rta nan, lost 100%...
```

Interestingly, if i switch the hosts, it works. However the order of the host should not make a difference:

```
%> sudo ./check_icmp -H 192.168.123.253 -H localhost -n 1 -m 1 
OK - 192.168.123.253: rta nan, lost 100% :: localhost: rta 0.018ms, lost 0%
```

jacobbaungard:
Does it work with -4 ? I can't replicate that in my setup.

sni:
no it doesn't, btw. the `192.168.123.253` is a random unavailable ip in that case.

jacobbaungard:
Strange, I seem to get the same behavior on both master and this branch.

```
./check_icmp -H localhost -H 192.168.123.253 -n 1 -m 1 
OK - localhost: rta 0.032ms, lost 0% :: 192.168.123.253: rta nan, lost 100%|localhostrta=0.032ms;200.000;500.000;0; localhostpl=0%;40;80;; localhostrtmax=0.032ms;;;; localhostrtmin=0.032ms;;;; 192.168.123.253rta=0.000ms;200.000;500.000;0; 192.168.123.253pl=100%;40;80;; 192.168.123.253rtmax=0.000ms;;;; 192.168.123.253rtmin=0.000ms;;;; 
```

plugins-root/check_icmp.c
```diff
@@ -456,20 +460,31 @@ main(int argc, char **argv)
...
-
-	/* support "--help" and "--version" */
-	if(argc == 2) {
-		if(!strcmp(argv[1], "--help"))
-			strcpy(argv[1], "-h");
-		if(!strcmp(argv[1], "--version"))
-			strcpy(argv[1], "-V");
+	/* Parse protocol arguments first */
+	for(i = 1; i < argc; i++) {
+		while((arg = getopt(argc, argv, "vhVw:c:n:p:t:H:s:i:b:I:l:m:64")) != EOF) {
```
sni: it feels like we should put the opts string into a variable, otherwise it will soon drift apart.

plugins-root/check_icmp.c
```diff
@@ -1160,31 +1323,59 @@ add_target_ip(char *arg, struct in_addr *in)
...
 		return add_target_ip(arg, &ip);
-		/* he = gethostbyaddr((char *)in, sizeof(struct in_addr), AF_INET); */
-		/* if(!he) return add_target_ip(arg, in); */
 	}
 	else {
 		errno = 0;
-		he = gethostbyname(arg);
-		if(!he) {
+		memset(&hints, 0, sizeof(hints));
+		hints.ai_family = AF_UNSPEC;
```
sni: AF_UNSPEC should only be used if `address_family` is -1. Otherwise we will set address_familiy to v6 even if -4 is specified on the command line.
ex.: `./check_icmp -H labs.consol.de -4 -v` uses the ipv6 address.

plugins-root/check_icmp.c
sni: i really think we should do something like this here:
```
result = inet_pton(AF_INET, arg, &sin->sin_addr);
if(result == 0)
    address_family = AF_INET;
```

plugins-root/check_icmp.c
ghciv6: the commandline interface seams not self explained.

one time you requested 
   ` ./check_icmp -4 -H ipv4host -6 -H ipv6host`
to work. the position of -4 and -6 is important.

in case of 
  `./check_icmp -H localhost -4`
you expected `-4` should change the name resolution of `localhost`. but if you want position dependend parameter handling, this call is not valid.

just a theoretical question. how about:
`./check_icmp -H localhost -4 -H ipv4host -6 -H ipv6host`
would you expect an IPv4 or IPv6 check for localhost? If localhost has a dual stack, i'm expecting localhost and  ipv6host are checked with IPv6 and ipv4host with IPv4. localhost should be checked by IPv6 because the OS would also prefer IPv6 over IPv4.

i don't have such complex use cases. i only need a working auto detection for any IPv4 and IPv6 address without specifying -4 or -6. if an address is specified, there is no option for interpretation. so -4 and -6 are redundant.
`./check_icmp ::1`
`./check_icmp 127.0.0.1`

sni: right now, it is not possible to mix ipv4 and ipv6 addresses. Maybe we should emit an error if somebody tries too.

jacobbaungard:
Attempted to address all the feedback with the new commits. Let me know what you think.

jacobbaungard:
Any news on this one?

sni:
thank you very much

PR #1564, opened 2018-11-29
check_cluster.c: Added data argument validation.

jacobbaungard:
This commit adds validation on the -d argument to ensure that valid Nagios status codes are are supplied (i.e 0,1,2,3).


sni:
thanks

PR #1567, opened 2018-12-08
[check_disk] add support to display inodes usage in perfdata

vdanjean:
This is not enabled by default
It can be enabled with the -P (--iperfdata) option

Of cause, option names can be changed, and it can be enabled by default if you think it would be a good thing.

Regards,
  Vincent


sni:
thanks

PR #1569, opened 2019-01-04
check_hpjd: Added -D option to disable warning on 'out of paper'

jacobbaungard:
Many organisations let the users refill the papers. so they are not interested if a printer is "out of paper". This commit adds a flag, -D that disables warnings on this particular printer status.


sni:
i triggered a rebuild, the check_snmp travis test fails all the time... maybe we should disable it till it works properly

sni:
thanks, merged.

PR #1577, opened 2019-01-21
check_by_ssh: fix child process leak on timeouts

sni:
When check_by_ssh runs into a timeout it simply exits keeping all child processes running.
Simply adopting the kill loop from runcmd_timeout_alarm_handler() fixes this.
In order to do this, i had to move some code into the common header. On the other hand, this
eliminates some duplicate code.

### DIFF
--- a/lib/utils_cmd.c
+++ b/lib/utils_cmd.c
@@ -40,6 +40,7 @@
 
 /** includes **/
 #include "common.h"
+#include "utils.h"
 #include "utils_cmd.h"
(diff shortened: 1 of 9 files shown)


waja:
@sni do you Think that is something that fit's for the maint-2.2 branch?

sni:
hmm, not sure, there might have been some follow up fixes to this. So it would be some work to go through the other commits and see if they are required as well. Might be easier and less work to simply do a new release.

PR #1587, opened 2019-03-21
check_mysql: Allow sockets to be specified to -H

jacobbaungard:
The help text says that -H accepts a "unix socket (must be an absolute
path)". Now that actually corresponds to reality.

Signed-off-by: Robin Sonefors <robin.sonefors@op5.com>


sni:
thanks.

PR #1589, opened 2019-03-29
check_snmp segfaults

adrb:
check_snmp will segfaults at line 489 if number of lines returned by SNMPD is greater than number of defined thresholds


adrb:
I messed up segfaulting line number, actually its [470](https://github.com/monitoring-plugins/monitoring-plugins/blob/75af0140b5f5306695eb17257cf1835df2f8cfa8/plugins/check_snmp.c#L470)

`		/* Process this block for numeric comparisons */
		/* Make some special values,like Timeticks numeric only if a threshold is defined */
		if (thlds[i]->warning || thlds[i]->critical || calculate_rate) {`

RincewindsHat:
Sorry to bring this up again this late, but is your fix still vallid and the problem still there?
Is there a way to provoke this behaviour to test it?

adrb:
Yes, it's still there, but currently I don't remember how to test it. Are you interested in fixing it?

waja:
> Yes, it's still there, but currently I don't remember how to test it. Are you interested in fixing it?

I think this was @RincewindsHat intention.

RincewindsHat:
indeed, sorry for not answering. I would like to have to a way consistently trigger the bug

adrb:
I needed to dig into that again, but finally come up with something that you should easily reproduce:

Example test on CentOS7 with disabled selinux (just to make it easier):

```
# printf "1\noid2 = 2\noid3 = 3\noid4 = 4\noid5 = 5\n" > /tmp/snmp_test
# echo "extend test /bin/cat /tmp/snmp_test" >> /etc/snmp/snmpd.conf
# systemctl restart snmpd
# /usr/bin/snmpget -Le -t 3 -r 5 -m ALL -v 1 -c public localhost:161 NET-SNMP-EXTEND-MIB::nsExtendOutputFull.\"test\"
NET-SNMP-EXTEND-MIB::nsExtendOutputFull."test" = STRING: 1
oid2 = 2
oid3 = 3
oid4 = 4
oid5 = 5
# /usr/lib64/nagios/plugins/check_snmp -H localhost -C public -o NET-SNMP-EXTEND-MIB::nsExtendOutputFull.\"test\"
Segmentation fault
```

It's because default OID line delimiter is '=', a so number of lines is higher than number of defined oids and thus threshold array.

```
# gdb /usr/lib64/nagios/plugins/check_snmp
...
(gdb) run -H localhost -C public -o NET-SNMP-EXTEND-MIB::nsExtendOutputFull.\"test\"
...
Program received signal SIGSEGV, Segmentation fault.
0x00005555555573eb in main (argc=7, argv=<optimized out>) at check_snmp.c:496
496                     if (thlds[i]->warning || thlds[i]->critical || calculate_rate || is_ticks || offset != 0.0) {
(gdb) list
491
492                     iresult = STATE_DEPENDENT;
493
494                     /* Process this block for numeric comparisons */
495                     /* Make some special values,like Timeticks numeric only if a threshold is defined */
496                     if (thlds[i]->warning || thlds[i]->critical || calculate_rate || is_ticks || offset != 0.0) {
497                             ptr = strpbrk (show, "-0123456789");
498
499                             if (ptr == NULL)
500                                     die (STATE_UNKNOWN,_("No valid data returned (%s)\n"), show);
(gdb) print i
$1 = 1
(gdb) print thlds[i]
$2 = (thresholds *) 0x0
```

RincewindsHat:
Note: for debian one has to also modify the permissions to reproduce the given example:
```diff
- view   systemonly  included   .1.3.6.1.2.1.1
+ view   systemonly  included   .1.3.6.1
```
or something like that

RincewindsHat (approved):
Looks good to me.

RincewindsHat:
Thank you very much for the contribution and sorry for the delay

PR #1591, opened 2019-04-09
include -P switch in help

odenbach:
Hi,

#1567 added support for including inode usage in perfdata (which is awesome). This small change adds the missing inline documentation.

Thanks,

Christopher


sni:
thanks

PR #1601, opened 2019-08-01
check_procs: support of light weight processes (threads)

bkuebler:
To use check_procs and it's mechanism to check threads on systems instead of processes. With this feature you can use "-n" flag to switch to threads overall instead of processes.

If you would like to use threads (lwp) in ps so calls NLWP (number of light weight processes) as metric now you can use "-m NLWP". For example if you use this you can check if the given process or any of the checked processes are reach the thread threshold defined with -c or -w

### DIFF
--- a/THANKS.in
+++ b/THANKS.in
@@ -426,3 +426,4 @@ Eunice Remoquillo
 Louis Sautier
 Sven Hartge
 Alvar Penning
+Benjamin Kübler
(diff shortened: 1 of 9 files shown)


bkuebler:
@waja why is it closed? I would like to get any feedback.

waja:
@bkuebler sorry ... I accidentally closed it by https://github.com/monitoring-plugins/monitoring-plugins/pull/1663

bkuebler:
@waja i updated my PR to the current version to resolve all conflicts

bkuebler:
JFTR I try to complete all open issues within the next 3 weeks. 

- Test cases for new features
- Fixing issue with current tests due to incorrect test file
- metric bases check currently not working as expected only threads instead of processes switch works properly.

waja:
> JFTR I try to complete all open issues within the next 3 weeks.

Take your time!

PR #1602, opened 2019-08-02
Allow mariadbclient to be used for check_mysql

tonvoon:
This allows mariadbclients to be used for check_mysql.

The previous ./configure failed because it was trying to link mysqlclient (from configure.ac) and mariadbclient (from mysql_config). This change will initially attempt to use mysqclient and then fallback to mariadbclient.

Tested on Ubuntu18 with mysqlclient installed, mariadbclient installed and neither installed.


tonvoon:
D'oh! You're right - that file wasn't meant to be there. I've deleted the file in the branch now

sni:
thanks

Issue #1142, opened 2013-09-24
check_ntp_time and check_ntp_peers cannot find offset [sf#3314686]


monitoring-user:
Submitted by silfreed on 2011-06-10 14:56:22

I'm having a similar problem with the "offset unknown" that others have had in the past, although I'm running 1.4.14 of the plugins using check_ntp_time and check_ntp_peer.  Below is the output of some of the commands.  For some reason check_ntp_peer doesn't even seem to be able to find the peers, although check_ntp_time seems fine, but doesn't get the correct offset.

```
# ntpdc -n -s localhost
     remote           local      st poll reach  delay   offset    disp
=======================================================================
 208.53.158.34   10.128.0.45      2   64    3 0.02493 -0.088040 1.98438
 204.9.54.119    10.128.0.45      1   64    3 0.00943 -0.035220 1.98444
 199.4.29.166    10.128.0.45      2   64    3 0.02950 -0.021567 1.98444
 216.129.110.22  10.128.0.45      2   64    3 0.01813 -0.027695 1.98444

# /usr/lib/nagios/plugins/check_ntp_peer -H 127.0.0.1 -w 5 -c 10 -v
0 candiate peers available
warning: no synchronization source found
warning: LI_ALARM bit is set
NTP CRITICAL: Server not synchronized, Offset unknown|

# /usr/lib/nagios/plugins/check_ntp_time -H 127.0.0.1 -w 5 -c 10 -v
sending request to peer 0
response from peer 0: offset 2.74181366e-06
sending request to peer 0
response from peer 0: offset -4.291534424e-06
sending request to peer 0
response from peer 0: offset -3.457069397e-06
sending request to peer 0
response from peer 0: offset -3.576278687e-06
discarding peer 0: stratum=0
overall average offset: 0
NTP CRITICAL: Offset unknown|
```

fmikker:
Also adding a cross-ref to a external issue, the information it contains might be useful: https://jira.op5.com/browse/MON-8045

ge-fa:
Is there any fix insight?

Fnux:
anyone ?

riskersen:
Could you please provide some more information? 

What OS are you using? 
Could you please provide verbose output?

BR
Oliver

Fnux:
OS : Debian 9 stretch.
```
tfloure@orwell:~ » uname -a                                                       
[...]

PR #1605, opened 2019-09-18
NTP monitoring update

eskyuu:
Hi,

I've been using the check_ntp and derivatives for many year, and have always had to add the "restrict" option to the ntp servers that I am monitoring in order for these checks to work reliably.  I came across this post today that talks about the NTP kiss of death, where a server purposely sets the stratum to 0 if a client polls too frequently.
https://serverfault.com/questions/625027/nagios-check-ntp-time-offset-unknown

The post above says that this is correct behaviour, and check_ntp should handle it.

I also had a look at the behaviour of ntpdate when querying the same NTP server, and noticed that it sends multiple queries to the NTP server and leaves a 2 second pause between queries.  ntpdate was able to query the server successfully.

I've modified the check_ntp  and check_ntp_time plugins to add a 2 second delay when querying the NTP servers, and this has fixed the problem for me.  I've made it a command-line option, so people can revert back to the original behaviour if they want.

This has been reported as an issue here (as well as on many forums if you do a web search for "nagios ntp check stratum 0"):
https://github.com/nagios-plugins/nagios-plugins/issues/329


hansbogert:
@weiss Sorry to just mention you out of the blue, but how do PRs get a prio? I can confirm the problem stated in the PR post.

@sharkyzz @petski

dermoth:
Hi, and thanks for your contribution.

I have the following issues with this patch:

1. Please squash the two commits into one

2. I see no sleep; are you implementing thew delay with the while loop? That would be pretty bad on heavy monitoring servers that can run thousands of checks per second.

3. The default behavior should not change; this too could have an impact on large monitoring hosts, and users do not expect checks behavior to change without a warning.

4. The plugin's timeout should be checked  and increased if too low (if I'm not mistaken should be something like `(delay+count+1) * 2` to take into account the fact the sending loop uses half the timeout value.)

Regards

eskyuu:
1 - I'm not sure how to squash the commits, but can go an find out how if this is the only blocker.  From memory I did the initial commit on check_ntp_time because that's what I use, and then thought about porting the patch to other plugins as an afterthought.

2-  There is an existing poll() call with a delay that avoids running a tight loop, and already handles the retries.  The patch was written to work with the existing retry mechanism.

3 - I agree that changing the default behaviour requires additional consideration, however I did cover this in my original post by commenting on other complaints I could see from monitoring-plugin users as well as noting that ntpdate has the same 2 second delay before retrying.  I also covered how the 0s delay causes the NTP servers to blacklist NTP requests from the monitoring server.

I will understand if you want to change the patch to leave it with a 0s delay by default, but if you do this I think the documentation should be updated to indicate that you need to whitelist the monitoring host on all ntp servers to stop them getting blacklisted.  For this reason I feel that changing the default behaviour is the right thing to do in order to fix the check_ntp* commands to behave nicely with NTP servers by default.  Having reviewed the timeout code (below), I don't think this change in behaviour would have any noticeable impact on any monitoring hosts.

4 - I hadn't thought of this, but I've had a quick look at the timeout code, and I think the check already had a built in safety for this.  The main loop will exit after timeout/2 seconds, and report as long as at least one packet has been processed (which will be sub 1s if the NTP server responds immediately).  This code does look a bit unusual to me for large timeout values, since I can't imagine the final processing taking more than a second.

I think the minimum timeout to complete all packets would be 1 + (delay * AVG_NUM), since the best case scenario is  that request 1 is sent, received quickly and processed, then we wait delay seconds before sending the request 2.  This process is repeated until AVG_NUM requests have been sent.  This would be extended if there are any requests that don't get a response.


I think that 2 is a non issue, I'd like to get confirmation that you want me to change 3 to the original broken implementation, and whether you're happy with the old behaviour for 4, where a small timeout will not break things, but just result in reduced accuracy.  Can you please provide your feedback, and I'll work out how to squash the commits together and implement any required changes.

hansbogert:
@dermoth @eskyuu myself or colleagues are willing to help with implementation and/or testing, I think @eskyuu and myself are waiting for @dermoth to coordinate this. How are PRs normally coordinated in this project?

saz:
Any news on this? Anything I can help with?

RincewindsHat:
@dermoth Are you still looking into this?

dermoth:
I ought to! - I actually was thinking of this one every now and then... I took a few days off first weeks of January so I have no excuse now. I will be looking at it this week.

eskyuu:
Should I create a branch and a new pull request from a branch so the commits can be squashed?

dermoth:
@eskyuu I will probably create my own branch/pr, but retain authorship info. It's up to you but it won't make a difference unless I'm happy with the PR as-is (it's been a while and I need to look at it, can't even remember what it's about or if I had any specific issue about it...)

RincewindsHat:
@dermoth Next year? :-P

waja:
> @dermoth Next year? :-P

no ETA was given... ;)

dermoth:
I clearly didn't take enough days in January :s

I do have this and a couple other things queued up I just haven't had the time to put any effort toward oss development this year, which is very sad. Lots of personal stuff to deal with on top of work, but this is the first thing i'll be looking into when I find the time - hopefully around the coming holidays...

RincewindsHat:
would like fix #1142

RincewindsHat:
The idea was implemented #2277. Sorry for the delay.

Issue #1608, opened 2019-11-14
check_disk - exclude squashfs, etc. by default


bryceharrington:
In Ubuntu, squashfs is routinely used for snap files, but since these are synthetic filesystems they display as 100% utilization and thus trigger check_disk to issue a false positive DISK CRITICAL error.  

Similar false positives have been reported to us for tracefs and tmpfs. The commonly discussed workaround is to exclude these when running the tools (e.g. using the '-X <fs>' parameter for check_all_disks). Since wrappers are typically used for running the underlying tools, it is possible to add a string of -X... parameters.

Unfortunately the number of flags needed is getting extensive, and I wonder if this could be solved more uniformly.  For filesystems that are always going to show 100% capacity, a cleaner solution is possible. monitoring-plugins' check_disk.c maintains an internal exclusion list, fs_exclude_list, which already excludes iso9660, and can be modified to add other filesystems to exclude by default.

In other words, check_disk.c would be modified thusly:

  np_add_name(&fs_exclude_list, "iso9660");
  np_add_name(&fs_exclude_list, "squashfs");
  np_add_name(&fs_exclude_list, "tmpfs");
  np_add_name(&fs_exclude_list, "tracefs");

This code is added prior to the command line parsing logic, and as such simply sets default behavior. It does not preclude further adding or removing filesystems via the -X and -N parameters. Indeed, if someone were to desire checking tmpfs, they are able to manually add it, via "-N tmpfs".

odenbach:
I personally do not care if I have to add 12 or 16 parameters to check_disk, as that is done once in the Icinga2 command definition. I already exclude

		"none",
		"tmpfs",
		"sysfs",
		"proc",
		"configfs",
		"devtmpfs",
		"devfs",
		"mtmfs",
		"tracefs",
		"cgroup",
		"fuse.gvfsd-fuse",
		"fuse.gvfs-fuse-daemon",
		"fdescfs",
		"overlay",
		"nsfs",
		"squashfs"

But that is of course site-dependant.

NHellFire:
I'm voting no on adding tmpfs to the exclude list, as excludes can't be overridden by command line arguments. tmpfs also reports correctly on my Ubuntu systems:
```
nathan@nagios:~$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
[...]

PR #1609, opened 2019-11-14
Exclude tmpfs, squashfs, and tracefs.

bryceharrington:
These synthetic filesystems always report 100% disk usage by design, so
can generate false-positive "disk full" alerts.  While it is possible to
exclude these via command line arguments, it is more convenient to
ignore them directly by the tool.

Fixes: https://github.com/monitoring-plugins/monitoring-plugins/issues/1608

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -498,6 +498,9 @@ process_arguments (int argc, char **argv)
     return ERROR;
 
   np_add_name(&fs_exclude_list, "iso9660");
+  np_add_name(&fs_exclude_list, "squashfs");
+  np_add_name(&fs_exclude_list, "tmpfs");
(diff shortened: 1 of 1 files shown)


bryceharrington:
Some of the various reports of false positives:

https://bugs.launchpad.net/ubuntu/+source/monitoring-plugins/+bug/1827159
https://bugs.launchpad.net/nrpe-charm/+bug/1710239
https://github.com/icinga/icinga2/issues/4184
https://github.com/nagios-plugins/nagios-plugins/issues/437

RincewindsHat (commented):
I would agree on tracefs, but I don't know how it looks for squashfs.
tmpfs is a valid, "normal" file system, if I understand correctly and I wouldn't want to exclude it generaly.

RincewindsHat:
closed until there is some opposition

PR #1613, opened 2019-12-25
Fix timeout_interval declarations

ChristianTacke:
There are different declarations/definitions for `timeout_interval`:

`lib/utils_base.c` has the definition:
```c
unsigned int timeout_interval = DEFAULT_SOCKET_TIMEOUT;
```
`lib/utils_base.h` has the appropriate declaration:
```c
extern unsigned int timeout_interval;
```

`plugins/popen.h` has an extra declaration:
```
extern unsigned int timeout_interval;
```
This doesn't hurt, but it's a dupe. The one in `utils_base.h` should be enough, so remove this one.

`plugins/popen.c` has a WRONG one:
```c
extern int timeout_interval;
```
Remove it!
Use `#include "utils.h"` to get the right one.
This makes the local defines for `max`/`min` unnecessary, so remove them also.

### DIFF
--- a/plugins/popen.c
+++ b/plugins/popen.c
@@ -39,9 +39,9 @@
 *****************************************************************************/
 
 #include "common.h"
+#include "utils.h"
 
 /* extern so plugin has pid to kill exec'd process on timeouts */
-extern int timeout_interval;
 extern pid_t *childpid;
 extern int *child_stderr_array;
 extern FILE *child_process;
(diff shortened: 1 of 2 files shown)


sni:
there are at least 3 variations on handling timeouts for historical reasons. thanks for your pull requests, i will have a look

ChristianTacke:
@sni

Thanks for looking!

My target was to fix the declarations, not the actual handling. So that the compile is cleaner.

Refactoring the actual handling wasn't my scope.
So, this might be a start?

PR #1617, opened 2020-01-26
check_mysql: Support named slaves

GerMalaz:
Add -N slave_name 
Supports both MySQL and MariaDB.
Makes -S fail if # slaves !=1 (and no name for a slave was provided)


GerMalaz:
Will redo.

PR #1618, opened 2020-01-30
Revert "simplify documentation"

datamuc:
This reverts commit 7693af147926d0eb7369a43070d114b6bd79dc70.

Now it is wrong:

After reading

    https://www.naemon.org/documentation/usersguide/pluginapi.html#plugin_output_spec

the old wording makes more sense.


sni:
Maybe we should add that link or a similar more visual explanation.

PR #1620, opened 2020-02-07
Fix checking of amount of mails in queue from/to the same domain.

tobiasstein:
It has been around for years but was buried under the parsing of cli arguments.


RincewindsHat:
Would you mind resolving the conflict (adding the long options again)?

RincewindsHat:
Rebased and put this in #1731

RincewindsHat:
Done in #1731

PR #1641, opened 2020-08-17
add LOAD prefix to load plugin

haraldj:
just a cosmetic fix so the load plugin display a LOAD prefix before check results


RincewindsHat:
The output change is simple enough, although we could start a bikeshed discussion if the additional text is necessary/useful :-) 

Apart from that, could you remove the second patch and put it in a different PR? Mixing up two completely different things could create some confusion.

haraldj:
Hi,

> The output change is simple enough, although we could start a bikeshed
> discussion if the additional text is necessary/useful :-)

well IMHO it standardizes the plugin output as all other plugins I use
prefix their output with their plugin suffix.

> 
> Apart from that, could you remove the second patch and put it in a different
> PR? Mixing up two completely different things could create some confusion.

As I was unable to comply with your request (seems the webinterface
doesn't support this workflow) I just closed my pull request #1641 and
created a new one (#1693) with just the first commit, will do a second
one for the other commit.

Kind regards
Harald Jenny

PR #1642, opened 2020-08-18
check_swap: Handle cached swap

mdavranche:
The "cached memory" is an "in-memory cache for files read from the disk (the pagecache)" ([ref.](https://github.com/torvalds/linux/blob/master/Documentation/filesystems/proc.rst)). It is marked as available, as if memory is needed and there is no more free memory, then the cached memory is released and then used.

This behaviour is well known, and all the monitoring plugins checking the RAM handle the cached memory: they consider the available memory as being the free momory + the cached one.

On the swap side, there is exactly the same behaviour. The cached swap is "Memory that once was swapped out, is swapped back in but still also is in the swapfile (if memory is needed it doesn't need to be swapped out AGAIN because it is already in the swapfile. This saves I/O)" ([ref.](https://github.com/torvalds/linux/blob/master/Documentation/filesystems/proc.rst)). It is marked as available, as if swap is needed and there is no more free swap, then the cached swap is released and then used.

This behaviour is not yet handled by the check_swap monitoring plugin. Like for the memory, the available swap should be considered as the free swap + the cached one. This is what this PR is for.

Note that this feature is limited to Linux, if we use /proc/meminfo only, as the free tool does not shows the CachedSwap.

**Example**

This is what my swap usage looks like:
```
# grep Swap /proc/meminfo 
SwapCached:       181904 kB
SwapTotal:       5242876 kB
SwapFree:        4556800 kB
```

The current check_swap plugin tells me:
```
# ./check_swap -c 10 -w 20
SWAP OK - 87% free (4450 MB out of 5119 MB) |swap=4450MB;0;0;0;5119
```

With my patch:
```
# ./check_swap -c 10 -w 20
SWAP OK - 91% free (4627 MB out of 5119 MB) |swap=4627MB;0;0;0;5119
```


mdavranche:
Note that this patch is now deployed on production on the 1000+ servers used to host the [OVHcloud CloudDB SaaS](https://docs.ovh.com/gb/en/clouddb/). The on-call is not awaken for false positives since then =)

RincewindsHat (approved):
Looks good too me.

RincewindsHat:
Sorry for the delay. I think this interpretation of "free" is probably the one most people would support.
Thank you very much for the contribution

Issue #1627, opened 2020-03-26
check_mailq doesn't recognize -W and -C arguments.


CeeTeih0:
Hi there! Either the documentation [1] is outdated or check_mailq is missing optional arguments.
```
./check_mailq -w 100 -c 200 -W 10 -C 20
Unknown option: W
Unknown option: C
```
I tested the version from the master branch.

[1]  https://www.monitoring-plugins.org/doc/man/check_mailq.html

RincewindsHat:
Can confirm for Debian stable

RincewindsHat:
This seems to be an artifact from https://github.com/monitoring-plugins/monitoring-plugins/issues/381 . This was fixed in aa79ab278ffc9bf3d41207d97bdfb04be35b5c74 .
This issue can be closed I guess.

PR #1646, opened 2020-10-11
Remove non-working -C and -W options from check_mailq

RincewindsHat:
[Issue](https://github.com/monitoring-plugins/monitoring-plugins/issues/1627)

I don't know enough Perl to fix this correctly, so removing the misleading information is the least I can do.


waja:
Fixed by #1650

PR #1647, opened 2020-10-19
Remove doc parts which mention non existing Files

mookie-:
There is no "ChangeLog" or "INSTALL" file anymore in the repository. Mentioning them in the README is confusing.

### DIFF
--- a/README
+++ b/README
@@ -1,14 +1,7 @@
 Monitoring Plugins
 ==================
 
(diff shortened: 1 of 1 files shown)


waja:
`INSTALL` is provided by the tarball. Even `ChangeLog` is there (even if its nearly empty in 2.3).

Issue #381, opened 2013-09-24
check_mailq still fails to run [sf#1094324]


monitoring-user:
Submitted by reuben on 2005-01-02 04:52:04

This has been in need of fixing for a long time now,
can someone please look at it before the next release?

[root@tornado plugins-scripts]# ./check_mailq
Duplicate specification "Warning=i" for option "warning"
Duplicate specification "Critical=i" for option "critical"
Usage: check_mailq -w <warn> -c <crit> [-W <warn>] [-C
<crit>] [-M <MTA>] [-t <timeout>] [-v verbose]
[root@tornado plugins-scripts]# 

Fixed by removing one of the 'duplicate' sets of
Warning and Critical parameters under

sub process_arguments(){
        GetOptions

PR #1650, opened 2020-12-08
check_mailq: restore accidentially removed options

sni:
with https://github.com/monitoring-plugins/monitoring-plugins/issues/381 -W and -C have been removed. It
would have been sufficient to only remove the long options. Restoring short options..

 - fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=954961


waja (approved):
$ /usr/lib/nagios/plugins/check_mailq -w 100 -c 200 -W 10 -C 20
OK: postfix mailq reports queue is empty|unsent=0;100;200;0

👍

PR #1659, opened 2021-02-10
Missing oldlog now aborts check_log + Typos + superfluous whitespace

RincewindsHat:
check_log can not work if the oldlog parameter is missing. This patch will abort the execution if this is the case.


RincewindsHat:
closed due to #1732

Issue #1660, opened 2021-02-15
check_pgsql: check of database name is too strict


waja:
In Debian we got the following [bug](http://bugs.debian.org/982847) reported:

Hi *,
to reproduce create a Database called freshports.git and try to
use it with `check_pgsql`:

```
flo@p5:/tmp/monitoring-plugins-2.3$ /usr/lib/nagios/plugins/check_pgsql -d freshports.devgit -l flo
check_pgsql: Database name is not valid - freshports.devgit
Usage:
check_pgsql [-H <host>] [-P <port>] [-c <critical time>] [-w <warning time>]
 [-t <timeout>] [-d <database>] [-l <logname>] [-p <password>]
[-q <query>] [-C <critical query range>] [-W <warning query range>]
```
```
flo@p5:/tmp/monitoring-plugins-2.3$ psql freshports.devgit
psql (13.1 (Debian 13.1-1+b1))
Type "help" for help.

freshports.devgit=# \d
Did not find any relations.
freshports.devgit=# 
```

The problem is that `check_pgsql` validates the Database name and has different assumptions
that postgres itself.

I fail to see a reason to validate the database name here. Postgres'es API should
do this - So i would suggest a fix like this by removing `is_pg_dbname` alltogether.

waja:
closed by dae075e

louislepage:
Hi,

if I am not mistaken, it seems like the actual fix never made it into the master and the current release.
Only the commit about `sprintf` made it.
The PR #1662 was closed without merging the fixes.

Is there any reason for this? 
The proposed solutions seems solid to to me.

Cheers,
Louis

waja:
okay ... I created #1803 for this ...

PR #1662, opened 2021-02-15
Removing is_pg_dbname alltogether,using postgres API. (Closes: #1660)

waja:
The problem is that check_pgsql validates the Database name and has different assumptions
that postgres itself.

I fail to see a reason to validate the database name here. Postgres'es API should
do this - So i would suggest a fix like this by removing is_pg_dbname alltogether.

### DIFF
--- a/plugins/check_pgsql.c
+++ b/plugins/check_pgsql.c
@@ -69,7 +69,6 @@ int process_arguments (int, char **);
 int validate_arguments (void);
 void print_usage (void);
 void print_help (void);
-int is_pg_dbname (char *);
 int is_pg_logname (char *);
 int do_query (PGconn *, char *);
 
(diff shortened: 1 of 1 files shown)


plugins/check_pgsql.c
```diff
@@ -344,10 +343,7 @@ process_arguments (int argc, char **argv)
 				pgport = optarg;
 			break;
 		case 'd':     /* database name */
-			if (!is_pg_dbname (optarg)) /* checks length and valid chars */
-				usage2 (_("Database name is not valid"), optarg);
-			else /* we know length, and know optarg is terminated, so us strcpy */
-				strcpy (dbName, optarg);
+			strcpy (dbName, optarg);
```
tobiaswiese, line 346: The ``is_pg_dbname`` check, also checked whether ``optarg`` is at most ``NAMEDATALEN`` octets long.
Using ``strcpy(2)`` without knowing that src string is not longer than the destination buffer is unsafe.

tobiaswiese (changes requested):

waja:
rebased the PR to latest master

PR #1663, opened 2021-02-15
Using snprintf which honors the buffers size and guarantees null temination. (Closes: #1601)

waja:
As strcpy may overflow the resulting buffer:

flo@p5:~$ /tmp/f/usr/lib/nagios/plugins/check_pgsql -d "$(seq 1 10000)"
*** buffer overflow detected ***: terminated
Aborted

I would propose to change the code rather like this, using snprintf
which honors the buffers size and guarantees null termination.


sni (approved):
thanks, looks fine

PR #1664, opened 2021-02-28
sslutils: use chain from client certificates

tobiaswiese:
sslutils used to load only the first certificate when it was given a
client certificate file.

Added tests for check_http to connect to a http server that expects a
client certificate (simple and with chain).

Because multiple additional certificates are now required for the check_http tests,
I created a script that regenerates the certificates.

Signed-off-by: Tobias Wiese <tobias@tobiaswiese.com>


tobiaswiese:
rebased onto upstream/master

RincewindsHat (commented):
Looks good to me, but there is a lot of magical perl :-)
I would like to have a second opinion on this.

sni (approved):
looks good to me

waja:
so lets rush this into master! 🎉

PR #1666, opened 2021-03-11
fix check_curl crash if http header contains leading spaces

sni:
check_curl crashes when a (broken) http server returns invalid http header with
leading spaces or double colons. This PR adds a fix and a test case for this.


jacobbaungard (approved):
This looks good to me

PR #1668, opened 2021-04-06
Update check_curl.c to display a specific human-readable error message where possible

bazzisoft:
Currently check_curl uses the libcurl function `curl_easy_strerror()` to convert from error codes to error messages. This function does not provide real context but rather returns a fixed string based on the error code. A much more accurate and user-friendly error code is available via the `CURLOPT_ERRORBUFFER` option. This PR tries to use the error message from that option where possible, falling back to the former in case it is not available.

For more details see:
https://curl.se/libcurl/c/curl_easy_strerror.html
https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html

### Sample output - BEFORE:
```
$ check_curl -H expired.badssl.com -S --ca-cert=/etc/ssl/certs/ca-certificates.crt
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 60 - Peer certificate cannot be authenticated with given CA certificates
```

### Sample output - AFTER:
```
$ check_curl -H expired.badssl.com -S --ca-cert=/etc/ssl/certs/ca-certificates.crt
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 60 - SSL certificate problem: certificate has expired
```

### DIFF
--- a/plugins/t/check_curl.t
+++ b/plugins/t/check_curl.t
@@ -46,7 +46,7 @@ $res = NPTest->testCmd(
 	);
 cmp_ok( $res->return_code, '==', 2, "Webserver $host_nonresponsive not responding" );
 # was CRITICAL only, but both check_curl and check_http print HTTP CRITICAL (puzzle?!)
-cmp_ok( $res->output, 'eq', "HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 28 - Timeout was reached", "Output OK");
+like( $res->output, "/HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 28 - Connection timed out after/", "Output OK");
 
 $res = NPTest->testCmd(
 	"./$plugin $hostname_invalid -wt 1 -ct 2"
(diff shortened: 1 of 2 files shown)


sni (approved):
thanks, looks good

sni:
Could you have a look at the test case? The error messages have to be adjusted accordingly.

bazzisoft:
@sni Yeah, trying to get them running locally but getting 50% failures. Will try to figure it out this AM.

sni:
Thoses tests can be a bit tricky sometimes, i guess it would be enough to simply adjust the test strings based on the last travis result.

plugins/t/check_curl.t
```diff
@@ -46,7 +46,7 @@ $res = NPTest->testCmd(
 	);
 cmp_ok( $res->return_code, '==', 2, "Webserver $host_nonresponsive not responding" );
 # was CRITICAL only, but both check_curl and check_http print HTTP CRITICAL (puzzle?!)
-cmp_ok( $res->output, 'eq', "HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 28 - Timeout was reached", "Output OK");
+cmp_ok( $res->output, 'eq', "HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 28 - Connection timed out after 3001 milliseconds", "Output OK");
```
sni, line 49: I guess it would be better to do a `like()` here as well, testing for exactly 3001 milliseconds is probably a bad idea.

bazzisoft: @sni Yep agreed, fixed that now & tests are passing locally.

sni:
perfect, thanks.

PR #1669, opened 2021-04-07
Add an option to check_curl to verify the peer certificate & host using the system CA's

bazzisoft:
One of the major considerations for moving to check_curl is the ability to verify peer certificates automatically, using the system CA settings (i.e. the default curl behaviour). While maintaining backwards compatibility is necessary for a drop-in replacement of check_http, it would be nice to have an option to reactivate certificate verification without having to specify a CA chain.

The new option is `-D` or `--verify-cert`
### Sample output :
```
$ check_curl -H expired.badssl.com -S
HTTP OK: HTTP/1.1 200 OK - 761 bytes in 0.767 second response time |time=0.766678s;;;0.000000;10.000000 size=761B;;;0

$ check_curl -H expired.badssl.com --ca-cert=/etc/ssl/certs/ca-certificates.crt
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 60 - SSL certificate problem: certificate has expired

$ check_curl -H expired.badssl.com -D
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 60 - SSL certificate problem: certificate has expired

$ check_curl -H expired.badssl.com --verify-cert
HTTP CRITICAL - Invalid HTTP response received from host on port 443: cURL returned 60 - SSL certificate problem: certificate has expired
```

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -195,6 +195,7 @@ int ssl_version = CURL_SSLVERSION_DEFAULT;
 char *client_cert = NULL;
 char *client_privkey = NULL;
 char *ca_cert = NULL;
+int verify_peer_and_host = FALSE;
 int is_openssl_callback = FALSE;
 #if defined(HAVE_SSL) && defined(USE_OPENSSL)
 X509 *cert = NULL;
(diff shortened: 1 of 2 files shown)


bazzisoft:
The test that was added in this PR fails because the CI environment uses a self-signed certificate for `localhost`. I don't really see a way around this apart from removing that test?

sni:
I'd still keep the test and adjust it to match the current result.

bazzisoft:
@sni Can do, only problem I see is it will work when running tests against self-signed but running locally against a valid certificate will fail (currently all other tests pass using a valid cert). I'll try doing a regex OR to capture both cases.

bazzisoft:
@sni That seems to work, let me know your thoughts. Cheers!

sni (approved):
looks fine to me

sni:
Looks good. Would it make sense to include certificate performance values if this new option is specified?

bazzisoft:
@sni I'm not sure what those are? Do you mean expiry time or similar like with the `-C` flag? If so not so much (for my use case at least) as we're really just trying to check if HTTPS connectivity to a site is working as a user would see it

PR #1678, opened 2021-04-23
Fix trailing whitespaces in C plugins

RincewindsHat:
I am kinda sorry about this, but I am highlighting trailing whitespaces in my editor and this annoys me.
So I executed `find . -type f -exec sed -i 's/\s\+$//' {} \;` in the plugins directory.
It would be nice if that could be made a permanent state.

### DIFF
--- a/plugins/check_apt.c
+++ b/plugins/check_apt.c
@@ -1,32 +1,32 @@
 /*****************************************************************************
-* 
+*
 * Monitoring check_apt plugin
-* 
+*
 * License: GPL
 * Copyright (c) 2006-2008 Monitoring Plugins Development Team
-* 
+*
 * Original author: Sean Finney
-* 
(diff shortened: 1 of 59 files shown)


sni (approved):
+1

waja:
Beside this is a great move, it will make backporting of fixes into maint-2.3 much more complex.

sni:
we should do this at some point and this point is as bad as the others. Cherry-Picking should be straight forward with `git cherry-pick -x -X ignore-all-space <commit>` since its really just white space changes.

phibos:
Maybe something like [pre-commit](https://pre-commit.com/) could help reducing this type of errors before committing and pushing. I can provide a PR if you are interested.

sni:
superseeded by #1680

PR #1682, opened 2021-04-28
Add CodeQL checks

phibos:
Add initial github actions to perform CodeQL securtiy checks

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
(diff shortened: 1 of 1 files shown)


phibos:
Please feel free to contact me if you need more information on this.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '15 18 * * 0'
```
RincewindsHat, line 21: I don't think, that we need to run this regularly, since this is a bunch of rather old C and perl stuff.
On push should suffice. If something is broken, this won't change for suddenly I guess

RincewindsHat (changes requested):
Hey, thanks for the submission. Generaly I think this won't hurt and might be worth trying.
Just a small annotation: I would like this to happen on PRs, if that is possible.
Running it regularly would be wasting ressources, but I might be wrong there.

.github/workflows/codeql-analysis.yml
phibos: I would recommend scanning the code on a regular basis.

RincewindsHat: well, you are the pro :-) (I guess)

RincewindsHat:
@sni @weiss just try it?

sni:
Why not, it seems well integrated in github. Tested it in my fork and it finds 12 alerts:
https://github.com/sni/monitoring-plugins/security/code-scanning?query=is%3Aopen+branch%3Acodeql

RincewindsHat (approved):
Lets try

PR #1683, opened 2021-04-29
Make tests run with custom settings

akselsjogren:
This PR has the patches I needed to make to run the test suite with different settings (`NP_*` vars), on a different host than what Travis CI uses.


jacobbaungard:
Included with https://github.com/monitoring-plugins/monitoring-plugins/pull/1686 , thanks!

PR #1689, opened 2021-05-27
fixes check_curl: "CRITICAL - Cannot retrieve certificate subject."

andreasbaumann:
- added verbose output in verify_callback
- pin refcounting for certs (avoid subject extraction error when checking
  certs in is_openssl_callback mode)


sni (approved):

sni:
thanks

PR #1690, opened 2021-06-24
check_http: changed 'STATE_CRITICAL' to 'STATE_WARNING' for infinite loop

xFuture603:
We use the check_http in our Icinga instance and once had the case that we got the message (HTTP WARNING - redirection creates an infinite loop) issued with the warning status. Since this is an infinite loop message I think the state should be rather 'critical', right?

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -1453,8 +1453,8 @@ redir (char *pos, char *status_line)
       !strncmp(server_address, addr, MAX_IPV4_HOSTLENGTH) &&
       (host_name && !strncmp(host_name, addr, MAX_IPV4_HOSTLENGTH)) &&
       !strcmp(server_url, url))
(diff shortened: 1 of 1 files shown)


RincewindsHat (approved):
Sounds reasonable.

andreasbaumann:
backported for -ffollow in check_curl too.
-fcurl already returns a CRITICAL:

```
HTTP CRITICAL - Invalid HTTP response received from host on port 80: cURL returned 47 - Maximum (16) redirects followed
```

sni (approved):

avalor1:
So uhm... when does this gets merged? :)

sni:
I'd say yes, waiting for a second opinion.

weiss (approved):

sni:
thanks Holger, merged it.

PR #1692, opened 2021-07-02
Modernize check log

RincewindsHat:
Some cleanup stuff and shellcheck corrections.
Also:
 * adds `-a|--all` option to print all matching lines,
 * add  `-p|--perl-regex` option to use grep with perl style regex
 * add  `-e|--extended-regex` option to use grep with extended regex style

### DIFF
--- a/plugins-scripts/check_log.sh
+++ b/plugins-scripts/check_log.sh
@@ -59,11 +59,11 @@
 
 PATH="@TRUSTED_PATH@"
 export PATH
-PROGNAME=`basename $0`
-PROGPATH=`echo $0 | sed -e 's,[\\/][^\\/][^\\/]*$,,'`
+PROGNAME=$(basename "$0")
+PROGPATH=$(echo "$0" | sed -e 's,[\\/][^\\/][^\\/]*$,,')
 REVISION="@NP_VERSION@"
 
-. $PROGPATH/utils.sh
+. "$PROGPATH"/utils.sh
 
 print_usage() {
     echo "Usage: $PROGNAME -F logfile -O oldlog -q query"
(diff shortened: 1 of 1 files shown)


plugins-scripts/check_log.sh
```diff
@@ -60,20 +59,25 @@
...
+	echo ""
+	echo "Other parameters:"
+	echo "	-a|--all : Print all matching lines"
+	echo "	-p|--perl-regex : Use perl style regular expressions in the query"
+	echo "	-e|--extended-regex : Use extended style regular expressions in the query (not necessary for GNU grep)"
 }
 
 print_help() {
-    print_revision $PROGNAME $REVISION
+    print_revision "$PROGNAME" $REVISION
```
sni, line 80: i guess you want to put quotes over all variables?

plugins-scripts/check_log.sh
```diff
@@ -100,111 +104,159 @@ while test -n "$1"; do
...
             ;;
         -h)
             print_help
-            exit $STATE_OK
+            exit "$STATE_OK"
             ;;
         --version)
-            print_revision $PROGNAME $REVISION
-            exit $STATE_OK
+            print_revision "$PROGNAME" $REVISION
```
sni, line 114: same here

plugins-scripts/check_log.sh
```diff
@@ -100,111 +104,159 @@ while test -n "$1"; do
...
+            exit "$STATE_OK"
             ;;
         -V)
-            print_revision $PROGNAME $REVISION
-            exit $STATE_OK
+            print_revision "$PROGNAME" $REVISION
```
sni, line 118: and here

sni (changes requested):
looks good, just a minor change request :-)

plugins-scripts/check_log.sh
RincewindsHat: I do and localy they are there oO. Also they are in my branch...

RincewindsHat:
Updated it

RincewindsHat:
@sni

sni (approved):

PR #1693, opened 2021-07-05
add LOAD prefix to load plugin

haraldj:
just a cosmetic fix so the load plugin display a LOAD prefix before check results

### DIFF
--- a/plugins/check_load.c
+++ b/plugins/check_load.c
@@ -209,7 +209,7 @@ main (int argc, char **argv)
 		else if(la[i] > wload[i]) result = STATE_WARNING;
 	}
 
-	printf("%s - %s|", state_text(result), status_line);
+	printf("%s %s - %s|", "LOAD", state_text(result), status_line);
 	for(i = 0; i < 3; i++)
 		printf("load%d=%.3f;%.3f;%.3f;0; ", nums[i], la[i], wload[i], cload[i]);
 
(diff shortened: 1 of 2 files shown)


sni:
agreed, that would match better with our other plugins. However, may i suggest a simple:
`printf("LOAD %s - %s|", state_text(result), status_line);`

haraldj:
Dear Sven Nierlein,

> agreed, that would match better with our other plugins. However, may i suggest
> a simple:
> printf("LOAD %s - %s|", state_text(result), status_line);

ok, changed my patch this way, thanks for the heads up.

Kind regards
Harald Jenny

RincewindsHat:
Hi @haraldj , 
Seems like you checked in the smtp change again :-)

sni:
Also please adjust the tests in https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/t/check_load.t

haraldj:
> Hi @haraldj ,

Hi,

> Seems like you checked in the smtp change again :-)

yes seems I'm not fit enough currently to do github ;-).

Thanks
Harald

haraldj:
Hi

> Also please adjust the tests in https://github.com/monitoring-plugins/
> monitoring-plugins/blob/master/plugins/t/check_load.t

please check #1694 it should be correct now and complete.

Kind regards
Harald

PR #1695, opened 2021-07-09
Trivial printf fix and a little bit of code style

RincewindsHat:
…g in check_disk.
Should update/replace #1326


waja:
Hmm ... doesn't look very destructive. :)

RincewindsHat:
I think, I managed to sneak most of it into other PRs :-)

waja (approved):
So I'd say: looks good

Issue #1696, opened 2021-07-09
Don't show rta if check_ping is not able to reach the host


RincewindsHat:
Pinging an not existing host like this:
```
./plugins/check_ping -H 192.168.0.123 -w 60,1% -c 100,5%
PING CRITICAL - Packet loss = 100%|rta=100.000000ms;60.000000;100.000000;0.000000 pl=100%;1;5;0
```
shows round trip average time altough this is wrong in this case and will produce wrong numbers.

RincewindsHat:
#1697

RincewindsHat:
fixed with #1697

PR #1697, opened 2021-07-09
check_ping: Do not show RTA if no connection was possible

RincewindsHat:
Displaying the RTA in perfdata will confuse the system using this data  and falsify averages.

### DIFF
--- a/plugins/check_ping.c
+++ b/plugins/check_ping.c
@@ -171,7 +171,7 @@ main (int argc, char **argv)
 									  crta>0?TRUE:FALSE, crta,
 									  TRUE, 0, FALSE, 0));
 		} else {
-			printf("| rta=U;;;;;");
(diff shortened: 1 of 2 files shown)


RincewindsHat:
#1696

sni:
i'd say it's not good practice to completely hide rta performance values, i'd prefer to keep it with a `U` to indicate unknown value as in https://www.monitoring-plugins.org/doc/guidelines.html#AEN201 under 8.
That way the number of performance values won't change.

RincewindsHat:
Sounds reasonable

sni:
sorry to be picky, but thinking about it, i'd say it would be better to even keep the thresholds. I mean, they still apply, even if the value cannot be determined.

RincewindsHat:
hm, true enough, thresholds are there and valid.

sni (approved):

PR #1700, opened 2021-07-28
Feature check curl

andreasbaumann:
- adds --automatic_decompression for automatically decompressing the body
- bugfixes:
  - certificate check with old OpenSSL versions (up_ref pining)
  - DNS cache with host:port:host is no longer allowed in newest curl version, must be host:port:ip (leads to an error
    in HTTPS virtual host cases)
  - the usual scoping and buffer overflow bugfixes
- backports from check_http:
  - infinitite redirection loops should be CRITICAL


sni (approved):
looks good

PR #1623, opened 2020-02-23
check_dns: Add --expect-nxdomain

Jonny007-MKD:
I want to check that certain domains are blocked and return a NXDOMAIN (that's how my adblock script configures my unbound instance). I added this option to check_dns.

PR #1704, opened 2021-09-01
Use netcat-openbsd for debian explicitely

RincewindsHat:
I think, this might be the reason for the test problem in #1623


sni:
thanks

Issue #408, opened 2013-09-24
long long support for check_swap on release 1.3 [sf#1153148]


monitoring-user:
Submitted by carenas on 2005-02-28 01:05:15

using an unsigned long int and therefore overflowing
for swap bigger than 2^32 bytes.

still is_intnonneg util function could wrongly assume
that the parameter  is not in the right range but my
guess is that it is better to fix the parametrized
parameter to also allow KB, MB and such instead of
forcing the user to fit it as an integer so i left it
untouched.

PR #1707, opened 2021-09-25
Fix perfdata for check swap

RincewindsHat:
The original problem was #1705 
    where the performance data output of check_swap did not conform to
    the parser logic of a monitoring system (which decided to go for
    "correct" SI or IEC units.
    The PR was accompanied by a change to byte values in the performance
    data which broke the _perfdata_ helper function which could not handle
    values of this size.
    The fix for this, was to use _fperfdata_ which could, but would
    use float values.

    I didn't like that (since all values here are discreet) and this
    is my proposal for a fix for the problem.

    It introduces some helper functions which do now explicitely work
    with (u)int64_t, including a special version of the _perfdata_ helper.

    In the process of introducing this to check_swap, I stumbled over
    several sections of the check_swap code which I found problematic.
    Therefore I tried to simplify the code and make it more readable
    and less redundant.

    I am kinda sorry about this, but sincerely hope my changes can
    be helpful.

### DIFF
--- a/plugins/check_swap.c
+++ b/plugins/check_swap.c
@@ -466,7 +466,7 @@ process_arguments (int argc, char **argv)
 				length = strlen(optarg);
 
 				if (optarg[length - 1] == '%') {
-					// It's percentage!
+					/* It's percentage */
 					warn.is_percentage = true;
 					optarg[length - 1] = '\0';
 					if (is_uint64(optarg, &warn.value)) {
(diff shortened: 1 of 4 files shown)


RincewindsHat:
#408  just found that

RincewindsHat:
@weiss @sni  Anyone of you got time to take a look at this?
And yes, I know I broke about two thirds of the rules for C code on the website, but I would also like to talk about that, especially about the boolean one.

weiss:
> I didn't like that (since all values here are discreet) and this is my proposal for a fix for the problem.

I fully agree with the problem (of course), and with your solution.

You added a few `#include` direcitives for standard library headers, those are already included via `common.h` (or Gnulib).

sni (approved):
looks good to me

PR #1712, opened 2021-10-25
Fix check proc ps detection

sni:
check_procs: improve ps args autodetection
    
one of the first ps commands in the configure.ac is `axwo 'stat comm vsz rss user uid pid ppid args'` which
works on most modern linux systems (checked debian 10/11 and centos 7/8). But this test misses the etime
argument. Therefore `check_procs --metric=ELAPSED` does not work.
    
To fix this, we simply do the same test including etime before that one.
    

RincewindsHat (approved):
Looks good to me, worked too, but I can't say that I understand all the changes (especially in the configuration section).
Apart from that, good to go.

sni:
Apart from the whitespace changes and new tests including a new test file, i simply copy/pasted another if clause in the configure with the only addition beeing the etime field. Thanks for the review.

PR #1715, opened 2021-10-28
check_smtp: add -L flag to support LMTP (LHLO instead of HELO/EHLO).

ghen2:


RincewindsHat (approved):
Looks good to me, although I just found out, that the server (my server at least) answers with
```
220 ...
502 ...
```
and may this should be seen as an error. But I guess I should open an issue for that then

RincewindsHat:
Thank you for the contribution :-)

PR #1716, opened 2021-10-29
check_mysql: Add a -N slave_name option

GerMalaz:
Checks for MariaDB in server_info on deciding to use `show slave 'slave_name' status` (if matches) or `show slave status for channel '%s'` (if it doesn't match).

INCOMPATIBILITY:
      When more than one slave is present, the first one returned was checked.
      Now it fails in that case, if the -N/--slave-name option is not used.


RincewindsHat:
Sorry for not responding.
What does this do/help/improve?

GerMalaz:
If you have more than one replica, you can name all of them, and check them separately.

GerMalaz:
Iv recently updated the patch for this.
there have been changes in this area that affect it.

RincewindsHat:
Yes, sorry about that. The logic changed a bit due to refactoring and the change in naming conventions (`slave` -> `replica`).
I might give it a go to reimplement the idea or, if you are willing, you could give it a try yourself.

GerMalaz:
Not a problem.
As I said, I already did.
Used the (now) available checks for mysql|mariadb to assign the different syntax if a named replica is requested.

GerMalaz:
Ooops, didn't mean to close this, so I reopened it.
Updated for current code.
Note there is also a bugfix (replicas were always (reportedly) unavailable)

plugins/check_mysql.c
```diff
@@ -50,6 +50,10 @@ const char *email = "devel@monitoring-plugins.org";
 
 static int verbose = 0;
 
+static char *replica_query = NULL;
```
RincewindsHat: Is it necessary for these variables to be global? If possible they should only be available where needed.

plugins/check_mysql.c
```diff
@@ -50,6 +50,10 @@ const char *email = "devel@monitoring-plugins.org";
...
+static char *replica_status_query = NULL;
+static int num_slaves = 0;
```
RincewindsHat: That should be `num_replicas`?

RincewindsHat (changes requested):
Also running `clang-format` would be nice to have the same style everywhere

plugins/check_mysql.c
GerMalaz: Resolved in https://github.com/monitoring-plugins/monitoring-plugins/pull/1716/commits/0b819588af4d7188079264e2e329c2ab5b839271

plugins/check_mysql.c
GerMalaz: Addressed in https://github.com/monitoring-plugins/monitoring-plugins/pull/1716/commits/462a20a341dba51e15429aef12d5e4f92eec7316

GerMalaz:
> Also running `clang-format` would be nice to have the same style everywhere

Tried this (without parameters), and affected code beyond the scope of this PR.

RincewindsHat (approved):
Looks good :-)

Thank you for the contribution and the patience.

c-kr:
@RincewindsHat 
cc @C0rn3j

This fails to build on Arch Linux with the default build flag `-Werror=format-security`

```
check_mysql.c: In function ‘main’:
check_mysql.c:338:46: error: format not a string literal and no format arguments [-Werror=format-security]
  338 |                                         :  _("too many replicas defined, please provide a replica name as an argument")
      |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
common.h:195:27: note: in definition of macro ‘_’
  195 | #define _(String) gettext(String)
      |                           ^~~~~~

```

Minimal fix (by LLM, tested locally):

```diff
-                       mopl_utils_xasprintf(&sc_replica.output,
+                       mopl_utils_xasprintf(&sc_replica.output, "%s",

GerMalaz:
I see, sorry.
Somehow this didn't show up on my environment (debian forky/sid), should have.

RincewindsHat:
fixed it

GerMalaz:
Thanks!

PR #1718, opened 2021-10-31
check_mysql: Detect running mysqldump

GerMalaz:
When checking a slave, if the IO Thread or the SQL Thread are stopped, check for running mysqldump threads, return STATE_OK if there is any.
Requires PROCESS privilege to work (else the mysqldump thread(s) would not be detected).

Enlarged SLAVERESULTSIZE to fit "Mysqldump: in progress" at the end of the string.
Got a NULL pointer in row[seconds_behind_field] instead of the "NULL" string when a mysqldump is running [mysql 5.7.34 + libmariadb3 10.3.31], so added a check for that.


plugins/check_mysql.c
```diff
@@ -275,11 +277,29 @@ main (int argc, char **argv)
...
+							}
+						}
+						/* free the result */
+						mysql_free_result (res_mysqldump);
+					}
+				}
+				if (mysqldump_threads == 0) {
+					die (STATE_CRITICAL, "%s\n", slaveresult);
+				} else {
+					strlcat (slaveresult, " Mysqldump: in progress", SLAVERESULTSIZE);
```
RincewindsHat, line 301: Is this BSD specific? If so, could you use `strncat`?

RincewindsHat (changes requested):
ok, seems to work, but I don't know enough about mysql/mariadb to test. Could you explain to me how trigger this case or provide a scenario in which this would happen?
Unit tests would be perfect, but I don't want to ask for that much more work.

plugins/check_mysql.c
GerMalaz: Not BSD specific, will change it.

GerMalaz:
> ok, seems to work, but I don't know enough about mysql/mariadb to test. Could you explain to me how trigger this case or provide a scenario in which this would happen? Unit tests would be perfect, but I don't want to ask for that much more work.

Any time a  replica is used to backup a DB being replicated, either scheduled or manually, using mysqldump with the --dump-slave option
Without this PR, a check with replica status while the backup is being performed triggers an error (as the replica is stopped). 
Which this PR, this is to be expected (as the replica needs to be frozen while the backup is being performed) and should not trigger an error.

plugins/check_mysql.c
GerMalaz: Requested changes were addressed in https://github.com/monitoring-plugins/monitoring-plugins/pull/1718/commits/d8bc51f465e984001ccaca616e106da6b92fb015 .

GerMalaz (commented):
Changes requested were addressed, strncat now used

plugins/check_mysql.c
```diff
@@ -275,11 +277,34 @@
...
+					}
+				}
+				if (mysqldump_threads == 0) {
+					die (STATE_CRITICAL, "%s\n", slaveresult);
+				} else {
+					if (strlen(slaveresult) + strlen(" Mysqldump: in progress") <= SLAVERESULTSIZE) {
+						strncat (slaveresult, " Mysqldump: in progress", SLAVERESULTSIZE);
+					} else {
+						die (STATE_CRITICAL, "%s %d\n", "Not enough space. Need SLAVERESULTSIZE >= ",
+								strlen(slaveresult) + strlen(" Mysqldump: in progress"));
```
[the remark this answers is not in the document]
waja: @GerMalaz do you want to have a look into this?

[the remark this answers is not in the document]
GerMalaz: https://github.com/monitoring-plugins/monitoring-plugins/commit/1ae408be800fd933522a2cb79e0594f3d06d7ad7 should fix this issue.

GerMalaz:
Please see https://github.com/monitoring-plugins/monitoring-plugins/pull/1718#issuecomment-1019279641 for the use case of this.
Summarizing:
When checking for replica status, a stopped replica is (usually) bad, and an error status is returned.
However, if a backup is in progress, a stopped replica is the normal status. So, in this situation (mysqldump threads are detected), an OK status is returned even if the replica is stopped.

RincewindsHat:
hm, ok, so I tested it, but I still get
```
Slave IO: Yes Slave SQL: No Seconds Behind Master: (null)
```
sometimes

RincewindsHat:
ok, I think I am going somewhere. I used `mysqldump test | pv -p -r -L 1K > /dev/null` to test it, turns out, `pv` is heavily buffering this, which means the query runs through and then the dump waits until everything is written ( no `'SELECT→/*!40001 SQL_NO_CACHE */%'` in info). 
This made the test query fail then.

GerMalaz:
> hm, ok, so I tested it, but I still get
> 
> ```
> Slave IO: Yes Slave SQL: No Seconds Behind Master: (null)
> ```
> 
> sometimes

This is expected. The slave IS stopped, which is usually an abnormal condition, and should be alerted.

But under the conditions detected here (a backup being taken), a stopped slave is normal, and as such it should not generate an alert.

GerMalaz:
> ok, I think I am going somewhere. I used `mysqldump test | pv -p -r -L 1K > /dev/null` to test it, turns out, `pv` is heavily buffering this, which means the query runs through and then the dump waits until everything is written ( no `'SELECT→/*!40001 SQL_NO_CACHE */%'` in info). This made the test query fail then.

And this case, (slave stopped, no mysqldump thread) it should be treated as an abnormal condition (as usual when the slave is stopped)

RincewindsHat:
So, taking a look at this again (sorry, finding time ... ).
Getting a `(Got an error reading communication packets)` on the mariadb side with with `-S`.
dunno why

RincewindsHat:
ah, I think a `mysql_close` is missing. Unrelated to main problem though.

RincewindsHat:
So sorry for all the delay. I think it works. Ok for me.

waja:
Looks like CodeQL test found something...

RincewindsHat (approved):
Thank you very much for your work and for your patience.

PR #1721, opened 2021-11-15
check_icmp: Fix pkt perfdata in check_host mode

akselsjogren:
Add missing "warn" threshold field in "pkt" perfdata output.
Perfdata should be interpreted as;

`  'label'=value[UOM];[warn];[crit];[min];[max]`

With one field missing, the hardcoded min value '0' ended up in the
"crit" field, making applications interpreting the perfdata thining that
critical threshold is always exceeded.

Signed-off-by: Aksel Sjögren <asjogren@itrsgroup.com>


sni (approved):
good catch, thanks

PR #1722, opened 2021-11-17
check_snmp: fix performance thresholds when using multiple oids

sni:
when using check_snmp with multiple oids it simply printed the unparsed content
from -w/-c into the thresholds for each oid. So each oid contained the hole -w
from all oids.

./check_snmp ... -o iso.3.6.1.2.1.25.1.3.0,iso.3.6.1.2.1.25.1.5.0 -w '1,2' -c '3,4'

before:
  SNMP ... | HOST-RESOURCES-MIB::hrSystemInitialLoadDevice.0=393216;1,2;3,4 HOST-RESOURCES-MIB::hrSystemNumUsers.0=24;1,2;3,4

after:
  SNMP ... | HOST-RESOURCES-MIB::hrSystemInitialLoadDevice.0=393216;1;3 HOST-RESOURCES-MIB::hrSystemNumUsers.0=24;2;4

This also applies to fixed thresholds since check_snmp translates negative infinities from: `~:-1` to `@-1:~`

### DIFF
--- a/lib/utils_base.c
+++ b/lib/utils_base.c
@@ -128,6 +128,7 @@ range
 	temp_range->end = 0;
 	temp_range->end_infinity = TRUE;
 	temp_range->alert_on = OUTSIDE;
+	temp_range->text = strdup(str);
 
 	if (str[0] == '@') {
 		temp_range->alert_on = INSIDE;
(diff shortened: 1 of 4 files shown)


RincewindsHat (approved):
Looks good to me

PR #1724, opened 2021-11-19
Allows check_ldap to read password from environment variable

mullumaus:
It's not secure to provide LDAP password through command line option
because other users on the same host can see the password in
'ps' command output.

This change allows check_ldap to get password from environment variable.

### DIFF
--- a/plugins/check_ldap.c
+++ b/plugins/check_ldap.c
@@ -432,6 +432,9 @@ validate_arguments ()
 		set_thresholds(&entries_thresholds,
 			warn_entries, crit_entries);
 	}
+	if (ld_passwd==NULL)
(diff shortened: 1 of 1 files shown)


RincewindsHat:
Sonds good, could you add a note to the help/usage to announce that possibility?

mullumaus:
> Sonds good, could you add a note to the help/usage to announce that possibility?

Yes, I have updated the usage to  announce that possibility, thanks.

sni (approved):

sni:
That's great. We should do this for the other plugins (which require passwords) as well.

RincewindsHat (approved):
Looks good, thank you very much.

PR #1731, opened 2021-11-29
Check mailq domain specific warnings

RincewindsHat:
#1620


sni:
I'd approve it, but there seems to be conflicts in the patch which need to be resolved anyway.

waja:
This was updated ... tests seems to be failed cause no running http?

waja (approved):
Don't see breaking changes, looks good to me. ;)

PR #1732, opened 2021-12-02
Missing oldlog now aborts check_log

RincewindsHat:


RincewindsHat:
Updates #1659

sni:
I don't think this is a good idea as it potentially would lead to lots of unknowns after a update and i am a fan of the "least surprise" approach.

RincewindsHat:
To my understanding the script doesn't even work without `oldlog`, but I might be wrong

sni (approved):

sni:
sorry, my bad. I forgot how check_log.sh worked :-) Looks good then.

PR #1733, opened 2021-12-03
check_icmp: buffer offerflow

RincewindsHat:


plugins-root/check_icmp.c
```diff
@@ -50,19 +50,24 @@ const char *email = "devel@monitoring-plugins.org";
 #if HAVE_SYS_SOCKIO_H
 #include <sys/sockio.h>
 #endif
+
+/*
```
sni, line 54: That's bad practice. Either remove it or keep it, but commenting out code should be avoided. That's what we have git for.

plugins-root/t/check_icmp.t
```diff
@@ -35,51 +35,72 @@ my $hostname_invalid   = getTestParameter( "NP_HOSTNAME_INVALID",
 
 my $res;
 
+# 1
```
sni, line 38: what's the idea of these comments?

plugins-root/check_icmp.c
```diff
@@ -1514,18 +1534,19 @@ get_threshold(char *str, threshold *th)
...
-icmp_checksum(unsigned short *p, int n)
+icmp_checksum(uint16_t *p, size_t n)
 {
 	unsigned short cksum;
 	long sum = 0;
 
-	while(n > 2) {
-		sum += *p++;
-		n -= sizeof(unsigned short);
+	/* sizeof(uint16_t) == 2 */
```
sni, line 1542: what's the reason of using hard coded 2 instead of calling sizeof?

RincewindsHat: No specific reason, it's the same anyway and at list it is consistent with line 1545 now. Would it be better understandable, if I used `sizeof`?

plugins-root/t/check_icmp.t
RincewindsHat: To bee honest, just to help me with counting them

sni: I would just run the test once after adding test cases. It then tells you the new number of tests and it can be changed.

plugins-root/check_icmp.c
sni: I'd rather change the other occurance as well. But that's only my opinion.

plugins-root/t/check_icmp.t
RincewindsHat: Removed the comments

RincewindsHat:
should be fixed, review?

plugins-root/check_icmp.c
```diff
@@ -1514,18 +1521,19 @@ get_threshold(char *str, threshold *th)
...
```
sni, line 1529: can this comment be removed?

plugins-root/check_icmp.c
```diff
@@ -1514,18 +1521,19 @@ get_threshold(char *str, threshold *th)
...
+	while(n >= 2) {
+		sum += *(p++);
+		n -= 2;
 	}
 
 	/* mop up the occasional odd byte */
-	if(n == 1) sum += (unsigned char)*p;
+	if(n == 1) sum += *((uint8_t *)p -1);
```
sni, line 1536: should this be uint16_t?

plugins-root/check_icmp.c
```diff
@@ -913,9 +902,27 @@ wait_for_reply(int sock, u_int t)
...
+					printf("%0.3f ms rtt from %s, outgoing ttl: %u, incoming ttl: %u, max: %0.3f, min: %0.3f\n",
+						(float)tdiff / 1000,
+						address,
+						ttl,
+						ip->ip.ip_ttl,
+						(float)host->rtmax / 1000,
+						(float)host->rtmin / 1000);
+					break;
+					  };
+				case AF_INET6: {
```
sni, line 917: what's the reason for this change?

sni (commented):
i've got some questions :-)

plugins-root/check_icmp.c
RincewindsHat: I don't think so, this should catch a single byte, but I might be thinking wrong.

plugins-root/check_icmp.c
RincewindsHat: I think, the problem is,, that we don't get an incoming ttl if IPv6 is used, due to differences in the data structure (or the protocol?). But I already forgot was the specifics were there, I can research them again, if you would like to know the details.

sni (approved):

Issue #1607, opened 2019-11-04
check_radius: Multiple NAS-IP-Address attributes


dirnbeja:
Trying to check a RADIUS server that doesn't handle if multiple NAS-IP-Address values set, I stumbled upon that when using check_radius, it sets multiple of those.

The machine the checks are running on has the following IP config
* bond0 - $ip_a
* vlan12 - $ip_b
* vlan17 - $ip_c

The connection actually goes out on the vlan12 Interface with $ip_b as a source address.

Doing a connection "out of the box", I see that the RADIUS request has both bond0's and vlan12's IPs listed as NAS-IP-Adress (tcpdump on the RADIUS server):
```
18:15:19.734166 IP (tos 0x0, ttl 64, id 356, offset 0, flags [DF], proto UDP (17), length 96)
    $ip_a.57109 > $server.radius: [udp sum ok] RADIUS, length: 68
	Access-Request (1), id: 0x92, Authenticator: c0d7f68619a47a78ccf48d4a0b3b1ed8
	  Service-Type Attribute (6), length: 6, Value: Authenticate Only
	    0x0000:  0000 0008
	  User-Name Attribute (1), length: 12, Value: XXX
	    0x0000:  6769 6274 736e 6963 6874
	  User-Password Attribute (2), length: 18, Value:
	    0x0000:  b828 6f5e 600c e198 86e8 091e d799 ba1f
	  NAS-IP-Address Attribute (4), length: 6, Value: $ip_a
	    0x0000:  3e80 013d
	  NAS-IP-Address Attribute (4), length: 6, Value: $ip_b
	    0x0000:  0a02 013d
```
The actual outgoing interfaces' adress seems to stick - but additonally - if I give check_radius `-N $ip_b` I get the same attribute twice:
```
18:17:34.778655 IP (tos 0x0, ttl 64, id 33465, offset 0, flags [DF], proto UDP (17), length 96)
    $ip_b.37092 > $server.radius: [udp sum ok] RADIUS, length: 68
	Access-Request (1), id: 0x96, Authenticator: 0cb68e18648af64b33904a0abff0af6d
	  Service-Type Attribute (6), length: 6, Value: Authenticate Only
	    0x0000:  0000 0008
	  User-Name Attribute (1), length: 12, Value: XXX
	    0x0000:  6769 6274 736e 6963 6874
	  User-Password Attribute (2), length: 18, Value:
[...]

PR #1736, opened 2021-12-21
check_radius: avoid two NAS-IP-Address pairs if radiusclient-ng is used

ghciv6:
fix for #1607

### DIFF
--- a/plugins/check_radius.c
+++ b/plugins/check_radius.c
@@ -190,6 +190,9 @@ main (int argc, char **argv)
 			die (STATE_UNKNOWN, _("Invalid NAS-Identifier\n"));
 	}
 
+#if ! defined(HAVE_LIBRADCLI) && ! defined(HAVE_LIBFREERADIUS_CLIENT) && defined(HAVE_LIBRADIUSCLIENT_NG)
+	/* radiusclient-ng will always set NAS_IP_ADDRESS again. verified for version 0.5.6 */
+#else
 	if (nasipaddress == NULL) {
 		if (gethostname (name, sizeof(name)) != 0)
 			die (STATE_UNKNOWN, _("gethostname() failed!\n"));
@@ -200,6 +203,7 @@ main (int argc, char **argv)
 	client_id = ntohl (((struct sockaddr_in *)&ss)->sin_addr.s_addr);
 	if (my_rc_avpair_add (&(data.send_pairs), PW_NAS_IP_ADDRESS, &client_id, 0) == NULL)
 		die (STATE_UNKNOWN, _("Invalid NAS-IP-Address\n"));
+#endif
 
 	my_rc_buildreq (&data, PW_ACCESS_REQUEST, server, port, (int)timeout_interval,
 	             retries);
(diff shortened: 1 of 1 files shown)


weiss:
Thanks!

So this happens with radiusclient-ng specifically, but not with the FreeRADIUS Client library?  Isn't radiusclient-ng dead for way over a decade?

ghciv6:
i tried to replicate #1607 and radiusclient-ng was the library available for my system. i got the same issue and the behaviour detected was able to explain all described tests of the original poster.

i have no idea if "radiusclient-ng dead for way over a decade". this plugin works with 4 different radius libraries and i will not try to bring them all to work with my system while the issue ist replicated with the first tested library.

weiss:
My question wasn't meant as a criticism of your work in any way, and it wasn't necessarily directed at you specifically.  The question is whether we'd want to merge that change, which _only_ affects radiusclient-ng users, as-is.  In my book, the answer depends on whether that library is still in use.  If it isn't, maybe we should rather remove radiusclient-ng support altogether.  Because as you say:

> this plugin works with 4 different radius libraries and i will not try to bring them all to work with my system

That's precisely my problem when reviewing patches such as this one.

weiss:
> In my book, the answer depends on whether that library is still in use.

And (as I said), the next question would be whether FreeRADIUS users are affected by #1607 as well.  Maybe the answer is "yes", in which case this patch wouldn't actually fix that issue.  So while I'm not saying that _you_ should do the work of figuring out these things, I do think that _someone_ should before we merge this.

ghciv6:
how you will detect if a library is still used? 

i checked the code for "-V". It doesn't contain a hint with which library the plugin was compiled. it could help, if this info is easier available. else we can only ask the issue reporter for ldd output, right? 

waiting for "someone" to check any other library option will delay this PR probably forever. that will make my analysis and the fix finally useless. that's why i'm not happy with your annotation.

the PR fixes a real problem with this combination of software. and the fix will only change the code for that specific combination.

weiss:
> how you will detect if a library is still used?

Others might have some insight as to which of those libraries are still offered by the popular Linux/BSD distributions, for example.

> waiting for "someone" to check any other library option will delay this PR probably forever.

Maybe "someone" could at least check FreeRADIUS, which, as far as I know, is the only supported library that's still maintained, and therefore probably the most important to address.

> the PR fixes a real problem with this combination of software. and the fix will only change the code for that specific combination.

Yes, I understood as much.  What I'm not sure about is:

1. Can #1607 be closed if we merge this patch, and:
2. Do we still want to support this combination of software, or rather take the chance to remove it altogether to avoid the maintenance cost in the future?

waja:
> Isn't radiusclient-ng dead for way over a decade?

From my POV, radiusclient-ng is dead: https://tracker.debian.org/pkg/radiusclient-ng

waja:
> Maybe "someone" could at least check FreeRADIUS, which, as far as I know, is the only supported library that's still maintained, and therefore probably the most important to address.

In Debian we have (as far as I know) two radius library: [radcli](https://packages.debian.org/source/bookworm/radcli) and [freeradius](https://packages.debian.org/source/bookworm/freeradius).

The monitoring-plugins package supports both for building while radcli is the [preferred](https://salsa.debian.org/nagios-team/pkg-monitoring-plugins/-/blob/1fe8724bfef4d82dde5e4b1402858be680ef689d/debian/control#L9) one.

RincewindsHat:
Decision: since `radiusclient-ng` is dead ([merged into FreeRadius](https://wiki.freeradius.org/project/Radiusclient) ) we remove any support for `radiusclient-ng`.
Therefor this PR gets closed, too.

PR #1737, opened 2022-01-09
Point to Icinga Exchange instead of dead Monitoring Exchange

RincewindsHat:


waja:
Seems reasonable. Even, I don't like adjusting more things in one commit/PR. :)

RincewindsHat:
Sorry, those were some trivial whitespace thingies and I couldn't resist to `sed` them. Should I change them back?

waja:
> Sorry, those were some trivial whitespace thingies and I couldn't resist to `sed` them. Should I change them back?

Nope. But next time it would be great if you can split this into separate commits, better issues.

Please don't get me wrong. It's great that this all is fixed. But with my (Debian) maintainers hat on, sometimes it's needed to patch specific issues (upstream) and if you have mixed commits, it might be needed to fiddle things out of one patch to address just the exact problem (to get things upstream accepted for example).

Thanks, Jan.

RincewindsHat:
You are right, of course. I Will try to keep that in mind for the future. If you are happy otherwise, could you review the change?

waja (approved):
Removed taring whitespaces are fine. Migrating away from monitoringexchange.org as it is dead seems also reasonable.

PR #1741, opened 2022-01-25
add --queryname parameter to check_pgsql

datamuc:
This is used in the long output instead of the actual query.
So instead of

    OK - 'select stuff from various, tables where some_stuff is null and other_stuff is not null' returned 42

one can use --queryname=check_greatest_basket and it will print

    OK - check_greatest_basket returned 42

That's nicer for alerting purposes, at least in our use case.


RincewindsHat (approved):

Issue #1684, opened 2021-05-11
Feature:  maximum redirection infinite


duven87:
Hello! 
It would be great if you could change the maximum number of redirects allowed to infinity and not limit them to 15.
I would like to be able to ignore this value because I am interested in the string but not being able to change the value always stays in Icinga/Nagios as a warning.

PR #1744, opened 2022-01-29
check_http and check_curl: added --max-redirs=N option (feature #1684)

andreasbaumann:
Did it for both check_http and check_curl. Only a long option (out of reasonable
short options IMHO).


sni (approved):
looks good, thanks

PR #1745, opened 2022-01-29
check_disk: Clarify usage possibilites 

RincewindsHat:
Clarify that `check_disk` can be used with either percentage OR absolut limits.


sni (approved):

sni:
looks good

PR #1746, opened 2022-01-29
check_disk: Description for -M was the wrong way around

RincewindsHat:
Using -M should show the mountpoint instead of the device the
file system originated from.
Seems like this was not the case for a long time and now
the default is to show the mount point. Using `-M` reverts
to showing the (block) device instead.
The usage Description was adjusted with this commit.


waja (commented):
Does it make sense to still use `--mountpoint` even if this option show NOT the mountpoint but the block device?

RincewindsHat:
No, but removing or renaming it will break compatibility. Would be fine by me, though

waja:
@sni what do you think?

sni:
i guess there is no sane way to fix this. I definitely would keep the options the way they are for compatibility reasons, but fixing the documentation seems to be a good idea.

waja (approved):

RincewindsHat:
fixes #1746

PR #1749, opened 2022-01-31
check_disk: Add synthetic file systems for default exclusion depending on command line options.

### LINKED ISSUE #1748: check_disk: Selectively exclude squashfs, tracefs and tmpfs by default.
Hi,

In Ubuntu we're taken up the issue of monitoring-plugins#1608 (false positive DISK CRITICAL error), as we're encountering some DISK UNKNOWN messages as a consequence (https://bugs.launchpad.net/ubuntu/+source/monitoring-plugins/+bug/1940916). This bug is a side-effect of bug 1827159, where the synthetic filesystems (tmp, squash, trace) were excluded by default to avoid false positives: if the user wants to explicitly check such filesystem (by including them in the command line) or check a mount point based on one of them, the result was DISK UNKNOWN (as it was excluded by default and exclusion takes precedence over inclusion).

To strike a balance between the two, we propose here a solution that works on Ubuntu and that we are considering accepting. The approach here has been to exclude this type of filesystem only if it is neither included via path (option -p) nor via include (option -N).

 To minimize some possible impact, these changes are located after all the arguments logic has been done, not altering what the user has requested via command line.

mirespace:
Hi, 

to avoid false-positive DISK CRITICAL error and DISK UNKNOWN message when using check_disk with tracefs, squasfs and tmpfs (synthetic filesystems), the approach here has been to exclude this type of filesystem only if it is neither included via path (option -p) nor via include (option -N).  I think it can help in cases that scripts are in place and flags cannot be wrapped.

Fixes #1748.

### DIFF
--- a/plugins/check_disk.c
+++ b/plugins/check_disk.c
@@ -445,7 +445,7 @@ double calculate_percent(uintmax_t value, uintmax_t total) {
 int
 process_arguments (int argc, char **argv)
 {
-  int c, err;
+  int c, err, i;
   struct parameter_list *se;
   struct parameter_list *temp_list = NULL, *previous = NULL;
   struct mount_entry *me;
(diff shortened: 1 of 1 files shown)


EnanTompil (approved):

RincewindsHat:
hm, is there any possible scenario in which I would l like to have those not exluded by default?

RincewindsHat:
@waja planning to replace this with #1835

PR #1750, opened 2022-02-12
check_uptime: Add option to report uptime in days instead of seconds

amotl:
Dear community. Dear @wopfel and @sni.

Firsts things first: Thanks a stack for conceiving and maintaining this excellent collection of monitoring plugins.

This is a humble patch for the `check_uptime.pl` sensor to make it report the system uptime in _days_ instead of _seconds_. It has been wired to a new `--days` option, in order not to change its default behavior.

The rationale is to reduce mental load for humans, when looking at the textual output of the sensor. While the _"pretty" format_ already implemented by `check_uptime.pl` has good intentions, it didn't quite nail it for us.

With kind regards,
Andreas.

---

### Previous situation

Currently, the baseline plugin output, with no way to change it, is:
```
CRITICAL: Uptime is 38829029 seconds.
```

We also tried using both options `--for` and `--since`. However, we still have not been happy with the corresponding output.

image

### New situation

When using the proposed `--days|-d` option, the output will be:
```
CRITICAL: Uptime is 449 days.
```

In our opinion, this is much more pleasant.

image

### DIFF
--- a/plugins-scripts/check_uptime.pl
+++ b/plugins-scripts/check_uptime.pl
@@ -25,7 +25,7 @@ use POSIX;
 use strict;
 use Getopt::Long;
 use vars qw($opt_V $opt_h $opt_v $verbose $PROGNAME $opt_w $opt_c
-					$opt_f $opt_s
+					$opt_f $opt_s $opt_d
 					$lower_warn_threshold $upper_warn_threshold
 					$lower_crit_threshold $upper_crit_threshold
 					$status $state $msg);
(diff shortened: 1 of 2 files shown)


sni:
Looks good. Could you please add a help entry as well?

sni (changes requested):
Looks fine. Except, please add a help entry for the new option.

amotl:
Dear Sven,

thank you for the quick reply. I've just amended this patch according to your suggestions and also added a corresponding test case. Fixing a one-letter typo has been diverted to #1751.

With kind regards,
Andreas.

plugins-scripts/check_uptime.pl
```diff
@@ -262,6 +274,7 @@ ()
 	print "-c (--critical)  = Min. number of uptime to generate critical alert ( w < c )\n";
 	print "-f (--for)       = Show uptime in a pretty format (Running for x weeks, x days, ...)\n";
 	print "-s (--since)     = Show last boot in yyyy-mm-dd HH:MM:SS format (output from 'uptime -s')\n";
+	print "-s (--days)      = Show uptime in days\n";
```
sni, line 277: this should be `-d` i guess?

sni (approved):
please fix the last typo

plugins-scripts/check_uptime.pl
amotl: Oh my, it was late at night ;]. Thanks. Fixed!

sni:
great, thanks.

PR #1753, opened 2022-03-25
added support for opensmtpd to check_mailq.pl

ichdasich:
Adds support for opensmtpd, strictly working with the mailq command. Related to #1495, but submitting this as a subsequent pull requests depends on this change.

### DIFF
--- a/plugins-scripts/check_mailq.pl
+++ b/plugins-scripts/check_mailq.pl
@@ -393,6 +393,61 @@ elsif ( $mailq eq "postfix" ) {
                 #}
         }
 } # end of ($mailq eq "postfix")
+elsif ( $mailq eq "opensmtpd" ) {
+
+	 ## open mailq
+	if ( defined $utils::PATH_TO_MAILQ && -x $utils::PATH_TO_MAILQ ) {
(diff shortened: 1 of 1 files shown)


RincewindsHat:
ok, finally came around to testing this and on a debian 11 this doesn't work out of the box (or I am doing something wrong).
If I understand correctly, `mailq` is not provided by the package `opensmtp`. What actually works though, is using `smtpctl show queue` on debian.
Now, I don't know how common that setup is.
I guess we have to test whether `PATH_TO_MAILQ` is empty and search for `smtpctl` then, execute that and use the output.

RincewindsHat:
```
                if ( $utils::PATH_TO_MAILQ eq "" ) {
                        # Got no Mailq
                        if (! open (MAILQ, "$sudo /usr/sbin/smtpctl show queue | ")) {
                                print "ERROR: could not open  $sudo /usr/sbin/smtpctl\n";
                                exit $ERRORS{'UNKNOWN'};
                        }
```
something like that, but probably less prone to breaking

waja:
@ichdasich what @RincewindsHat might want to say ... can you maybe try to implement his suggested changes? ;)

RincewindsHat:
Sorry for the delay and all. I just merged #2330 which implements the same functionality, therefor I am going to close this

Issue #1099, opened 2013-09-24
Please add "use sudo" option to check_mailq  [sf#3573821]


monitoring-user:
Submitted by chris_schultz on 2012-10-02 17:11:04

Using check_mailq 1.4.15.

The check_mailq plug-in calls /usr/bin/mailq which, in my environments, requires root access. This is easily accomplished by giving nagios sudo access to /usr/bin/mailq but check_mailq needs to call sudo in order to use those privileges.

I have made a small patch that accomplishes this goal with an additional parameter (-s or --sudo). The user still needs to edit /etc/sudoers to give the nagios user permissions to execute /usr/bin/mailq without a password.

I attached my patch to my original issue report over at the old Nagios plugins tracker here: http://tracker.nagios.org/view.php?id=368

PR #1754, opened 2022-03-25
Add doas support to check_mailq.pl

ichdasich:
Adds doas support to check_mailq.pl (following that already in for sudo) for similar reasons as #1099.

Depends on #1753.


RincewindsHat:
Sorry for the lack of response here for such a long time.
I just removed the sudo functionality from `check_mailq` since it was a security problem
and also not necessary (MTAs allow access to the mail queues via different methods).

For the same reasons this here will also not be added.

Sorry again :-|

PR #1755, opened 2022-03-25
remove duplicate W=i/C=i args in check_mailq.pl

ichdasich:
currently, check_mailq.pl defines the C/W arg twice (slipped in in a single commit in 2020, only adding those duplicates), leading to warnings on plugin execution.


RincewindsHat:
indeed, don't know what I tought there.

RincewindsHat (approved):
Looks good.

RincewindsHat:
Thank you!

Issue #1752, opened 2022-03-23
check_icmp: floating constant exceeds range of ‘double’


andreasbaumann:
I get: 

```
check_icmp.c: In function ‘finish’:
check_icmp.c:1223:28: warning: floating constant exceeds range of ‘double’ [-Woverflow]
 1223 |                            (targets > 1) ? host->name : "", (host->rtmin < DBL_MAX) ? (float)host
      |                            ^
check_icmp.c: In function ‘add_target_ip’:
check_icmp.c:1326:9: warning: floating constant exceeds range of ‘double’ [-Woverflow]
 1326 |         host->rtmin = DBL_MAX;
      |         ^~~~
```

This warning goes away if I include <float.h> in check_icmp.c.

I don't think defining DLB_MAX here (when undefined) is really good idea:

```
#ifndef DBL_MAX
# define DBL_MAX 9.9999999999e999
#endif
```

sni:
probably a leftover or workaround for a >20 year old unix flavor.

RincewindsHat:
so, remove it or fix it?

PR #1757, opened 2022-03-26
Check icmp compiler warnings

RincewindsHat:
Resolves: #1752

@sni @andreasbaumann : Better this way?


andreasbaumann:
Yep, all ok. :-)

PR #1760, opened 2022-03-26
Rename output to cmd_output

RincewindsHat:
The name of the type `output` is hugely misleading, so I changed it to `cmd_output`.


RincewindsHat:
@sni @waja Comments?

sni:
tbh, i don't see how the name output is hugely misleading?

RincewindsHat:
I guess I went through that code at the time and was wondering how the output (think "stuff this programm should print") was in there and why it was handled this way.
It took me some time to understand, that it is the output of another command.
I guess, that annoyed me a bit.

waja:
Another change that makes it harder to integrate patches from np ...

RincewindsHat:
ok ok, I was just very irritated

PR #1762, opened 2022-04-10
Enhancement --continue-after-certificate (backport from nagios-plugins)

andreasbaumann:
Also added to check_curl..


sni (approved):
looks good, thanks.

PR #1770, opened 2022-06-03
Cdmiub

CDMIUB:
In my case some NAS devices need to spin up their desks before answering the request. This causes the test to time out many times in a row, while the NAS is not actively used. This causes _unnecessary_ alarms.
I styled the changes along check_ircd.pl


RincewindsHat (approved):
Looks good to me, thank you very much

PR #1771, opened 2022-06-05
check_icmp: Fix "Invalid Argument" from sendmsg() under FreeBSD 13.1 and "setsockopt failed" for TTL setting

### LINKED ISSUE #1769: check_icmp fails on FreeBSD 13.1
Starting it from the command line with:

./check_icmp -vvvvvvH <some-ip>

fails with:

Sending ICMP4 echo-request of len 24, id 7765, seq 0, cksum 0x3AC5 to host <some-ip>
Failed to send ping to <some-ip>: Invalid argument

See

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=264342

for details.

eriksejr:
Should fix #1769


opsec (approved):
Tested on one FBSD 13.1 system, works fine.

RincewindsHat:
ah, ok. I  will setup 13.1 then, when I find the time

RincewindsHat (approved):

Issue #1123, opened 2013-09-24
check_by_ssh exit status when reaching timeout [sf#2859548]


monitoring-user:
Submitted by guillomovitch on 2009-09-15 22:30:06

check_by_ssh currently exist with CRITICAL status when reaching timeout. When used just as a transport mechanism for invoking a remote plugin, it should rather exit with UNKNOWN status in this case.

The patch is trivial, I'd be happy to contribute it if this is agreed.

barry-luijten:
> There is an easy workaround: you can achieve the same effect by adding the -oConnectTimeout=XX flag where XX is slightly less than the timeout set with the -t flag.

The workaround results in an exit code of 255, which is still translated into Critical by Nagios and mod_gearman. The result message however starts with "UNKNOWN", so this is a cosmetic workaround imho.
I would really appreciate it if the -u switch will be implemented for check_by_ssh...

archiecobbs:
> The workaround results in an exit code of 255, which is still translated into Critical by Nagios and mod_gearman. The result message however starts with "UNKNOWN", so this is a cosmetic workaround imho.
I would really appreciate it if the -u switch will be implemented for check_by_ssh...

Done - see my PR #1774.

RincewindsHat:
Resolved by #1123

RincewindsHat:
#1774  ok, yes :-)

PR #1774, opened 2022-07-02
check_by_ssh: Add "-U" flag (#1123).

archiecobbs:
This causes a 255 exit value from ssh(1), which indicates a connection failure,
to return UNKNOWN instead of CRITICAL; similar to check_nrpe's "-u" flag.

### DIFF
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -50,6 +50,7 @@ unsigned int services = 0;
 int skip_stdout = 0;
 int skip_stderr = 0;
 int warn_on_stderr = 0;
+bool unknown_timeout = FALSE;
 char *remotecmd = NULL;
 char **commargv = NULL;
 int commargc = 0;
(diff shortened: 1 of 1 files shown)


RincewindsHat:
Resolves #1123 

Thank you very much for the PR.
I tried it and it seems, that it will change the behaviour only if `stderr` is ignored, since `ssh (1)` will output a message which will cause a `STATE_UNKNOWN` if `skip_stderr` is not present.
Seems like a valid use case anyway.
I would like one small change though, could you use to the `bool` type for the `unknown_timeout` variable to make the semantics more obvious?

plugins/check_by_ssh.c
```diff
@@ -50,6 +50,7 @@ unsigned int services = 0;
 int skip_stdout = 0;
 int skip_stderr = 0;
 int warn_on_stderr = 0;
+int unknown_timeout = 0;
```
RincewindsHat, line 53: Please change the type to `bool` to make the semantics clearer

RincewindsHat (changes requested):

archiecobbs:
> it will change the behaviour only if stderr is ignored, since ssh (1) will output a message which will cause a STATE_UNKNOWN if skip_stderr is not present.

Not sure what you're talking about. The added patch happens on [line 105](https://github.com/archiecobbs/monitoring-plugins/blob/ca4b657381abdc760b8a54f74b8983877f506be9/plugins/check_by_ssh.c#L105) which is prior to [line 125](https://github.com/archiecobbs/monitoring-plugins/blob/ca4b657381abdc760b8a54f74b8983877f506be9/plugins/check_by_ssh.c#L125) where `skip_stderr` takes effect.

> I would like one small change though, could you use to the bool type for the unknown_timeout variable to make the semantics more obvious?

Good idea - thanks. Done.

RincewindsHat (approved):

RincewindsHat:

PR #1775, opened 2022-07-03
Update CodeQL and update runner before installing

RincewindsHat:
* Update QuodeQl test to version 2 to fix the complaining in the test runs ( https://github.blog/changelog/2022-04-27-code-scanning-deprecation-of-codeql-action-v1/ )
* Update test machine package cache before installing stuff to avoid trying to fetch old ( and not available ) packages


waja (approved):
LGTM

PR #1778, opened 2022-07-24
Display total and scaled load values if check_load scales the values by number of CPUs

RincewindsHat:
Replaces #1548

### DIFF
--- a/plugins/check_load.c
+++ b/plugins/check_load.c
@@ -70,7 +70,7 @@ double cload[3] = { 0.0, 0.0, 0.0 };
 #define la15 la[2]
 
 char *status_line;
-int take_into_account_cpus = 0;
+bool take_into_account_cpus = false;
 
 static void
 get_threshold(char *arg, double *th)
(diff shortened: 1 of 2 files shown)


plugins/check_load.c
```diff
@@ -211,9 +206,26 @@ main (int argc, char **argv)
...
+			char *tmp = NULL;
+			xasprintf(&tmp, _("load average: %.2f, %.2f, %.2f"), scaled_la[0], scaled_la[1], scaled_la[2]);
+			xasprintf(&status_line, "scaled %s - %s", tmp, status_line);
+	}
+
 	printf("LOAD %s - %s|", state_text(result), status_line);
-	for(i = 0; i < 3; i++)
-		printf("load%d=%.3f;%.3f;%.3f;0; ", nums[i], la[i], wload[i], cload[i]);
+	for(i = 0; i < 3; i++) {
+		if (scaled_la[i] != 0.0) {
```
sni, line 222: i don't think it's a good idea to make printing the performance values depend on the actual value. Either you want it or not, but don't magically hide some values if they are zero.

plugins/t/check_load.t
```diff
@@ -11,8 +11,10 @@ use NPTest;
 my $res;
 
 my $loadValue = "[0-9]+\.?[0-9]+";
-my $successOutput = "/^LOAD OK - load average: $loadValue, $loadValue, $loadValue/";
-my $failureOutput = "/^LOAD CRITICAL - load average: $loadValue, $loadValue, $loadValue/";
+my $successOutput = "/^LOAD OK - total load average: $loadValue, $loadValue, $loadValue/";
+my $successScaledOutput = "/^LOAD OK - scaled load average: $loadValue, $loadValue, $loadValue - total load average: $loadValue, $loadValue, $loadValue/";
```
sni, line 15: is this used anywhere?

plugins/check_load.c
RincewindsHat: This is completely true, this was the "poor man's version" of testing whether the scaled values are actually in usage or not.
Solved this another way now.

plugins/t/check_load.t
RincewindsHat: It is now :-)

sni (approved):

PR #1779, opened 2022-07-24
check_swap: Fix unit for total in perfdata

RincewindsHat:
Seems someone (me) forgot to fix the unit for totals in the perfdata, after changing things last time.
Mea culpa.


sni (approved):
looks good

PR #1780, opened 2022-07-25
fix parsing swap values

sni:
tmp_KB changed from float to uint64, so change the sscanf format accordingly.


RincewindsHat (approved):
Good catch

PR #1793, opened 2022-09-30
Check apt usage

### LINKED ISSUE #1066: check_apt: clarify that it doesn't upgrade per default [sf#3539328]
Submitted by calestyo on 2012-07-01 21:27:21

(Roughly) From: http://tracker.nagios.org/view.php?id=302

Hi.

check_apt uses its "--upgrade" option per default:
 -U, --upgrade=OPTS
    [Default] Perform an upgrade. If an optional OPTS argument is provided,
    apt-get will be run with these command line options instead of the
    default (-o 'Debug::NoLocking=true' -s -qq).
    Note that you may be required to have root privileges if you do not use
    the default options.

I was quite confused by this, given the meanings of upgrade (install new packages) and update (get information about new packages) in Debian.

The following seems to be true:
- Using the upgrade mode in APT (and therefore the --upgrade option to the plugin) is really required, even to just get a list of upgradable packages.
  So this really needs to be the default.
- No upgrades (in the sense of installing packags) are however executed, given the options  (-o 'Debug::NoLocking=true' -s -qq). This is true for both, normal users (obviously) and root.

In order to clear things up, I'd suggest to add a notes that:
- the --upgrade option of the plugins (with it's default OPTS) doesn't do package upgrades (in the sense of installing new ones), neither as root nor as normal user.

and optionally:
- the --update option (and upgrade mode of APT) is needed even to just get a list of upgradable pacakges.

Cheers,
Chris.

RincewindsHat:
Resolves #1066 

* Clarify whether `check_apt` will try to perform an upgrade or not
* Use real booleans
* Fix some comment errors
* Fix some trailing whitespace stuff


waja:
Looking into it, changes seems reasonable.

waja (approved):

PR #1796, opened 2022-10-19
check_mailq: Fixing nullmailer regex

waja:
In Debian we got reported this issue: https://bugs.debian.org/1019406

> it seems that the output from nullmailer's mailq has changed format since
that portion was implemented.  i noticed that it was not alerting on a
host w/many messages in its queue:
>
> % mailq | wc -l
12488
% check_mailq -M nullmailer -w 5 -c 10
>OK: nullmailer mailq (0) is below threshold (5/10)|unsent=0;5;10;0
>
> attached is a patch that updates the format expected in the nullmailer
mailq output.  the regex is a little more flexible and less specific
than the previous version.


RincewindsHat (approved):
Compared that to https://github.com/bruceg/nullmailer/blob/master/test/tests/mailq . Looks good to me.

PR #1798, opened 2022-10-20
Remove superflous CRLF in HTTP-Requests in check_http

### LINKED ISSUE #1576: check_http sending extra CRLF after POST data
In Debian we got reported this issue: https://bugs.debian.org/913142

> This is a known bug which could lead to security issues (disabled WAF checks to allow requests). cf : https://github.com/nagios-plugins/nagios-plugins/issues/266

This was also reported earlier in https://github.com/monitoring-plugins/monitoring-plugins/issues/143.

By nagios-plugins this was fixed via a _￼Minor two-line fix_: https://github.com/nagios-plugins/nagios-plugins/pull/298

I'm a bit late on the party, but Debian is starting to prepare the next stable release and on 12th Feb 2019 it's starting the [Soft Freeze](https://release.debian.org/buster/freeze_policy.html). It would be good to get that fixed until then.

RincewindsHat:
Resolves #1576
Trying to resolve https://github.com/monitoring-plugins/monitoring-plugins/issues/1576 . Can somebody take a look and see if it resolves the problem with very strict webservers?

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -966,7 +966,8 @@ check_http (void)
     }
     asprintf (&buf, "%sProxy-Connection: keep-alive\r\n", buf);
     asprintf (&buf, "%sHost: %s\r\n", buf, host_name);
-
+    /* we finished our request, send empty line with CRLF */
+    asprintf (&buf, "%s%s", buf, CRLF);
     if (verbose) printf ("%s\n", buf);
     send(sd, buf, strlen (buf), 0);
     buf[0]='\0';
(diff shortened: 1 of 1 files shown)


waja:
I released 2.3.2-1 in Debian with the following patch: https://salsa.debian.org/nagios-team/pkg-monitoring-plugins/-/raw/debian/2.3.2-1/debian/patches/13_check_http_fix_superflous_crlf

Dunno if removing the first CRLF destroys anything.

sni:
I'd say this patch is the right thing to do. It's plain wrong to send a content-length header and then send something with that size + 1.
If any webserver requires that extra crlf, it should be added to the post data.

sni (approved):

PR #1800, opened 2022-10-21
using PRId64 and PRIu64 instead of %ld directly

andreasbaumann:


RincewindsHat (approved):

RincewindsHat:
Thank you very much.

andreasbaumann:
perfdata_uint64: has some PRId64 instead of PRIu64. (line 618, 620).. I'll make another pull request.. Sorry about that..

PR #1802, opened 2022-10-24
fixed two PRId64 to PRIu64 in perfdata_uint64

andreasbaumann:


RincewindsHat (approved):

RincewindsHat:
Should totally have noticed that in the first one... Mea culpa
Thanks again :-)

andreasbaumann:
Sorry for the fast double patching. :-)

PR #1803, opened 2022-10-25
check_pgsql: check of database name is too strict

waja:
In Debian we got the following [bug](http://bugs.debian.org/982847) reported:

Hi *,
to reproduce create a Database called freshports.git and try to
use it with `check_pgsql`:

```
flo@p5:/tmp/monitoring-plugins-2.3$ /usr/lib/nagios/plugins/check_pgsql -d freshports.devgit -l flo
check_pgsql: Database name is not valid - freshports.devgit
Usage:
check_pgsql [-H <host>] [-P <port>] [-c <critical time>] [-w <warning time>]
 [-t <timeout>] [-d <database>] [-l <logname>] [-p <password>]
[-q <query>] [-C <critical query range>] [-W <warning query range>]
```
```
flo@p5:/tmp/monitoring-plugins-2.3$ psql freshports.devgit
psql (13.1 (Debian 13.1-1+b1))
Type "help" for help.

freshports.devgit=# \d
Did not find any relations.
freshports.devgit=# 
```

The problem is that `check_pgsql` validates the Database name and has different assumptions
that postgres itself.

I fail to see a reason to validate the database name here. Postgres'es API should
do this - So i would suggest a fix like this by removing `is_pg_dbname` alltogether.

This is a followup of https://github.com/monitoring-plugins/monitoring-plugins/issues/1660#issuecomment-1288575020

### DIFF
--- a/plugins/check_pgsql.c
+++ b/plugins/check_pgsql.c
@@ -69,7 +69,6 @@ int process_arguments (int, char **);
 int validate_arguments (void);
 void print_usage (void);
 void print_help (void);
-int is_pg_dbname (char *);
 int is_pg_logname (char *);
 int do_query (PGconn *, char *);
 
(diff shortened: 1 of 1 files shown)


sni:
Sounds fine for me, except... what about still checking strlen? I'd think an error message is better than silently cutting of the name.

RincewindsHat:
@waja Could you add this:
```
--- a/plugins/check_pgsql.c
+++ b/plugins/check_pgsql.c
@@ -346,6 +346,9 @@ process_arguments (int argc, char **argv)
                                pgport = optarg;
                        break;
                case 'd':     /* database name */
+                       if (strlen(optarg) >= NAMEDATALEN) {
+                               usage2 (_("Database name exceeds the maximum length"), optarg);
+                       }
                        snprintf(dbName, NAMEDATALEN, "%s", optarg);
                        break;
                case 'l':     /* login name */
```
I think I agree with @sni here, postgres would automatically cut off the name of the database and I would prefer to get an error message instead of some silent modification.

waja:
> @waja Could you add this:

@RincewindsHat did as advised. :)

RincewindsHat (approved):

RincewindsHat:
Thank you very much :-)

Issue #1808, opened 2022-11-05
check_curl: Invalid URL format if used with IPv6


RincewindsHat:
So I tried something like `./check_curl '-I' '::1'` (or `./check_curl '-H' '::1'` and it breaks before even connecting due to the invalid URL format [RFC3986](https://www.ietf.org/rfc/rfc3986.txt). This is saved in the `url` variable and generated from `server_address` ( [link](https://github.com/monitoring-plugins/monitoring-plugins/blob/2d9c6276d0a380f831bf94734359f071d0c7e958/plugins/check_curl.c#L480) ).

If given an IPv6 address, this address must be surround by square brackets (`[]`).

RincewindsHat:
Also in [this line](https://github.com/monitoring-plugins/monitoring-plugins/blob/2d9c6276d0a380f831bf94734359f071d0c7e958/plugins/check_curl.c#L482): Why is the hostname only set, if we use TLS? Shouldn't that be independent of that?

RincewindsHat:
Resolved by #1809

PR #1809, opened 2022-11-05
Check curl detect ipv6

RincewindsHat:
If an IPv6 address is give as `server_address` it must be surround by square brackets.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
               ${{ matrix.distro }} \
               /bin/sh -c '${{ matrix.prepare }} && \
                 tools/setup && \
-                ./configure --enable-libtap && \
+                ./configure --enable-libtap  --with-ipv6=no && \
                 make && \
                 make test'
           docker container prune -f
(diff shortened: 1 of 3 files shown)


RincewindsHat:
Resolves #1808

sni:
Would it be possible to create a test case for this?

RincewindsHat:
good question, afaik IPv6 is disabled in our testing setup and IPv6 in docker is generally a pain. I will investigate.

RincewindsHat:
[This here](https://github.com/monitoring-plugins/monitoring-plugins/blob/edc84e5d669089fa3f047dc52740679e7cb58fc1/.github/prepare_debian.sh#L31) looks suspicios in that regard. I guess this is there to allow test using `localhost` as IP. Can we change that?

RincewindsHat:
@sni I am currently out of ideas for this one. I think IPv6 is just to new for docker and Microsoft. I fear, we have to skip the test for now.
Can we add it conditionally to have it in different test environments?

sni:
Sure, better than nothing :-)
There is a ipv6 test switch in check_tcp or check_dig already:
https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/t/check_tcp.t#L13

RincewindsHat:
@sni should be fine now

sni (approved):

sni:
looks good, thanks

PR #1811, opened 2022-11-07
Adding dependabot config which takes care of updating github actions

waja:


sni (approved):

sni:
nice, i think i gonna borrow that for my other projects :)

Issue #994, opened 2013-09-24
check_http: gzip deflate chunked problem [sf#3294169]


monitoring-user:
Submitted by pzlatcpp on 2011-04-28 04:24:57

check_http can't deal with compressed data by gzip or deflate and also can't surpport chunked data. This patch add some functions to deal with the above situations. We add two parameters to commandline -d/--decompress and -X/--chunked; when users want to surpport gzip and deflate data format the can use -d; when they want to deal with chunked data they can use  -X.

Plugin Version: check_http v1.4.15.23.g1bc7.dirty (nagios-plugins 1.4.15)
Exmaple Commandline: ./check_http -H www.aipai.com -p 80 -t 60 --header='Accept-Encoding: gzip,deflate' -u /hredu.html -v -d -X
Tested on Operating system: debian 6
Tested with complier: gcc 4.4.5

alaslums:
6 years later... tumbleweeds drift by...

```
.................................................
.                                   |           .
._-""--__                    '.  \  |  /  .'    .
.  |j  _-'"-__-"=         ~    '. \ | / .'      .
. L| -'   _-'    '=        _____-- .-. -______  .
.  | L|_,    . * ."=_-_________---' ~ '---______.
.     |      _\|/_             ,'~~~~~~~'  | ~-_.
.                      ~~~   <')  ~~~     L|,   .
.                  ~~~~~     (_\\          |    .
.                             ||\\              .
.           .-.               ((__))            .
.          (  o)-.            ( O O)            .
.           ) )\|\)            "\_\\.-~         .
.        _./ (,_    Yes,         "'             .
.       ( '**"  )  the Dead Cow restaurant used .
.       \\\   ///    to be here, but its been   .
.        \\\ ///    closed for quite some time. .
.         //|\\                                 .
.        ///|\\\                                .
..............................................MJP
```

sni:
please give the check_curl from the feature_check_curl branch a try.

RincewindsHat:
Chunked encoding is now implemented, deflate will probably not implemented at all. closing this therefore.

PR #1815, opened 2022-11-14
Check http chunked encoding

### LINKED ISSUE #1122: check_http asks for HTTP/1.1, but doesn't understand chunked [sf#2786808]
Submitted by wferi on 2009-05-04 21:19:43

Fixing 2638765 uncovered this bug (actually, most HTTP servers respect Host: headers in HTTP/1.0 requests, even though it isn't part of the spec).
Anyway: if a client sends a HTTP/1.1 request, it must be able to handle the incoming HTTP/1.1 reply. Which includes coping with a chunked one.
This exact same command worked with check_http version 1.4.5 (and still does so with -H replaced by -I, but that isn't the point).
```
$ /usr/lib/nagios/plugins/check_http -v -H idp.niif.hu -u /idp/profile/Status --ssl -r "^ok$"
GET /idp/profile/Status HTTP/1.1
Host: idp.niif.hu:443
User-Agent: check_http/v1991 (nagios-plugins 1.4.12)
Connection: close

https://idp.niif.hu:443/idp/profile/Status is 141 characters
STATUS: HTTP/1.1 200 OK
***\* HEADER ****
Date: Mon, 04 May 2009 19:12:13 GMT
Connection: close
Transfer-Encoding: chunked
Content-Type: text/plain
***\* CONTENT ****
2
ok
0

HTTP CRITICAL - pattern not found|time=0.067933s;;;0.000000 size=141B;;;0
```

### LINKED ISSUE #1533: check_http sometimes fails when Transfer-Encoding: chunked is used
Sometimes check_http goes into CRITICAL state when Transfer-Encoding: chunked is used. Then the check fails for a couple of minutes.

We have spent some time with debugging but the response we get from the webserver is always complete. If we choose another string to check the content the check is fine again. So it looks like check_http has a problem with reassembling the chunked response.

We are using nagios-plugins-http-2.2.1-4git.el7.x86_64 on CentOS Linux release 7.4.1708 (Core).

The used command is as followed (anonymized):
check_http -H "www.example.com" -I "192.168.90.107" -c 20 -e 200 -k "https:true" -p 8810 -s "Realisiert mit dem Government Site Builder" -u /Webs/Example/DE/Startseite/startseite_node.html -w 10

The flapping check output looks like this:
HTTP OK: Status line output matched "200" - HTTP/1.1 200 OK - 60925 bytes in 0.005 second response time |time=0.005078s;10.000000;20.000000;0.000000 size=60925B;;;0
HTTP CRITICAL: Status line output matched "200" - HTTP/1.1 200 OK - string 'Realisiert mit dem Governm...' not found on 'http://www.example.com:8810/Webs/Example/DE/Startseite/startseite_node.html' - 71720 bytes in 0.090 second response time |time=0.089791s;10.000000;20.000000;0.000000 size=71720B;;;0

Here you can find two straces that also show the problem:

[strace-check_http-s407-8810.gz](https://github.com/monitoring-plugins/monitoring-plugins/files/1987948/strace-check_http-s407-8810.gz)
[strace-check_http-s409-8810.gz](https://github.com/monitoring-plugins/monitoring-plugins/files/1987950/strace-check_http-s409-8810.gz)

The string is read in the server reply in strace:
"<!--Realisiert mit dem Government Site Builder"

But it still outputs:
write(1, "HTTP CRITICAL: Status line output matched "200" - HTTP/1.1 200 OK - string 'Realisiert mit dem Governm...' not found on 'http://www.example.com:8810/Webs/Example/DE/Startseite/startseite_node.html'

If you need further information, logs or anything else let me know.

Regards, Christian

RincewindsHat:
So, my take on implementing chunked encoding for check_http.
Tell me, where it's broken.

fix #1533
also related to #994
fix #1122

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -1267,11 +1267,15 @@ int check_http(void) {
   regmatch_t chre_pmatch[1]; // We actually do not care about this, since we only want to know IF it was found
 
   if (regexec(&chunked_header_regex, header, 1, chre_pmatch, 0) == 0) {
+    if (verbose) {
+      printf("Found chunked content\n");
+    }
     // We actually found the chunked header
     char *tmp = unchunk_content(page);
     if (tmp == NULL) {
       die(STATE_UNKNOWN, "HTTP %s: %s\n", state_text(STATE_UNKNOWN), "Failed to unchunk message body");
     }
+    page = tmp;
   }
 
   if (strlen(string_expect) > 0) {
@@ -1374,9 +1378,10 @@ char *unchunk_content(const char *content) {
   // https://www.rfc-editor.org/rfc/rfc7230#section-4.1
   char *result = NULL;
   size_t content_length = strlen(content);
-  char *start_of_chunk, end_of_chunk;
+  char *start_of_chunk;
+  char* end_of_chunk;
   long size_of_chunk;
-  char *pointer = content;
+  const char *pointer = content;
   char *endptr;
   long length_of_chunk = 0;
   size_t overall_size = 0;
@@ -1396,13 +1401,12 @@ char *unchunk_content(const char *content) {
       if (verbose) {
         printf("Chunked content did not start with a number at all (Line: %u)\n", __LINE__);
       }
-      return NULL
+      return NULL;
     }
 
     // So, we got the length of the chunk
     if (*endptr == ';') {
       // Chunk extension starts here
-      // TODO
       while (*endptr != '\r') {
         endptr++;
       }
@@ -1410,7 +1414,8 @@ char *unchunk_content(const char *content) {
 
     start_of_chunk = endptr + 2;
     end_of_chunk = start_of_chunk + size_of_chunk;
-    length_of_chunk = end_of_chunk - start_of_chunk;
+    length_of_chunk = (long)(end_of_chunk - start_of_chunk);
+    pointer = end_of_chunk + 2; //Next number should be here
 
     if (length_of_chunk == 0) {
       // Chunk length is 0, so this is the last one
@@ -1442,7 +1447,8 @@ char *unchunk_content(const char *content) {
     result_ptr = result_ptr + size_of_chunk;
   }
 
-  return result
+  result[overall_size] = '\0';
+  return result;
 }
 
 /* per RFC 2396 */
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -34,7 +34,7 @@
 /* splint -I. -I../../plugins -I../../lib/ -I/usr/kerberos/include/ ../../plugins/check_http.c */
 
 const char *progname = "check_http";
-const char *copyright = "1999-2013";
+const char *copyright = "1999-2022";
 const char *email = "devel@monitoring-plugins.org";
 
 #include "common.h"
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -57,8 +57,8 @@ enum {
 };
 
 #ifdef HAVE_SSL
-int check_cert = FALSE;
-int continue_after_check_cert = FALSE;
+bool check_cert = false;
+bool continue_after_check_cert = false;
 int ssl_version = 0;
 int days_till_exp_warn, days_till_exp_crit;
 char *randbuff;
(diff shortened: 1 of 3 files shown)


RincewindsHat:
@waja @sni

sni:
could you please remove the clang format commit. It really makes the diff unreadable.
Besides that, i don't think we should enforce 80col strictly on all code. But for now, i would
like to move those format changes into a new PR once this is merged.

RincewindsHat:
@sni Should be done, also rebased on master

plugins/check_http.c
```diff
@@ -1257,32 +1242,63 @@ check_http (void)
...
-        bcopy("...",&output_header_search[sizeof(output_header_search)-4],4);
+  if (strlen(header_expect) > 0) {
+    if (strstr(header, header_expect) == NULL) {
+      // We did not find the header, the rest is for building the output and setting the state
+      char output_header_search[30] = "";
+      strncpy(&output_header_search[0], header_expect,
+              sizeof(output_header_search));
+      if (output_header_search[sizeof(output_header_search) - 1] != '\0') {
+        bcopy("...", &output_header_search[sizeof(output_header_search) - 4],
+              4);
```
sni, line 1253: tbh, i don't think this is more readable than before.

plugins/check_http.c
```diff
@@ -57,8 +58,8 @@ enum {
 };
 
 #ifdef HAVE_SSL
-int check_cert = FALSE;
-int continue_after_check_cert = FALSE;
+bool check_cert = false;
```
sni, line 61: what's wrong with FALSE?

sni (approved):
looks good. Except i think those format changes are not necessary. Also a test case would be really nice.

plugins/check_http.c
RincewindsHat: IMHO, booleans should be "real" booleans, not aliased integers. Sure, it might be the same in the end, but this gives the compiler and future developers more context about what this is doing and what could/should possibly happen there.

I personally find it hugely confusing if an `int` variable or function turns out to be used for booleans only.

RincewindsHat:
A test case would be nice, but I don't have a good (meaning easy to implement/integrate) idea on how to do this.
For testing I wrote a go program on my end, but obviously that is hard to add to the current test infrastructure.
And I don't know how to force the testing webserver to use chunked encoding

sni:
lets see, i added a test case...

plugins/check_http.c
sni: ok, agreed

plugins/check_http.c
sni: i agree with the two if statements, but how is moving three characters to the next line making things more readable?

RincewindsHat:
@sni: I have been taking a look a the test case (which took me some time, since I do not know a lot of Perl) and I think they look good.
IMHO we can merge this if it is fine for you

sni:
alright, thanks a lot

sthen:
Hi, the chunked decoding doesn't take no_body into account; check a server which uses chunked with --no-body and you now get "HTTP UNKNOWN: Failed to unchunk message body".

Is it enough to just do something like this?

```
@@ -1278,7 +1278,7 @@ check_http (void)
 
   regmatch_t chre_pmatch[1]; // We actually do not care about this, since we only want to know IF it was found
 
-  if (regexec(&chunked_header_regex, header, 1, chre_pmatch, 0) == 0) {
+  if (!no_body && regexec(&chunked_header_regex, header, 1, chre_pmatch, 0) == 0) {
     if (verbose) {
       printf("Found chunked content\n");
     }
```

waja:
> Hi, the chunked decoding doesn't take no_body into account; check a server which uses chunked with --no-body and you now get "HTTP UNKNOWN: Failed to unchunk message body".
> 
> Is it enough to just do something like this?
> 
> ```
> @@ -1278,7 +1278,7 @@ check_http (void)
>  
>    regmatch_t chre_pmatch[1]; // We actually do not care about this, since we only want to know IF it was found
>  
> -  if (regexec(&chunked_header_regex, header, 1, chre_pmatch, 0) == 0) {
> +  if (!no_body && regexec(&chunked_header_regex, header, 1, chre_pmatch, 0) == 0) {
>      if (verbose) {
>        printf("Found chunked content\n");
>      }
> ```

@sthen how about opening an Issue about that?

PR #1817, opened 2022-11-20
check_by_ssh: Add random wait option -r

kompetenzbolzen:
Added option -r, --random-wait=MAXTIME to wait for a random amount of seconds between 0 and MAXTIME.

This should help with deployments which check multiple services over SSH. I found that adding a wrapper to wait randomly for up to 10s reduced the amount of failed SSH connections dramatically, and thus eliminated many false alarms.

### DIFF
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -94,12 +94,8 @@ main (int argc, char **argv)
 
 	/* Wait a random amount of time */
 	if (random_wait_max) {
-		uint8_t random = 0;
-		int fd = open("/dev/urandom", O_RDONLY);
-		read(fd, &random, sizeof(uint8_t));
-		close(fd);
-
-		sleep( (random_wait_max * random) / 255 );
+		srand(time(NULL));
+		sleep(random() % random_wait_max);
 	}
 
 	/* Set signal handling and alarm timeout */
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -35,6 +35,10 @@ const char *email = "devel@monitoring-plugins.org";
 #include "netutils.h"
 #include "utils_cmd.h"
 
+#include <stdint.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
 #ifndef NP_MAXARGS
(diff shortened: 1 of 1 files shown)


sni:
not sure if this is necessary to be builtin. Couldn't you simply do a `sleep $(($RANDOM%10)) && .../check_...`. Or even better, make use of ssh control master.
Ex.: like described here: https://labs.consol.de/nagios/omd/2012/06/12/running-check_by_ssh-over-a-persistent-ssh-connection.html

kompetenzbolzen:
I never tried ssh control master, thanks for the tip!
The original reason I built this was, that to wrap the check I needed to create a whole new check command, since I cant trivially overwrite the built-in one, which messed up the metric collection tagging based on the command name. 
I've seen this solution in some environments and I would think that adding a simple flag would be easier than wrapping the check.

sni:
ok, then my next question is why you use `dev/urandom` instead of `srand` and `random` from the stdlib. Ex.:
https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins-root/check_dhcp.c#L469

something like (untested):
```
srand(time(NULL));
sleep((random() % random_wait_max);
```

kompetenzbolzen:
Force of habit. There's no particular reason to use urandom here. I normally do because of its far better (at least afaik) randomness, but here thats ofc irrelevant. I'll change it to make it more uniform.

plugins/check_by_ssh.c
```diff
@@ -338,6 +350,11 @@ process_arguments (int argc, char **argv)
 			comm_append("-F");
 			comm_append(optarg);
 			break;
+		case 'r':
+			if (!is_integer (optarg))
```
RincewindsHat, line 354: I guess this should be a test, whether this is a positive integer.

RincewindsHat (changes requested):

RincewindsHat:
Sorry for the long wait. I am gonna make a decision here.
I think it is not a good idea to add this as a feature in the plugin and therefore close this PR.
Scheduling checks is a problem of the monitoring system and bad SSH should be circumvented elsewhere.

PR #1827, opened 2023-01-18
fix check_snmp regex matches

sni:
the multiplier function always tried to extract a number, even if the result is a string because of using a mib.

before:
```
./check_snmp -H hostname -P2c -c public -o IF-MIB::ifAdminStatus.11466 -vvv -r 0
/usr/bin/snmpget -Le -t 10 -r 5 -m ALL -v 2c [context] [authpriv] 10.0.13.11:161 IF-MIB::ifAdminStatus.11466
IF-MIB::ifAdminStatus.11466 = INTEGER: up(1)
Processing oid 1 (line 1)
  oidname: IF-MIB::ifAdminStatus.11466
  response:  = INTEGER: up(1)
SNMP OK - 0 | IF-MIB::ifAdminStatus.11466=0;;
```

the regexp 0 matches, even if the actual result is "up(1)".

after this patch:
```
./check_snmp -H hostname -P2c -c public -o IF-MIB::ifAdminStatus.11466 -vvv -r 0
/usr/bin/snmpget -Le -t 10 -r 5 -m ALL -v 2c [context] [authpriv] 10.0.13.11:161 IF-MIB::ifAdminStatus.11466
IF-MIB::ifAdminStatus.11466 = INTEGER: up(1)
Processing oid 1 (line 1)
  oidname: IF-MIB::ifAdminStatus.11466
  response:  = INTEGER: up(1)
SNMP CRITICAL - *up(1)* |
```

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -1165,17 +1165,36 @@ nextarg (char *str)
 char *
 multiply (char *str)
 {
-	double val = strtod (str, NULL);
-	val *= multiplier;
+	char *endptr;
+	double val;
 	char *conv = "%f";
+
+	if(verbose>2)
+		printf("    multiply input: %s\n", str);
+
+	val = strtod (str, &endptr);
+	if ((val == 0.0) && (endptr == str)) {
+		if(multiplier != 1) {
+			die(STATE_UNKNOWN, _("multiplier set (%.1f), but input is not a number: %s"), multiplier, str);
+		}
+		return str;
+	}
+
+	if(verbose>2)
+		printf("    multiply extracted double: %f\n", val);
+	val *= multiplier;
 	if (fmtstr != "") {
 		conv = fmtstr;
 	}
 	if (val == (int)val) {
 		sprintf(str, "%.0f", val);
 	} else {
+		if(verbose>2)
+			printf("    multiply using format: %s\n", conv);
 		sprintf(str, conv, val);
 	}
+	if(verbose>2)
+		printf("    multiply result: %s\n", str);
 	return str;
 }
 


RincewindsHat:
I would prefer to not use the `multiplier` as a global, I had to search why a `multiply` function gets only one argument.

sni:
I don't like it either, but that's how options are kept in check_snmp. I didn't want to rewrite the hole plugin.

RincewindsHat:
Fair enough. Is it surprising behaviour to just ignore the multiplier if snmp result is an enum and not fail loudly because the user tried to do something wrong?

sni:
it is not wrong from user perspective, the user requests an int value which maps to strings with the given mib and then checks by regex if the result is ok. That's a total valid use case.
We just cannot apply a multiplier in this case because the result is not a number anymore.

RincewindsHat:
hm, just took a look at the code. Previously I thought we just apply the multiplier everytime and therefore, in my thoughts it should be an error if we cannot do that, but given, that it is only applied if applicable that looks fine.

sni:
the multiplier has been introduced 4 weeks ago with #1813. This PR just restores the functionality when using a mib with -m.

RincewindsHat:
Correct me, if I am wrong, but does this patch just work because you didn't set the multiplier and there the function return early (good thing in any case)? Wouldn't be the fix to see if `strtod` fails or not?

sni:
sure, we can do this as well. Now i even added a nice error message if somebody sets a mib and a multiplier.

RincewindsHat (approved):
Looks good too me.

PR #1828, opened 2023-01-19
Added --exclude, cleanup args, fix -a count bug

lgmu:
Added --exclude to exclude patterns
Cleaned up duplicated code in the args
Fixed a bug when using --all because the count always returned "1" even when nothing matched

entry=$($GREP "$query" "$tempdiff")
count=$(echo "$entry" | wc -l)

Example:

$ touch testfile
$ TEST123=$(grep 'test' testfile)
$ echo "$TEST123" | wc -l
1


sni:
the patch looks fine, would be nice if you could add a test case for this new exclude feature.

lgmu:
Added some tests :)

sni:
Sorry, didn't realize there haven't been any tests before. Thanks for adding them.

sni (approved):

PR #1830, opened 2023-01-30
check_http: Fix memory reallocation error in chunk decoding logic

RincewindsHat:
This patch should fix an error with the way memory reallocation was used, which resulted in "realloc(): invalid next size". It is not completely clear to me as to what caused this problem, but apparently one can not depend handing a pointer to "realloc(3)" and expect that it still works afterwards, but one should/must use the one returned by the function.

Also this patch replaces a variable which was used to remember the position in the array by just computing that from the current values.


sni (approved):
looks good

PR #1835, opened 2023-02-03
check_disk: Add more filesystems to ignore by default

### LINKED ISSUE #1748: check_disk: Selectively exclude squashfs, tracefs and tmpfs by default.

RincewindsHat:
Replaces #1749 
Fixes #1748


waja:
good catch by the CI pipeline :)

RincewindsHat:
Anyone got an idea why that fails?

KriSchu:
> Anyone got an idea why that fails?

Tests and the test setup are absolutely not my expertise, so please excuse if this is completely stupid, but isn't it failing because, by your code change you exclude tmpfs filesystems, but the test pipeline adds tmpfs filesystems to the docker container which is running the tests? 
At least I think that this could explain, why the test expects two filesystems, but found 0 filesystems in the output:
```
#   Failed test 'Got two mountpoints in output'
#   at ./t/check_disk.t line 36.
#          got: 0
#     expected: 2
```

The other check is about the exit code being 3 (Unknown), while 0 (OK) is expected.
If both filesystems are excluded, that there is no filesystem to check and the plugin prints:
`DISK UNKNOWN - No disks were found for provided parameters;|`

I believe the test setup could explain these failed tests and that it can be solved by either changing to a different fs type or by modifying the tests.

RincewindsHat:
That sounds like a likely explanation. At this point I see a general problem with those changes here. I am not able to remove stuff from the exclusion list during runtime, so this is equivalent to saying "this plugin ignores those filesystem types ALWAYS" which is going to far IMHO.

waja:
@RincewindsHat so same counts for #1749?

RincewindsHat:
@waja sorry, this went under. Yes, same problem. Question is whether we want to ignore those filesystems ALL THE TIME.

RincewindsHat:
Hard coding the ignored filesystems is probably a bad idea in general.
There I am closing this again until I find a better solution.

Issue #1834, opened 2023-02-03
monitoring-plugins-2.3.3 fails to build with gnutls


hydrapolic:
Same for version 2.3.2 reported in Gentoo: https://bugs.gentoo.org/880211

```
./configure --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --d
atarootdir=/usr/share --disable-dependency-tracking --disable-silent-rules --disable-static --docdir=/usr/share/doc/monitoring-plugins-2.3.3 --htmldir=/usr/share/doc/monitoring-plugins-2.3.3/html --
with-sysroot=/ --libdir=/usr/lib64 --with-libcurl --with-uriparser --with-mysql --with-ipv6 --without-ldap --without-pgsql --without-radius --without-openssl --with-gnutls=/usr --with-ping-command=/
usr/bin/ping -4 -n -U -w %d -c %d %s --with-ping6-command=/usr/bin/ping -6 -n -U -w %d -c %d %s --libexecdir=/usr/lib64/nagios/plugins --sysconfdir=/etc/nagios
...
checking for gnutls/openssl.h... yes
checking for main in -lgnutls-openssl... yes
...
            --with-apt-get-command: 
              --with-ping6-command: /usr/bin/ping -6 -n -U -w %d -c %d %s
               --with-ping-command: /usr/bin/ping -4 -n -U -w %d -c %d %s
                       --with-ipv6: yes
                      --with-mysql: /usr/bin/mysql_config
                    --with-openssl: no
                     --with-gnutls: yes
               --enable-extra-opts: yes
                       --with-perl: /usr/bin/perl
             --enable-perl-modules: no
                     --with-cgiurl: /nagios/cgi-bin
               --with-trusted-path: /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
                   --enable-libtap: no
                    --with-libcurl: yes
                  --with-uriparser: yes
...
/usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../x86_64-pc-linux-gnu/bin/ld/bin/sh ../libtool  --tag=CC   --mode=link x86_64-pc-linux-gnu-gcc -DNP_VERSION='"2.3.3"' -mtune=native -O2 -pipe -fno-stric
[...]

PR #1839, opened 2023-02-06
fallback to SSL_CTX_use_certificate_file for gnutls

waja:
This might closes: #1834

### DIFF
--- a/plugins/sslutils.c
+++ b/plugins/sslutils.c
@@ -134,7 +134,18 @@ int np_net_ssl_init_with_hostname_version_and_cert(int sd, char *host_name, int 
 		return STATE_CRITICAL;
 	}
 	if (cert && privkey) {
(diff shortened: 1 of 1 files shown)


plugins/sslutils.c
```diff
@@ -134,7 +134,18 @@ int np_net_ssl_init_with_hostname_version_and_cert(int sd, char *host_name, int
 		return STATE_CRITICAL;
 	}
 	if (cert && privkey) {
-		SSL_CTX_use_certificate_chain_file(c, cert);
+#ifdef USE_OPENSSL
+		if (!SSL_CTX_use_certificate_chain_file(c, cert)) {
+#else
+#if USE_GNUTLS
```
RincewindsHat, line 140: I think using `#elif` would be easier to read here.

RincewindsHat (changes requested):

RincewindsHat (approved):

PR #1840, opened 2023-02-06
check_http: fix for single zero size chunk only

RincewindsHat:
check_http did not handle the edge case of a single zero length chunk in the content well and segfaults then.

This patch fixes that behaviour and also adds a test case.


sni (approved):
looks good.

RincewindsHat:
I wrote extra tests and "looks good" is all I get?! :-P

sni:
isn't writing tests in perl rewarding enough? :-)

PR #1842, opened 2023-02-08
check_radius: cope with radcli-1.3.1 RC_BUFFER_LEN

sthen:
radcli 1.3.1 now uses RC_BUFFER_LEN instead of BUFFER_LEN. Add an #ifdef to allow working with either.


RincewindsHat:
@sthen  Hm, I could not really find a reference to `RC_BUFFER_LENGTH` in that context, but something different ( https://github.com/radcli/radcli/blob/02b6d672f782b12cb6a8f57ba47f1c7bb75530be/lib/sendserver.c#L226 ). Am I looking at it from the wrong angle?

sthen:
It's RC_BUFFER_LEN, not _LENGTH. https://github.com/radcli/radcli/pull/80/commits/45e7e47b36773d5f2ef13a3d6ea34d6ad9d01244

RincewindsHat:
Ahhh, thanks, my fault.

RincewindsHat (approved):

PR #1843, opened 2023-02-12
Various fixes and improvements

andreasbaumann:
- fixed: a potential regex error buffer overflow
- memory leaks and cleanup of libcurl after die (was simply not called)
- enabled the libcurl cookie engine to store cookies on check side


plugins/check_curl.c
```diff
@@ -160,10 +160,16 @@ char *http_method = NULL;
 char *http_post_data = NULL;
 char *http_content_type = NULL;
 CURL *curl;
+int curl_global_initialized = 0;
```
RincewindsHat, line 163: Could we use `bool` here to, well, tell readers and the compiler that it is really a boolean?

plugins/check_curl.c
```diff
@@ -688,9 +720,13 @@ check_http (void)
       handle_curl_option_return_code (curl_easy_setopt (curl, CURLOPT_MAXREDIRS, max_depth+1), "CURLOPT_MAXREDIRS");
 
       /* for now allow only http and https (we are a http(s) check plugin in the end) */
+#if LIBCURL_VERSION_NUM >= MAKE_LIBCURL_VERSION(7, 85, 0)
+      handle_curl_option_return_code (curl_easy_setopt (curl, CURLOPT_REDIR_PROTOCOLS_STR, "http,https"), "CURLOPT_REDIR_PROTOCOLS_STR");
+#else
```
RincewindsHat, line 725: Would the use of `#elif` make this more readable?

RincewindsHat (changes requested):
Thanky you very much for the changes, but I have some questions/suggestions I would like to address.

plugins/check_curl.c
andreasbaumann: Sure, can change that, it' is logically a bool (so are many more variables in check_curl.c). But check_curl.c was on purpose looking similar to check_http.c, so that backports (if necessary) would be easier.

plugins/check_curl.c
andreasbaumann: Sure thing, will change that. Seems to be a old habit of mine to avoid '#elif' :-)

plugins/check_curl.c
RincewindsHat: I will happily change all the pig booleans to real ones in `check_http` :-)

RincewindsHat: `monitoring-plugins` doesn't compile with anything lower than C99 anyways (which is partially my fault :-) ). That's said and done. Given enough time and motivation I would like to change that everywhere.

andreasbaumann: Ok, then it's no problem, will change.. :-)

RincewindsHat: Great, thanks!

RincewindsHat (approved):

RincewindsHat:
Looks goods too me :-) I don't know much about all those curl options and functions, but it looks like you do.

PR #1846, opened 2023-03-08
check_curl.c: Include all IPs from getaddrinfo() in curl DNS cache

bazzisoft:
@andreasbaumann Based on the discussions on your branch https://github.com/monitoring-plugins/monitoring-plugins/commit/03f86b5d0809967855fbaafb4d600dc5b82081fa, I've created this PR which extends it to include all IPs returned by `getaddrinfo()` in the curl resolver cache. Let me know your thoughts!

Ref issue: https://github.com/monitoring-plugins/monitoring-plugins/issues/1844

### Sample output:

```
$ plugins/check_curl -H iptest.websitenameserver.com -S -v
* getaddrinfo IPv4 address: 143.42.96.219
* getaddrinfo IPv6 address: 2a01:7e00::f03c:93ff:fefe:fbc
* curl CURLOPT_RESOLVE: iptest.websitenameserver.com:443:143.42.96.219,2a01:7e00::f03c:93ff:fefe:fbc
* curl CURLOPT_URL: https://iptest.websitenameserver.com:443/
* Added iptest.websitenameserver.com:443:143.42.96.219,2a01:7e00::f03c:93ff:fefe:fbc to DNS cache
* Hostname iptest.websitenameserver.com was found in DNS cache
*   Trying 143.42.96.219:443...
* Connected to iptest.websitenameserver.com (143.42.96.219) port 443 (#0)
```

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -422,8 +422,8 @@ lookup_host (const char *host, char *buf, size_t buflen)
     // Append all IPs to buf as a comma-separated string
     addrstr_len = strlen(addrstr);
     if (buflen_remaining > addrstr_len + 1) {
-      if (buf[0] != NULL) {
-        strncat(buf, ",", 1);
+      if (buf[0] != '\0') {
+        strncat(buf, ",", buflen_remaining);
         buflen_remaining -= 1;
       }
       strncat(buf, addrstr, buflen_remaining);
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -384,9 +384,12 @@ int
 lookup_host (const char *host, char *buf, size_t buflen)
 {
   struct addrinfo hints, *res, *result;
+  char addrstr[100];
+  size_t addrstr_len;
   int errcode;
   void *ptr;
   int s;
+  size_t buflen_remaining = buflen - 1;
 
   memset (&hints, 0, sizeof (hints));
   hints.ai_family = address_family;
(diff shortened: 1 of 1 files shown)


plugins/check_curl.c
```diff
@@ -453,7 +463,7 @@ check_http (void)
   int i;
   char *force_host_header = NULL;
   struct curl_slist *host = NULL;
-  char addrstr[100];
+  char addrstr[DEFAULT_BUFFER_SIZE/2];
```
andreasbaumann, line 466: I would actually introduce a HOSTNAME_SIZE or so.. The host name should be 64 or so for a machine name and maybe 255 for a FQDN? I'm lacking actual data here..

andreasbaumann (approved):

plugins/check_curl.c
andreasbaumann: Ah. This is the whole list of hostname.. ok then..

bazzisoft: Yeah, this is a bit tricky. We need enough space to store all the IPs in `addrstr`, but also enough extra space in the final buffer passed to curl to store the FQDN + port + all IPs. Halving the default buffer size seemed like a quick way to have enough space for both, without hard-coding numbers. 1024 bytes should be enough for up to 25 IPv6 addresses, which i think is enough? In any case it'll just drop any additional IPs that don't fit without failing.

But, feel free to modify the buffer sizes as you see fit if there are more logical values.

plugins/check_curl.c
```diff
@@ -422,8 +422,8 @@ lookup_host (const char *host, char *buf, size_t buflen)
     // Append all IPs to buf as a comma-separated string
     addrstr_len = strlen(addrstr);
     if (buflen_remaining > addrstr_len + 1) {
-      if (buf[0] != NULL) {
-        strncat(buf, ",", 1);
+      if (buf[0] != '\0') {
+        strncat(buf, ",", buflen_remaining);
```
bazzisoft, line 426: 👍

PR #1847, opened 2023-03-09
Fix for SSL host list messup when picking from multiple IPs (#1844)

andreasbaumann:


plugins/check_curl.c
```diff
@@ -384,8 +384,12 @@ int
 lookup_host (const char *host, char *buf, size_t buflen)
 {
   struct addrinfo hints, *res, *result;
+  char addrstr[100];
+  size_t addrstr_len;
   int errcode;
   void *ptr;
+  int s;
```
RincewindsHat, line 391: `s` is unused?

RincewindsHat (changes requested):

plugins/check_curl.c
andreasbaumann: Yep, removed in ea53555f

RincewindsHat:
Resolves #1844

Issue #1789, opened 2022-09-20
build failure with upcoming clang-16


orlitzky:
clang-15 tried to enable the following by default:

* `-Werror=implicit-function-declaration`
* `-Werror=implicit-int`
* `-Werror=strict-prototypes`

This caused some breakage, so they're delaying it until clang-16. But it's still coming. With those flags, monitoring-plugins fails to build:

```
/usr/include/openssl/macros.h:62:52: note: expanded from macro 'OSSL_DEPRECATED'
#     define OSSL_DEPRECATED(since) __attribute__((deprecated))
                                                   ^
sslutils.c:198:9: error: call to undeclared function
      'np_net_ssl_check_certificate'; ISO C99 and later do not support implicit
      function declarations [-Werror,-Wimplicit-function-declaration]
```

(This was reported to us at https://bugs.gentoo.org/871519 on Gentoo)

RincewindsHat:
Ich changed the title, since the problem occurs with the openssl bindings and `check_disk` does not use those. But this impacts several other plugins.

Thank you for reporting this issue.

RincewindsHat:
The error should be fixed with c4704e163ebf54277ff901f06f09126ef3a3bc7f when it is merged. The OpenSSL API deprecation is probably a topic for another time

RincewindsHat:
The rest should be fixed with #1848

RincewindsHat:
@orlitzky Parts of this should be fixed by #1868
I tried the master with `Debian clang version 16.0.6 (15)` and it compiles (with warnings).
I will close this issue, please reopen it, if you think this is not yet resolved.

orlitzky:
Wow, that wound up being a lot of work. Thank you!

PR #1848, opened 2023-03-12
Compiler warnings

RincewindsHat:
This PR tries to fix some of the compiler warnings for the C based plugins, including the deprecated functions.

### DIFF
--- a/lib/extra_opts.c
+++ b/lib/extra_opts.c
@@ -93,14 +93,14 @@ char **np_extra_opts(int *argc, char **argv, const char *plugin_name){
 			/* append the list to extra_args */
 			if(extra_args==NULL){
 				extra_args=ea1;
-				while(ea1=ea1->next) ea_num++;
+				while((ea1 = ea1->next)) ea_num++;
 			}else{
 				ea_tmp=extra_args;
 				while(ea_tmp->next) {
 					ea_tmp=ea_tmp->next;
 				}
 				ea_tmp->next=ea1;
-				while(ea1=ea1->next) ea_num++;
+				while((ea1=ea1->next)) ea_num++;
 			}
 			ea1=ea_tmp=NULL;
 		}
--- a/lib/maxfd.c
+++ b/lib/maxfd.c
@@ -1,0 +1,26 @@
+#include "./maxfd.h"
+#include <errno.h>
+
+long open_max (void) {
+	long maxfd = 0L;
+	/* Try sysconf(_SC_OPEN_MAX) first, as it can be higher than OPEN_MAX.
+	 * If that fails and the macro isn't defined, we fall back to an educated
+	 * guess. There's no guarantee that our guess is adequate and the program
+	 * will die with SIGSEGV if it isn't and the upper boundary is breached. */
+
+#ifdef _SC_OPEN_MAX
+	errno = 0;
+	if ((maxfd = sysconf (_SC_OPEN_MAX)) < 0) {
+		if (errno == 0)
+			maxfd = DEFAULT_MAXFD;   /* it's indeterminate */
+		else
+			die (STATE_UNKNOWN, _("sysconf error for _SC_OPEN_MAX\n"));
+	}
+#elif defined(OPEN_MAX)
+	return OPEN_MAX
+#else	/* sysconf macro unavailable, so guess (may be wildly inaccurate) */
+	return DEFAULT_MAXFD;
+#endif
+
+	return(maxfd);
+}
--- a/lib/parse_ini.c
+++ b/lib/parse_ini.c
@@ -324,14 +324,16 @@ add_option(FILE *f, np_arg_list **optlst)
 
 	read_pos = 0;
 	optnew->arg = malloc(cfg_len + 1);
+
 	/* 1-character params needs only one dash */
 	if (opt_len == 1) {
-		strncpy(&optnew->arg[read_pos], "-", 1);
+		strncpy(&optnew->arg[read_pos], "-", 2);
 		read_pos += 1;
 	} else {
-		strncpy(&optnew->arg[read_pos], "--", 2);
+		strncpy(&optnew->arg[read_pos], "--", 3);
 		read_pos += 2;
 	}
+
 	strncpy(&optnew->arg[read_pos], optptr, opt_len);
 	read_pos += opt_len;
 	if (value) {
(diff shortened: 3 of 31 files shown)


RincewindsHat:
I fear this needs a review, maybe @sni or @andreasbaumann ?

RincewindsHat:
Fixes #1789

RincewindsHat:
@waja you've got time for this? :-)

waja:
Dunno if I'm skilled enough for this task. :)

lib/extra_opts.c
```diff
@@ -93,14 +93,14 @@ char **np_extra_opts(int *argc, char **argv, const char *plugin_name){
...
 				extra_args=ea1;
-				while(ea1=ea1->next) ea_num++;
+				while((ea1 = ea1->next)) ea_num++;
 			}else{
 				ea_tmp=extra_args;
 				while(ea_tmp->next) {
 					ea_tmp=ea_tmp->next;
 				}
 				ea_tmp->next=ea1;
-				while(ea1=ea1->next) ea_num++;
```
sni, line 103: what's the reason for adding duplicate brackets? And why change line 96 and 103 into different styles?

lib/maxfd.c
```diff
@@ -0,0 +1,26 @@
...
+		else
+			die (STATE_UNKNOWN, _("sysconf error for _SC_OPEN_MAX\n"));
+	}
+#elif defined(OPEN_MAX)
+	return OPEN_MAX
+#else	/* sysconf macro unavailable, so guess (may be wildly inaccurate) */
+	return DEFAULT_MAXFD;
+#endif
+
+	return(maxfd);
```
sni, line 25: so if _SC_OPEN_MAX and OPEN_MAX is undefined, we will return 0? Was it like that before? Why not return DEFAULT_MAXFD in that case?

lib/parse_ini.c
```diff
@@ -324,14 +324,16 @@ add_option(FILE *f, np_arg_list **optlst)
 
 	read_pos = 0;
 	optnew->arg = malloc(cfg_len + 1);
+
 	/* 1-character params needs only one dash */
 	if (opt_len == 1) {
-		strncpy(&optnew->arg[read_pos], "-", 1);
```
sni, line 329: Do we have a test case for this? Such changes make me nervous :-)

plugins-root/check_dhcp.c
```diff
@@ -470,10 +470,6 @@ int send_dhcp_discover(int sock){
 	packet_xid=random();
 	discover_packet.xid=htonl(packet_xid);
 
-	/**** WHAT THE HECK IS UP WITH THIS?!?  IF I DON'T MAKE THIS CALL, ONLY ONE SERVER RESPONSE IS PROCESSED!!!! ****/
-	/* downright bizzarre... */
-	ntohl(discover_packet.xid);
```
sni, line 475: is there a test case for this one or is it no longer the case?

plugins/check_curl.c
```diff
@@ -297,6 +297,7 @@ main (int argc, char **argv)
 
 int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 {
+	(void) preverify_ok;
```
sni, line 300: i would prefer using `__attribute__ ( (unused))` instead.

plugins/check_curl.c
```diff
@@ -1023,23 +1055,60 @@ check_http (void)
 
   if (strlen (header_expect)) {
     if (!strstr (header_buf.buf, header_expect)) {
+
```
sni, line 1058: i would prefer to not start blocks with a blank line.

sni (changes requested):
a PR from hell... :-) Personally i would prefer to separate this into multiple PRs and separate arguable changes from simple changes and real code changes which might break functionality.

RincewindsHat:
If you insist, I can split this up :-) I just lazily worked through all the warning I could find and put them in one thing here

lib/extra_opts.c
RincewindsHat:
```
warning: suggest parentheses around assignment used as truth value [-Wparentheses]
```
Arguably this one is not so important.

As for changing the style, I think it is easier to read with spaces around `=`, I don't know why I skipped it for 103.

lib/maxfd.c
RincewindsHat: if both are not defined, it should return `DEFAULT_MAXFD` from `lib/maxfd.h` with is 256.
I am pretty sure it was like this before.

plugins/check_curl.c
RincewindsHat: that's a gcc extension?

plugins/check_curl.c
RincewindsHat: Indeed, sorry about that.

plugins-root/check_dhcp.c
RincewindsHat: there is not test case to my knowledge and I wouldn't know how to test that. In  theory this does not do anything.
Testing on my machine did not show any changes in behaviour, so I removed it. I mean, it cannot possibly do anything!!! (Or we really need to find out why it does that).
Not sure what problem they had back then, might just have been a timing problem.

dermoth: It comes from [this commit](https://github.com/monitoring-plugins/monitoring-plugins/commit/d50a9917ef38c44ac2067dba508cf2e690ff1412), a port to non-linux platform. At best it's an indication of a bug somewhere, and iirc those type of calls are often implemented as macros and could possibly have side effects. Without knowing what architecture was affected it'll be pretty hard to reproduce anyway :/

dermoth: BTW if you can at least test under different optimization levels (-Os, -O0, -O1, -O2, -O3) - sometimes bugs are hidden until specific optimisations are applied (or not applied). You can also run it with valgrind while testing as it may help catch some hidden issues too.

dermoth: Actually GH followed the move from `plugins/` to `plugins-root/`, but not the earlier move form `contrib/`, so it's [this commit](https://github.com/monitoring-plugins/monitoring-plugins/commit/44a321cb8a42d6c0ea2d96a1086a17f2134c89cc#diff-02bebc64297416b78f09d437e98879fccab6216970fcc30ec32cd86a0584552c).

It's the initial import by Ethan and check_dhcp was in `contrib/` even though it bears Ethan's copyright notice without any other authorship information. So TL;DR it's been there like forever...

RincewindsHat: Someone suggested to me, that this thing might have been a compiler bug. Probably the compiler optimized that part out of the code and broke something with it.
That line tells the compiler, that this is actually something needed.

Issue #1849, opened 2023-03-13
check_snmp: unit removed from check result


jdumalaonITRS:
Hi, 
I am failing at some tests:
```
#   Failed test 'Skipping first two thresholds, result printed rather than parsed'
#   at ./t/check_snmp.t line 137.
#                   'SNMP OK - 1828140 174 | HOST-RESOURCES-MIB::hrMemorySize.0=1828140;; HOST-RESOURCES-MIB::hrSystemProcesses.0=174;; '
#     doesn't match '/^SNMP OK - \d+ \w+ \d+\s.*$/'
#   Failed test 'Skipping all thresholds, result printed rather than parsed'
#   at ./t/check_snmp.t line 141.
#                   'SNMP OK - 1828140 174 | HOST-RESOURCES-MIB::hrMemorySize.0=1828140;; HOST-RESOURCES-MIB::hrSystemProcesses.0=174;; '
#     doesn't match '/^SNMP OK - \d+ \w+ \d+\s.*$/'
```
Previously this returns: 
```
 ./check_snmp -H 192.168.102.120 -C public -o host.hrStorage.hrMemorySize.0 -w ,,1 -c ,,2
SNMP OK - 1828140 KBytes | HOST-RESOURCES-MIB::hrMemorySize.0=1828140;;
```
It seems that the current multiply function removes the unit part in snmpget's  return due to the new multiply function which only returns the number part. 

https://github.com/monitoring-plugins/monitoring-plugins/blob/40dc41225a3ce86bd6f1f446920cd5a6c32b8443/plugins/check_snmp.c#L1166

What's the expected behavior in this case?
Should the test be updated or the unit be put back?

sni:
Seems like two things are going wrong here. If there is no multiplier set, it should behave like before. Also i'd say the unit should be added if possible.

jdumalaonITRS:
Hmm this seems to be a tricky one. There is a possibility of "overrun" since multiply() overwrites snmpgets command output string. e.g.
```
[root@localhost plugins]#  ./check_snmp -H localhost -C public -o host.hrSystem.hrSystemProcesses.0,host.hrStorage.hrMemorySize.0 -w ,,1 -c ,,2 -vvv -M .7 -f %.20f
/usr/bin/snmpget -Le -t 10 -r 5 -m ALL -v 1 [context] [authpriv] localhost:161 host.hrSystem.hrSystemProcesses.0 host.hrStorage.hrMemorySize.0
[...]

PR #1850, opened 2023-03-14
check_snmp: disable multiplier when unused

sni:
- if no multiplier is set, simply return the given string. Otherwise we would strip off the unit.
 - if used, allocate new space to hold the result which might be larger than the initial input

### DIFF
--- a/plugins/check_snmp.c
+++ b/plugins/check_snmp.c
@@ -1179,10 +1179,7 @@ multiply (char *str)
 
 	val = strtod (str, &endptr);
 	if ((val == 0.0) && (endptr == str)) {
-		if(multiplier != 1) {
-			die(STATE_UNKNOWN, _("multiplier set (%.1f), but input is not a number: %s"), multiplier, str);
-		}
-		return str;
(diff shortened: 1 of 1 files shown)


plugins/check_snmp.c
```diff
@@ -1186,16 +1190,17 @@ multiply (char *str)
 	if (fmtstr != "") {
 		conv = fmtstr;
 	}
+	buf = malloc(100);
```
jdumalaonITRS, line 1193: How about we just have this as a static buffer so we do not deal with the malloc? We would have 100 byte leaks for every line in snmpgets output.

sni: sounds good, will do

jdumalaonITRS (approved):
Looks good. Thanks!

RincewindsHat:
close #1849

RincewindsHat (changes requested):
Aparts from that little not, looks good I guess

RincewindsHat (approved):

waja:
Hmm ... might it be a good idea to ship a test for this?

PR #1855, opened 2023-03-16
check_curl: including netinet/in.h (for FreeBSD), fixed an ambigous compare warning

andreasbaumann:
fixed en passant.

On FreeBSD we need:

```
pkg install uriparser curl
./configure --with-libcurl=/usr/local --with-uriparser=/usr/local
```


RincewindsHat:
Looks good to me :-) Thank you

RincewindsHat (approved):

PR #1858, opened 2023-04-03
check_snmp: Fix the detection of the presence of perfdata types

RincewindsHat:
This fix originates from OpenSuse, can't find a definitive author, but might be Lars Vogdt ( @lrupp )


RincewindsHat:
@sni since ware already touching `check_snmp`, what about this one?

RincewindsHat:
obsolete now

PR #1863, opened 2023-04-13
Fix 'requres' typo

Donien:
There is just that one obvious typo that bothered me.  
Here's a fix I hope (I `grep`ed for every occurrence).


RincewindsHat (approved):

RincewindsHat:
Thanks!

PR #1865, opened 2023-04-14
CI: Adding Codespell (and Super Linter)

waja:


RincewindsHat:
@waja do we really want to run that with every push/pr/... ? Seems like wasting ressources a little bit.

waja:
Running it on every PR ensures that not slipping low hanging spelling errors into the code and it can be fixed in the PR itself.
Fixing spelling errors afterwards (in a bunch) makes backporting functional fixes very hard (which might be a thing for downstream maintainers, like me).

sni (approved):

sni:
fine for me, maybe we should fix some of the spelling errors before instead of putting them on the ignore list, ex.: `psuedo`

waja:
When you are looking into the code, `psuedo` is not likely a spelling error in this case. :)

sni:
Really, I assumed it is... I cannot find any meaning for this word

waja:
```
✦ ❯ rgrep psuedo .
./plugins/check_radius.c:The (psuedo?)literate programming XML is contained within \@\@\- <XML> \-\@\@
./plugins/check_pgsql.c:The (psuedo?)literate programming XML is contained within \@\@\- <XML> \-\@\@
./.github/workflows/test.yml:          ignore_words_list: allright,psuedo,gord,didi,hda,nd,alis,clen,scrit,ser,fot,te,parm,isnt,consol,oneliners
```

https://www.google.de/search?q=psuedo+xml

@sni would you bet? ;)

sni:
Well, google says: did you mean "pseudo"? I am not convinced yet, but i try to avoid xml whenever possible, so no idea if that's a real thing or just a common typo:
https://en.wiktionary.org/wiki/psuedo-

sni:
> @sni would you bet? ;)

Sure, i am in. Next beer is on me if you are right.

waja:
What a shit show. Let's change the XML text to "pseudo".

sni:
do i get a beer now? :beers:

waja:
Let's say two. But the real challenge is: where and when to meet in person.

sni:
we'll figure something out :-)

waja:
Anything else you spotted we should remove from the filter?

sni (approved):

sni:
fine for me :)

PR #1867, opened 2023-04-17
Compiler warning part 2

RincewindsHat:
Parts of #1848
Easier to digest :-)

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -244,7 +244,7 @@ void curlhelp_freewritebuffer (curlhelp_write_curlbuf*);
 int curlhelp_initreadbuffer (curlhelp_read_curlbuf *, const char *, size_t);
 int curlhelp_buffer_read_callback (void *, size_t , size_t , void *);
 void curlhelp_freereadbuffer (curlhelp_read_curlbuf *);
-curlhelp_ssl_library curlhelp_get_ssl_library (CURL*);
+curlhelp_ssl_library curlhelp_get_ssl_library ();
 const char* curlhelp_get_ssl_library_string (curlhelp_ssl_library);
 int net_noopenssl_check_certificate (cert_ptr_union*, int, int);
 
@@ -297,6 +297,7 @@ main (int argc, char **argv)
 
 int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 {
+	(void) preverify_ok;
   /* TODO: we get all certificates of the chain, so which ones
    * should we test?
    * TODO: is the last certificate always the server certificate?
@@ -321,6 +322,8 @@ int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 
 CURLcode sslctxfun(CURL *curl, SSL_CTX *sslctx, void *parm)
 {
+	(void) curl; // ignore unused parameter
+	(void) parm; // ignore unused parameter
   SSL_CTX_set_verify(sslctx, SSL_VERIFY_PEER, verify_callback);
 
   return CURLE_OK;
@@ -646,7 +649,7 @@ check_http (void)
   }
 
   /* detect SSL library used by libcurl */
-  ssl_library = curlhelp_get_ssl_library (curl);
+  ssl_library = curlhelp_get_ssl_library ();
 
   /* try hard to get a stack of certificates to verify against */
   if (check_cert) {
@@ -2381,7 +2384,7 @@ get_content_length (const curlhelp_write_curlbuf* header_buf, const curlhelp_wri
 
 /* TODO: is there a better way in libcurl to check for the SSL library? */
 curlhelp_ssl_library
-curlhelp_get_ssl_library (CURL* curl)
+curlhelp_get_ssl_library ()
 {
   curl_version_info_data* version_data;
   char *ssl_version;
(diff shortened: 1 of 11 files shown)


plugins/check_curl.c
```diff
@@ -1654,6 +1654,7 @@ process_arguments (int argc, char **argv)
       break;
     case 'R': /* regex */
       cflags |= REG_ICASE;
+			// fall through
```
andreasbaumann, line 1747: Making compilers happy with comments, ok. :-)
This is a fall-through here..

plugins/check_dbi.c
```diff
@@ -426,6 +426,7 @@ process_arguments (int argc, char **argv)
 			else
 				timeout_interval = atoi (optarg);
 
+			break;
```
andreasbaumann, line 429: This looks really like a break to me, as 't' (timeout) and 'H' (host) probably have nothing in common to be parsed. ok.
Actually, good catch..

plugins/check_http.c
```diff
@@ -498,6 +498,7 @@ bool process_arguments (int argc, char **argv)
       break;
     case 'R': /* regex */
       cflags |= REG_ICASE;
+			// fall through
```
andreasbaumann, line 501: same as in check_http, is a copy-paste of me originally. ok.

andreasbaumann (approved):
Reviewed the once, I can judge.

plugins/check_curl.c
```diff
@@ -244,7 +244,7 @@ void curlhelp_freewritebuffer (curlhelp_write_curlbuf*);
 int curlhelp_initreadbuffer (curlhelp_read_curlbuf *, const char *, size_t);
 int curlhelp_buffer_read_callback (void *, size_t , size_t , void *);
 void curlhelp_freereadbuffer (curlhelp_read_curlbuf *);
-curlhelp_ssl_library curlhelp_get_ssl_library (CURL*);
```
andreasbaumann, line 247: ok, CURL* no longer used inside the function, or never used.

plugins/check_curl.c
```diff
@@ -297,6 +297,7 @@ main (int argc, char **argv)
 
 int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 {
+	(void) preverify_ok;
```
andreasbaumann, line 300: Could also be marked unused with something like:

_GL_ATTRIBUTE_UNUSED __attribute__ ((__unused__))

(but that one is from gnulib, so not sure, we should use that one).

The (void)PARAM approach works too.

plugins/check_curl.c
```diff
@@ -321,6 +322,8 @@ int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 
 CURLcode sslctxfun(CURL *curl, SSL_CTX *sslctx, void *parm)
 {
+	(void) curl; // ignore unused parameter
```
andreasbaumann, line 325: see verify_callback

plugins/check_curl.c
```diff
@@ -2381,7 +2384,7 @@ get_content_length (const curlhelp_write_curlbuf* header_buf, const curlhelp_wri
 
 /* TODO: is there a better way in libcurl to check for the SSL library? */
 curlhelp_ssl_library
-curlhelp_get_ssl_library (CURL* curl)
+curlhelp_get_ssl_library ()
```
andreasbaumann, line 2504: ok

andreasbaumann (commented):
commented on check_curl things

plugins/check_curl.c
```diff
@@ -378,8 +378,12 @@ void
 handle_curl_option_return_code (CURLcode res, const char* option)
 {
   if (res != CURLE_OK) {
-    snprintf (msg, DEFAULT_BUFFER_SIZE, _("Error while setting cURL option '%s': cURL returned %d - %s"),
```
andreasbaumann, line 378: This is just argument rearangements, I cannot find a reason, why this should have generated a compiler
warning before (if so, which compiiler warning is it fixing for clang and for which flag?).

This is a coding style issue: do we prefer compacter code or split arguments per line?

andreasbaumann (commented):
Not 100% sure about all the changes, the beautifications I don't mind/care, but why various
and different buffer allocations for message buffers?

plugins/check_curl.c
RincewindsHat: this is most like not a compiler warning, but a styling thing I did to actually be able to read it

plugins/check_curl.c
```diff
@@ -1026,23 +1055,60 @@ check_http (void)
...
     if (!strstr (header_buf.buf, header_expect)) {
+
       strncpy(&output_header_search[0],header_expect,sizeof(output_header_search));
+
       if(output_header_search[sizeof(output_header_search)-1]!='\0') {
         bcopy("...",&output_header_search[sizeof(output_header_search)-4],4);
       }
-      snprintf (msg, DEFAULT_BUFFER_SIZE, _("%sheader '%s' not found on '%s://%s:%d%s', "), msg, output_header_search, use_ssl ? "https" : "http", host_name ? host_name : server_address, server_port, server_url);
-      result = STATE_CRITICAL;
+
```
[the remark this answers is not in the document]
RincewindsHat: Because the compiler was unhappy about using the same buffer to read AND write (which could result in overlap and weird results).
Allocating a buffer on the heap is probably the wrong approch. Using the stack might be better.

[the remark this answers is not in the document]
RincewindsHat: Fixed it

RincewindsHat:
@sni what do you think about the rest of those changes? The ones @andreasbaumann did not judge?

PR #1868, opened 2023-04-17
Replace deprecated TLS client functions

RincewindsHat:
Parts of #1848
Easier to digest :-)

Only replacement for deprecated OpenSSL methods.


RincewindsHat:
This also needs a (short) review. @sni @andreasbaumann

sni (approved):
looks reasonable

PR #1869, opened 2023-04-17
Compiler warning part 4

RincewindsHat:
Parts of #1848
Easier to digest :-)


plugins/check_curl.c
```diff
@@ -1133,6 +1133,11 @@ redir (curlhelp_write_curlbuf* header_buf)
     &status_line.http_major, &status_line.http_minor, &status_line.http_code, &status_line.msg, &msglen,
     headers, &nof_headers, 0);
 
+	if (res == -1) {
```
andreasbaumann, line 1140: Why only here? There is also one in 2304, 2362?
phr_parse_response is not returning -1 on error. What should be done on -2 (partial response)?

andreasbaumann (commented):
This needs a closer look IMHO..

andreasbaumann (commented):
Do I see this correctly, this is also a refactoring taking out the max_fd function for future reuse in other places (which is a good thing)?

RincewindsHat:
Implicitely yes, because it was previously included in places where it was not used.

plugins/check_curl.c
RincewindsHat: I don't know :-| What would you suggest? Die in all non-OK cases?

RincewindsHat:
Cold and outdated

PR #1884, opened 2023-05-26
check_disk: Display SI units correctly

### LINKED ISSUE #1786: check_disk support new units like MiB
New unit names Mib, KiB, GiB, TiB should be added to be able to solve problems with systems like icinga and grafana which redefined the old names (MB, GB ...) but understand the new names. See: https://github.com/nagios-plugins/nagios-plugins/blob/master/plugins/check_disk.c

RincewindsHat:
Allow users to choose SI units as the units of choice in `check_disk`

fixes #1786


sni:
i guess this is only used in the text part of the plugin output, so i don't think it will break things.

PR #1888, opened 2023-06-16
check_users: prefer systemd-logind over utmp

thkukuk:
Prefer systemd-logind over utmp to get the number of logged in users. utmp is not reliable for this (e.g. some terminals create utmp entries, other not) and utmp is not Y2038 safe with glibc on Linux.


plugins/check_users.c
```diff
@@ -115,6 +120,11 @@ main (int argc, char **argv)
 
 	WTSFreeMemory(wtsinfo);
 #elif HAVE_UTMPX_H
+#ifdef HAVE_LIBSYSTEMD
```
RincewindsHat, line 123: is systemd somehow dependent on `UTMPX`?

thkukuk: It does not depend on UTMP, but all other projects preferred to have utmp support as fallback for the case, that the application is linked against libsystemd, but systemd-logind is not running.
I can change it, that, if compiled against libsystemd, utmp will be always ignored, if wanted.

thkukuk:
The failing test: `#   Failed test 'String matches invalid host'` seems to come from another change or test host problem, that code was not touched by this PR.

RincewindsHat:
Jup, seems to be an change in debian testing. We are on it.

plugins/check_users.c
thkukuk: @RincewindsHat should I change the PR to use logind only if compiled against libsystemd and no fallback to utmp, or should I let it as it is and resolve the conversation?

thesamesam: I'd prefer it if applications did the same thing here - just fallback.

RincewindsHat: So sorry for the delay, totally on the side of having a fallback, just wondering why this is in the branch of `HAVE_UTMPX_H`.

thkukuk: I changed that now.

RincewindsHat:
Thanks! Could you run the translation thing again though (`cd po && make update-po`)?

Also, please remove the change to the workflow description, it would be offtopic.

thkukuk:
> Also, please remove the change to the workflow description, it would be offtopic.

Ok, I removed your commit which made this change.

RincewindsHat (approved):
Thank you very much

PR #1889, opened 2023-06-18
Add support for SMTP over TLS

franzs:
This PR adds new flags `-s/--ssl/--tls` to use SMTP over TLS (smtps). Per default port 465 is used if one of these new flags is set.

By changing the order of code -r/--proxy can be used with -s/--ssl/--tls, too.

These combination of options were tested successfully:

```
./plugins/check_smtp -H backend-smtp.example.com --tls --proxy
./plugins/check_smtp -H backend-smtp.example.com --tls --proxy -p 587  (fails as expected)
./plugins/check_smtp -H backend-smtp.example.com --starttls -p 587 --proxy
./plugins/check_smtp -H frontend-smtp.example.com --tls
./plugins/check_smtp -H frontend-smtp.example.com --tls -p 587 (fails as expected)
./plugins/check_smtp -H frontend-smtp.example.com --starttls -p 587
```

### DIFF
--- a/plugins/check_smtp.c
+++ b/plugins/check_smtp.c
@@ -505,6 +505,7 @@ process_arguments (int argc, char **argv)
 		{"help", no_argument, 0, 'h'},
 		{"lmtp", no_argument, 0, 'L'},
 		{"ssl", no_argument, 0, 's'},
+		{"tls", no_argument, 0, 's'},
 		{"starttls",no_argument,0,'S'},
 		{"sni", no_argument, 0, SNI_OPTION},
 		{"certificate",required_argument,0,'D'},
@@ -715,7 +716,7 @@ process_arguments (int argc, char **argv)
 		from_arg = strdup(" ");
 
 	if (use_starttls && use_ssl) {
-		usage4 (_("Set either -s/--ssl or -S/--starttls"));
+		usage4 (_("Set either -s/--ssl/--tls or -S/--starttls"));
 	}
 
 	return validate_arguments ();
@@ -875,7 +876,7 @@ print_help (void)
 #ifdef HAVE_SSL
   printf (" %s\n", "-D, --certificate=INTEGER[,INTEGER]");
   printf ("    %s\n", _("Minimum number of days a certificate has to be valid."));
-  printf (" %s\n", "-s, --ssl");
+  printf (" %s\n", "-s, --ssl, --tls");
   printf ("    %s\n", _("Use SSL/TLS for the connection."));
   printf (_("    Sets default port to %d.\n"), SMTPS_PORT);
   printf (" %s\n", "-S, --starttls");
--- a/plugins/check_smtp.c
+++ b/plugins/check_smtp.c
@@ -50,7 +50,8 @@ int days_till_exp_warn, days_till_exp_crit;
 #endif
 
 enum {
-	SMTP_PORT	= 25
+	SMTP_PORT	= 25,
+	SMTPS_PORT	= 465
 };
 #define PROXY_PREFIX "PROXY TCP4 0.0.0.0 0.0.0.0 25 25\r\n"
 #define SMTP_EXPECT "220"
@@ -650,6 +651,7 @@ process_arguments (int argc, char **argv)
 		case 's':
 		/* ssl */
 			use_ssl = TRUE;
+			server_port = SMTPS_PORT;
 			break;
 		case 'S':
 		/* starttls */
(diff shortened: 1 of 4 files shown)


RincewindsHat:
Putting it in an etra PR wouldn't make any difference for Backporting wouldn't it? (As long as we don't squash it here)

franzs:
@waja you are right: in this PR there are unrelated changes for the pot/po files. this is a result of not upating these files in previous PR. so these missing updates occur here.

the general question is: where should these files being updated? IMHO there are three options:

1. within a PR where new localizable strings are added
2. during a release process
3. no updates at all

to me it seems option 3 was used up to now... 😉 so the comments pointing to the origin in the source code were way out of sync.

i tried to go with option 1. given that the po/pot files are in sync before starting to implement a new feature or bugfix, `make update-po` would only reflect the changes introduced with a PR. so only related changes would be in the PR.

Of course the monitoring-plugins project should decide which option it prefers. i tried to be a good contributor and take care of aspects besides the actual code changes. 😇

waja:
@franzs how did you run `make update-po` after checking out the lastest git version?

waja:
Hi @franzs,

thanks, but I can't reproduce this. :/

```
waja@unstable:~/upstreams/monitoring-plugins (master)$ make update-po
make: *** No rule to make target 'update-po'.  Stop.
waja@unstable:~/upstreams/monitoring-plugins (master)$ ls -la Makefile*
-rw-r--r-- 1 waja waja 2103 Jan 22  2022 Makefile.am
```

franzs:
@waja, please change to the `po` directory first:

```bash
% cd po
% make update-po
make monitoring-plugins.pot-update
make[1]: Entering directory '/home/franz/git/monitoring-plugins/monitoring-plugins/po'
[...]
```

There is no target for this in the top level makefile.

waja:
Okay ... so I guess the reason why most people don't update the `po/` is, that indeed nobody is running `tools/setup` on the git tree, cause it creates files which are not intended to be in the tree and ppl have to care about.
For myself I also would prevent to take this extra work.

Anybody an idea how to lower the bar and making it easier to contribute to the `po/`?

RincewindsHat:
I would like to continue the translation thing in #1891, if that's ok. @franzs  Sorry for the delay on the rest, I will take a look, when I find the time and energy for a proper review.

franzs:
@RincewindsHat it's a good idea to move this discussion to a separate place. thanks! i'll jump in there... :nerd_face:

what about removing the two commits for updating the pot/po files here? this way we could move forward independently of the outcome of the discussion in https://github.com/monitoring-plugins/monitoring-plugins/issues/1891. i think the files can be synced later or maybe be removed in another PR.

waja:
> what about removing the two commits for updating the pot/po files here?

Don't care about it in this branch ... if I'm porting this fix back, I will drop the corresponding commits (thanks god you did a separate one!).

waja:
> And you have to run `tools/setup` first to create the makefiles.

Indeed I needed to run:

```
tools/setup && ./configure && cd po && make update-po
```

`./configure` created the `Makefile`s.

RincewindsHat:
I introduced a merge conflict, sorry!

RincewindsHat:
hopefully fixed the merge conflict

plugins/check_smtp.c
```diff
@@ -632,10 +649,14 @@ process_arguments (int argc, char **argv)
 #else
 			usage (_("SSL support not available - install OpenSSL and recompile"));
 #endif
-			// fall through
+		case 's':
+		/* ssl */
+			use_ssl = TRUE;
+			server_port = SMTPS_PORT;
```
RincewindsHat, line 655: this will overwrite a previous `--port` option.

RincewindsHat (changes requested):
Ok, finally got around to think through this. Looks good too me, except for the little problem I noted there.

plugins/check_smtp.c
franzs: Good catch, @RincewindsHat . Thanks! I added another commit to move the port option more to the bottom. Thus `-p` takes precedence over the default port with `-s`/`--ssl`.

RincewindsHat: Thanks @franzs, but this does not change the behaviour, since the `getopt` loop executes for every option, the position in the `switch` `case` does not matter at all (except for fall througs).
Could you undo that last change and perhaps use some local variable (let's say `bool port_explicitely_set`) to determine whether the port was set explicitely via the option?

franzs: :man_facepalming:
Yes, of course. It was too late yesterday... :innocent:
The order of command line options matters and not the order of the case blocks.
Thanks for pointing out, @RincewindsHat
I changed this commit. Please have a look at it.

RincewindsHat (approved):

RincewindsHat:
Thank you very much @franzs

franzs:
Thanks for your efforts and your patience. I'll create some additional PR based on this one. Stay tuned!

PR #1892, opened 2023-06-28
Fix CI workflow syntax

waja:


RincewindsHat (approved):
if that works better :-)

PR #1894, opened 2023-06-28
CI specfile

waja:
This PR adds two things:

* A spec file `.github/monitoring-plugins.spec`
* A build matrix running a rpm build based on the spec file on several rpm distributions

Maybe we want shrink the distributions, but I'd like to run it again at least against one bleeding edge distro and a couple of matured once. Maybe anybody want to add a suse flavor?


RincewindsHat:
Sorry for the wait.
You put docker in your docker? Is that healthy? XD

In principle I am fine with that, it would be pretty cool to have this stuff available. Now I am only hesitant about the ressource usage.
Maybe we should just build the packages for master (and make them downloadable) to reduce the waste?
Also, how much free runtime do we get on github?

waja:
> Sorry for the wait. You put docker in your docker? Is that healthy? XD

Did I? It's just the same kind of dockerization like we to for running our usually tests.

> In principle I am fine with that, it would be pretty cool to have this stuff available. Now I am only hesitant about the ressource usage. Maybe we should just build the packages for master (and make them downloadable) to reduce the waste? Also, how much free runtime do we get on github?

It's arguable how much distributions we like to put into the matrix. While resource usage is one side, I'd like to have a valuable test coverage. So maybe running against Alma Linux 9 and RockyLinux 8 (or visa versa) and one Fedora (not sure if latest or rawhide yet).
On the other side, do we wonna check, if recent changes also running fine on rpm based distros? Maybe we run just one rpm based (beside Debian testing)? (Is there an easy way to run our tests when building the rpm from the specfile?).

The full matrix of distributions can be just done on changes in `master` I think (if we think thats usefull).

RincewindsHat:
I think we should do that, but I would reduce the number of distributions/versions. Rocky and Alma are supposedly the same (not sure if this is still the current state). So `8` and `9` and Fedora (don't know the release scheme there) would be enough IMHO.

Additionaly for releases (and possibly the master) uploading the RPMs as downloadable assets would be helpful I guess.

PR #1895, opened 2023-06-29
Try to detect if there are translation updates via CI

waja:
I've prepared a draft to detect missing translation updates in a way. I don't know if that's the golden way ... but it's a starting point.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,6 +74,10 @@ jobs:
               /bin/sh -c '${{ matrix.prepare }} && \
                 tools/setup && \
                 ./configure --enable-libtap  --with-ipv6=no && \
+                set -x && \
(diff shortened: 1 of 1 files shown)


RincewindsHat:
hm, I do like the idea, but can we rely on the ordering of translation messages in the `.po` files?
I mean, if we generate the translation files two times in a row, do we get the same files? Otherwise this will generate a lot(!) of false positives

waja:
> hm, I do like the idea, but can we rely on the ordering of translation messages in the `.po` files?

I did test the pipeline with an updated `po/` and it worked flawless ... yes, this is no safe proof, but we can throw it away again, when it will be a hassle.

waja:
Okay ... do we wonna merge it?

RincewindsHat (approved):
Let's try this

PR #1898, opened 2023-07-18
check_http: New parameter --verify-host will check if -H hostname matches the S…

chanke-mpcdf:
…SL certificate

Backported from nagios-plugins:
nagios-plugins/nagios-plugins@9ceee1ba404cd5f320cf43fa2c37542903bc8e6c

### DIFF
--- a/plugins/check_http.c
+++ b/plugins/check_http.c
@@ -152,6 +152,8 @@ void print_help (void);
 void print_usage (void);
 char *unchunk_content(const char *content);
 
+extern int check_hostname;
+
 int
 main (int argc, char **argv)
 {
@@ -209,6 +211,7 @@ bool process_arguments (int argc, char **argv)
   enum {
     INVERT_REGEX = CHAR_MAX + 1,
     SNI_OPTION,
+    VERIFY_HOST,
     MAX_REDIRS_OPTION,
     CONTINUE_AFTER_CHECK_CERT
   };
@@ -220,6 +223,7 @@ bool process_arguments (int argc, char **argv)
     {"nohtml", no_argument, 0, 'n'},
     {"ssl", optional_argument, 0, 'S'},
     {"sni", no_argument, 0, SNI_OPTION},
+		{"verify-host", no_argument, 0, VERIFY_HOST},
     {"post", required_argument, 0, 'P'},
     {"method", required_argument, 0, 'j'},
     {"IP-address", required_argument, 0, 'I'},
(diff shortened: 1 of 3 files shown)


plugins/check_http.c
```diff
@@ -152,6 +152,8 @@ void print_help (void);
 void print_usage (void);
 char *unchunk_content(const char *content);
 
+extern int check_hostname;
```
RincewindsHat, line 155: Sorry, but I think using external variables is making the code overly complex and unreadable.

Also this is a boolean and therefore should be `bool` ( or `_Bool`)

plugins/check_http.c
```diff
@@ -220,6 +223,7 @@ bool process_arguments (int argc, char **argv)
     {"nohtml", no_argument, 0, 'n'},
     {"ssl", optional_argument, 0, 'S'},
     {"sni", no_argument, 0, SNI_OPTION},
+		{"verify-host", no_argument, 0, VERIFY_HOST},
```
RincewindsHat, line 226: Please use the same kind of indentation as in the rest of the file.

plugins/check_http.c
```diff
@@ -1748,6 +1755,10 @@ print_help (void)
   printf ("    %s\n", _("1.2 = TLSv1.2). With a '+' suffix, newer versions are also accepted."));
   printf (" %s\n", "--sni");
   printf ("    %s\n", _("Enable SSL/TLS hostname extension support (SNI)"));
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L
```
RincewindsHat, line 1758: Why this exact `OPENSSL_VERSION_NUMBER`? Please add a comment at least which explains why this matters.

plugins/sslutils.c
```diff
@@ -35,6 +35,8 @@ static SSL_CTX *c=NULL;
 static SSL *s=NULL;
 static int initialized=0;
 
+int check_hostname = 0;
```
RincewindsHat, line 38: I would prefer not to have more global variables

RincewindsHat (changes requested):
I am verrryyy hesistant to introduce this with more global variables to be honest, the code is confuse enough IMHO

plugins/sslutils.c
chanke-mpcdf: Fair enough, I just copied it from nagios-plugins.
I see 3 possibilities:
1: add check_hostname as argument to np_net_ssl_init_with_hostname_version_and_cert.
  As far as I can see, only check_http is using that function
2: create an new function in sslutils.c
3: make np_net_ssl_init_with_hostname_version_and_cert variadic

Which one do you prefer ? I have a slight preference to 1, since also other checks might be interested in that check.

PR #1901, opened 2023-08-08
check_http: Fixing unchunking body when no body is present

trickert76:
When no-body is enabled and the server responses with a chunked body, there is always an unknown response from check_http because it tries to read the body, which is not present.


RincewindsHat (approved):
Looks good to me, thank you very much.

paladox:
Is there any chance this could be backported to https://packages.debian.org/bookworm/monitoring-plugins?

RincewindsHat:
Hi @paladox, maybe waja is interested to do that. Otherwise just use `check_curl` which does the same as `check_http` but better.

paladox:
Thanks! Switched to check_curl and seems to fix it for me.

waja:
There was opened a bugreport in the Debian BTS: https://bugs.debian.org/1061956

RincewindsHat:
@waja Backport #1901 maybe?

waja:
This was requested for approval: https://bugs.debian.org/1062063

waja:
Should be included into the next point release (this weekend).

teutat3s:
When would be the next scheduled release? We are really looking forward for the integration of this fix.

RincewindsHat:
we don't have scheduled releases, but if @waja  feels like it, we could do one :-)

PR #1907, opened 2023-08-27
Implement option for check_snmp to ignore mib parsing errors (to fix test pipeline)

RincewindsHat:


RincewindsHat:
@sni fixing the test is actually done in the last commit by changing the test.
Not quite sure about the other changes.
I would leave them there as "can't hurt", but what's your opinion on that?

sni:
what other changes do you mean? The index++ changes? I am fine with this PR. Except i would remove the commented code. We have git now, no need keep dead code :-)

RincewindsHat:
@sni done

sni (approved):

sni:
On the long run it might be better to add one general option to pass through arguments to snmp instead of adding a new argument for each thing we want to set for snmp.
Besides that, i am fine :-)

PR #1908, opened 2023-08-30
Add tests for connecting with TLS in check_smtp.t

franzs:
This is a addition to PR #1889

There is an explicit checking of the order of parameters with `-ssl` and `-p`. See https://github.com/monitoring-plugins/monitoring-plugins/pull/1889#pullrequestreview-1597055504


RincewindsHat (approved):
Looks good to me. Thank you very much.

PR #1909, opened 2023-08-30
Fix certificate checks in check smtp

franzs:


RincewindsHat:
@franzs  Would you mind explaining the changes a little bit?

franzs:
@RincewindsHat I updated the description of the PR. Hope this helps a bit. :smile:

RincewindsHat:
@franzs: Are you happy with that or should I wait for more changes? :-P

franzs:
:thinking: Let's see... Maybe I could... :wink: 

No, we are good to go, now. :relieved:

plugins/check_smtp.c
```diff
@@ -288,16 +288,17 @@ main (int argc, char **argv)
 			printf("%s", buffer);
 		}
 
-#  ifdef USE_OPENSSL
-		  if ( check_cert ) {
+		}
+#endif
```
RincewindsHat, line 292: Please add `/* HAVE_SSL */` here, so it is a little bit easier to see what ends here

plugins/check_smtp.c
```diff
@@ -650,6 +651,7 @@ process_arguments (int argc, char **argv)
 #else
 			usage (_("SSL support not available - install OpenSSL and recompile"));
 #endif
+			break;
```
RincewindsHat, line 654: Since there now two possibilities to use TLS, the `break` is correct to prevent the usage of direct TLS in every case.
However I see a slight problem here: This changes the API. If a user did not specify STARTTLS previously, but choose the certificate check option, it would initiate STARTTLS regardless.
Now the `-D` flag is silently ignored.
While this behaviour arguably "better" it will silently change existing setups.

This change actually happened in the previous PR and I did not see it. Mea culpa.

No I see two options here:
1) Restore the previous behaviour, aumatically enable STARTTLS when `-D` is used, which I do not like very much, but would preserve backward compatibility.
2) Let the plugin fail with `UNKNOWN` and inform the user, that a TLS option has to be chosen.

@sni @waja opinions?

plugins/t/check_smtp.t
```diff
@@ -55,6 +57,14 @@ SKIP: {
 	# SSL connection for STARTTLS
 	$res = NPTest->testCmd( "./check_smtp -H $host_tcp_smtp_starttls -p 25 -S" );
 	is ($res->return_code, 0, "OK, with STARTTLS" );
+
+	$res = NPTest->testCmd( "./check_smtp -H $host_tcp_smtp_starttls -p 25 -S -D 14" );
```
RincewindsHat, line 61: Could you use `--starttls` here instead of `-S`. Would be more obvious in my eyes

plugins/t/check_smtp.t
```diff
@@ -55,6 +57,14 @@ SKIP: {
...
+	is ($res->return_code, 0, "Check rc of successful certificate check with STARTTLS");
+	like ($res->output, qr/^OK - Certificate '$host_tls_cert' will expire on /, "Check output of successful certificate check with STARTTLS");
+
+	$res = NPTest->testCmd( "./check_smtp -H $host_tcp_smtp_starttls -p 25 -S -D 3650" );
```
RincewindsHat, line 65: Same as above

plugins/t/check_smtp.t
```diff
@@ -16,6 +16,8 @@ my $host_tcp_smtp_nostarttls = getTestParameter( "NP_HOST_TCP_SMTP_NOSTARTTLS",
 					   "A host providing SMTP without STARTTLS", "");
 my $host_tcp_smtp_tls        = getTestParameter( "NP_HOST_TCP_SMTP_TLS",
 					   "A host providing SMTP with TLS", $host_tcp_smtp);
+my $host_tls_cert            = getTestParameter( "NP_HOST_TLS_CERT",
```
RincewindsHat, line 19: I like the additonal sanity check, but not the name of this variable. It does not obviously tell me, that this is the CN in the certificate.
Maybe `certificate_common_name` or something like that?
What do you think?

RincewindsHat (changes requested):
I do like this in general, indeed it fixes a bug, but I do see some problems

plugins/check_smtp.c
franzs: Sure! I added a separate commit to preserve the original backported commits, though.

plugins/t/check_smtp.t
franzs: See my comment above... :wink:

plugins/check_smtp.c
franzs: Sorry for having introduced this issue. Although these were just backports from nagions-plugins... :innocent: 

If I would have been asked I would prefer option 2. In general it's better to be explicit instead of implicit. This change of the interface should be announced in the `NEWS` then.

franzs: @waja @RincewindsHat I added a check for the SSL/TLS method when a certificate check is requested. Meaning if `-D`/`--certificate` is given either `-s`/`--ssl`/`--tls` or `-S`/`--starttls` have to be set.

Please have a look at it if it suits you.

plugins/t/check_smtp.t
RincewindsHat: Agreed. Then, let's leave them for now.

plugins/t/check_smtp.t
RincewindsHat: agreed.

plugins/check_smtp.c
RincewindsHat: Since @waja seems to not have the time, I say we go the safe way and keep backwards compatibility. `-D` should assume starttls if `--tls` is not set explicitely.

I would like to change this in a major release though.

RincewindsHat:
Sorry @franzs, couldn't quite find the time lately. I made a decision about the whole option logic and have a wish about the preprocessor directives.
It would be gread if you could add those changes :-)

plugins/check_smtp.c
waja: It doesn't matter, from my (package maintainers) perspective, if you break backward compatibility with a minor or a major release.

sthen: Since this part has now been handled in a backwards-compatible way with 2eea6bb2a04bbfb, the remaining parts of this PR to allow checking validity of certs with both STARTTLS and TLS-on-connect seem straightforward

RincewindsHat:
@franzs Sorry to be annoying. Are you still on this?

PR #1911, opened 2023-09-04
check_smtp: Reimplement proxy protocol

franzs:
Add and use new function proxy_protocol_v1_header()
    
    check_smtp uses a dummy header for the proxy protocol without real IP
    addresses and a fixed port 25. Since check_smtp is able to use different
    ports with various options we add a new function to synthesize the header
    properly with real addresses and port numbers.
    
    As a side effect using a function other plugins can implement an option
    for proxy protocol quite easy, too.


franzs:
This PR waits for https://github.com/monitoring-plugins/monitoring-plugins/pull/1909 to be merged.

PR #1912, opened 2023-09-04
Remove translation comments

RincewindsHat:
`xgettext` adds comment to the original string - translation string pair where the original string occurs.
While this is helpful in general, it makes PRs with even small changes a mess since all the comments get updated if some change adds/removes a line.

This PR removes the generation of comments while generating the translation files and updates the translation files correspondingly.


RincewindsHat:
@waja, what do you think about this PR?

waja (approved):
While I like the comments, it's really a mess when creating RPs. So I think this might be the better way to go. If we change our mind at any day, we can it easily turn it on again.

PR #1916, opened 2023-09-05
Fix error in translation file introduced during a merge

RincewindsHat:
sorry about that. Should have let the test case run through

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -3409,8 +3409,7 @@ msgid ""
 "The server to test must be specified in the invocation, as well as a user"
 msgstr ""
 
-msgid ""
-"name and password. A configuration file must be present. The format of"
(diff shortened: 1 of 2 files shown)


RincewindsHat:
@waja sorry to be annoying. This should fix the problem of #1912

franzs:
@RincewindsHat Sorry for nitpicking. :innocent:

The change doesn't reflect the output of `make update-po` which produces just one line in `fr.po` and `de.po`:

```
msgid "name and password. A configuration file must be present. The format of"
```

So my suggestion is to run `make update-po` to have a better basis for further changes.

RincewindsHat:
@franzs satisfied? :-)

franzs:
Much better... :nerd_face: Thanks, @RincewindsHat!

PR #1917, opened 2023-09-07
Additional perl taint fixes

RincewindsHat:
This addresses parts of #1413 which probably should be addressed.
It replaces general calls to `snmpget` which are undertaken without a path prefix (so `PATH` is used, which could be manipulated) whith the absolute path as determined at compile time.

I am not totally sure, whether this is needed, but it homogenizes the usage of external executables across the plugins.


RincewindsHat:
It also addresses the unsetting of the `CDPATH` environment variable and therefore completely replaces #1413.
Thanks to @glensc for those changes.

RincewindsHat:
@sni : you know Perl, don't you? :-)

sni (approved):
lgtm

PR #1274, opened 2014-07-19
Several changes to remove NP-variables

waja:
Feel free to catch up part of this.

PR #1918, opened 2023-09-07
Renaming everything from Nagios Plugins to Monitoring Plugins

RincewindsHat:
Reintroducing #1274

Trying to rename everything from `np` to `mp`.
Tell me, what I forgot.


waja:
This will make it (just a little) harder to port patches over from NP. But maybe it's worth.

RincewindsHat:
Postponed until there are not PRs anymore in the way anymore

PR #1922, opened 2023-09-12
Update docs regarding C coding style

RincewindsHat:
This PR intends to update the development documentation a little in regard of more or less recent changes.

 * Recommend strongly the use "real" booleans instead of just using numbers
 * Allow the usage of the C++ style comments.


RincewindsHat:
@sni: Objections?

sni:
no objections, but i also don't know any reason why this has been in place at first.

RincewindsHat:
well, at least the boolean thing is from 2007, the comment stuff from 2003.

sni (approved):

PR #1924, opened 2023-09-22
Centralise and refactor maxfd related functionality

RincewindsHat:


sni (approved):
i am fine with that but what's the reason for moving it from utils to a separate file?

RincewindsHat:
The idea was/is, that it is a separate functionality mostly unrelated to other things, so putting it in it's own separate file seemed like the right thing.
But that's up to opinion. I you think it belongs somewhere else I am open to suggestions.

sni:
no worries, i was just wondering

PR #1925, opened 2023-09-22
Place _cmd_pids in object not header to avoid unsused variables

RincewindsHat:


waja (approved):
looks obvious for me ...

PR #1926, opened 2023-09-23
Remove check for RETSIGTYPE in autoconf stuff

RincewindsHat:
autoupdate tells me, that since C89 I can safely assume RETSIGTYPE is void.

Therefore to simplify things I removed the corresponding configure.ac line and replaced all mentions of RETSIGTYPE with void.


waja:
Looks like this is the case since a little... https://lists.isc.org/pipermail/inn-workers/2009-April/017010.html

RincewindsHat:
it is.

RincewindsHat:
@sni like to greenlight this?

sni (approved):
looks good to me

PR #1927, opened 2023-09-23
Somehow this fixes detection of the availability of struct timeval fo…

RincewindsHat:
…r me


RincewindsHat:
@sni opinions?

sni:
what was the actual issue?

RincewindsHat:
I got a compiler warning for have `struct timeval` redefined inside the project (`plugins/utils.h:55`) (it was already available on my system).
The part in the `configure.ac` checks for the availability of this type, but for some reason it did not set `HAVE_STRUCT_TIMEVAL` on my system.

Separating the the configure statements did the trick for me, althoug I cannot say why.

RincewindsHat:
@sni yes/no?

configure.ac
```diff
@@ -644,12 +644,16 @@ AC_TRY_COMPILE([#include <sys/time.h>],
...
                AC_DEFINE(HAVE_STRUCT_TIMEVAL,1,[Define if we have a timeval structure])
-               AC_TRY_COMPILE([#include <sys/time.h>],
-                              [struct timeval *tv;
-                               struct timezone *tz;
-                               gettimeofday(tv, tz);],
-                              AC_DEFINE(HAVE_GETTIMEOFDAY,1,[Define if gettimeofday is found]),
-                              AC_DEFINE(NEED_GETTIMEOFDAY,1,[Define if gettimeofday is needed])))
+			   FOUND_STRUCT_TIMEVAL="yes")
+
+if test x$FOUND_STRUCT_TIMEVAL = x"yes"; then
```
datamuc, line 649:
```suggestion
if test "$FOUND_STRUCT_TIMEVAL" = yes; then
```

waja: Ähm ... good catch, look wired.

waja: Anyway ... https://github.com/monitoring-plugins/monitoring-plugins/pull/1927/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810L630 looks equal

RincewindsHat: isn't that how one has to compare strings to be the comparison does the right thing if one of the strings is empty?

waja: I'd opt for

```suggestion
if test x"$FOUND_STRUCT_TIMEVAL" = x"yes"; then
```

or

```suggestion
if test "x$FOUND_STRUCT_TIMEVAL" = "xyes"; then
```

waja (approved):
okay ... dunno why this is needed, but looks good from syntax (any styling) ;)

PR #1928, opened 2023-09-23
check_icmp: Add support to Jitter, MOS and Score

datamuc:
trying to integrate #1503

after all those years :)


plugins-root/check_icmp.c
```diff
@@ -540,11 +579,34 @@ main(int argc, char **argv)
 				break;
 			case 'V': /* version */
 				print_revision (progname, NP_VERSION);
-				exit (STATE_UNKNOWN);
+				exit (STATE_OK);
```
sni, line 582: this might be slipped in and is unrelated to this patch. It should stay on unknown according to our guidelines:
`The --help or --version output should also result in Unknown state. `

sni:
thanks for getting onto this. Just a minor remark for now. Also a test case for the new options would be really nice, but no idea how hard that would be.

RincewindsHat (changes requested):
I make a superfical review and will have an in-depth look later.
Please address those points at your convenience

plugins-root/check_icmp.c
```diff
@@ -685,7 +757,7 @@ main(int argc, char **argv)
 	}
 
 	host = list;
-	table = (struct rta_host**)malloc(sizeof(struct rta_host **) * targets);
+	table = malloc(sizeof(struct rta_host *) * targets);
```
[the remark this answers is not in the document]
datamuc: 1ad7e163

plugins-root/check_icmp.c
```diff
@@ -1530,6 +1827,46 @@ get_threshold(char *str, threshold *th)
 	return 0;
 }
 
+/* not too good at checking errors, but it'll do (main() should barfe on -1) */
+static int
+get_threshold2(char *str, threshold *warn, threshold *crit, int type)
+{
+	char *p = NULL, i = 0;
```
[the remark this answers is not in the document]
datamuc: e365f9f5

plugins-root/check_icmp.c
```diff
@@ -224,6 +242,12 @@ static unsigned int warn_down = 1, crit_down = 1; /* host down threshold values
 static int min_hosts_alive = -1;
 float pkt_backoff_factor = 1.5;
 float target_backoff_factor = 1.5;
+int	rta_mode=0;
+int	pl_mode=0;
+int	jitter_mode=0;
+int	score_mode=0;
+int	mos_mode=0;
+int	order_mode=0;
```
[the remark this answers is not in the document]
datamuc: e365f9f5

plugins-root/check_icmp.c
```diff
@@ -1530,6 +1832,47 @@ get_threshold(char *str, threshold *th)
 	return 0;
 }
 
+/* not too good at checking errors, but it'll do (main() should barfe on -1) */
+static int
+get_threshold2(char *str, threshold *warn, threshold *crit, int type)
+{
+	char *p = NULL;
+	bool i = false;
```
RincewindsHat, line 1840: Sorry to be annoying, but what "does" `i`? and could it be renamed in a way to not be confused with the typical iterator?

RincewindsHat (changes requested):

plugins-root/check_icmp.c
datamuc: I don't know :shrug: I have not written that code. I just tried to make #1503 mergable and fixed some tests. NP is having it and people are requesting it for MP as well.

RincewindsHat: fair enough :-)

RincewindsHat:
ok, I am not entirely unhappy with that. @sni do you know more about the underlying logic and can take a look at that?
Otherwise i am prone to merge this.

datamuc:
We tried to do some research on where this is comming from:

* https://metacpan.org/dist/Algorithm-MOS/source/lib/Algorithm/MOS.pm#L35 (this is a known formula :rofl:)
* https://github.com/sipcapture/statstrmr/issues/2

They try to explain something without actually explaining anything:

https://netbeez.net/blog/impact-of-packet-loss-jitter-and-latency-on-voip/

So all we know is, that everybody seems to use this constants  and formula :shrug:

sni (approved):

sni:
looks good to me. I cannot say more to that logic than those links.

RincewindsHat (approved):

PR #1929, opened 2023-09-25
run tests with debian:stable

datamuc:
debian:testing is broken

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,10 +49,10 @@ jobs:
       fail-fast: false
       matrix:
         distro:
-          - 'debian:testing'
+          - 'debian:stable'
           #...
         include:
-          - distro: 'debian:testing'
+          - distro: 'debian:stable'
             prepare: .github/prepare_debian.sh
           #...
     steps:


waja:
Can you please explain, what is broken?

datamuc:
They removed squid

https://github.com/monitoring-plugins/monitoring-plugins/actions/runs/6302177559/job/17108828316

waja:
So maybe anybody has to investigate further and fix the issue within testing.

waja:
In a First view, the issue seems temporary by a transition: https://release.debian.org/transitions/html/auto-openldap.html

datamuc:
> In a First view, the issue seems temporary by a transition: https://release.debian.org/transitions/html/auto-openldap.html

Now we are in opinionland, so here is mine:

Sure, I'm convinced that this is temporary. But for an action that is run on every PR I want something more reliable. If you want to test your stuff against the next version of debian I would suggest an action that runs on a regular basis on the master/main branch and opens an issue if it fails or something like that.

But that's just my 2 ct.

RincewindsHat (approved):
Well, until the transition is done, I would like to have tests.
We can switch back later or add another pipeline for testing in general

PR #1930, opened 2023-10-01
Cleanup/check dhcp

RincewindsHat:
While trying to understand #1906 I found some rather irritating things in `check_dhcp`
which I am trying to fix here.


RincewindsHat:
@sni any objections?

sni (approved):
looks good

PR #1931, opened 2023-10-03
Fix some more typos

RincewindsHat:
Somehow codespell finds them just now.


waja (approved):
Looks good. Dunno about the whitelist of `esponse`, but might be legit.

PR #1932, opened 2023-10-06
use pack_sockaddr_in rather than hand-rolled

RincewindsHat:
On some OS, sockaddr structs include a length field. Perl's pack_sockaddr_in takes this into account; the hand-rolled "pack('S n a4 x8'..." doesn't do so, resulting in connection failures.

This replaces #1463 since github doesn't let me update the original branch


RincewindsHat:
@sni would you like to take a quick look at this one?

sni (approved):

sni:
sure, looks good

Issue #888, opened 2013-09-24
Range support for check_nt [sf#2804825]


monitoring-user:
Submitted by rakhun on 2009-06-11 14:43:20

Plugin name and version: check_nt v1.4.13.138.g9eab.dirty (nagios-plugins 1.4.13)
Example commandline: plugins/check_nt -H IP -v USEDDISKSPACE -l C -w 50 -c 5:95
Tested on: Fedora 10, i686 with compiler gcc (GCC) 4.3.2 20081105 (Red Hat 4.3.2-7)
The patch makes it possible to determine the status based on ranges, like in the example where the status is critical if disk usage is below 5 or above 95 (not the best example, but something might be seriously wrong if the usage gets too low)

RincewindsHat:
Link is dead, closing this

sni:
Btw, those links have been added long time ago, meanwhile they are available under https://www.monitoring-plugins.org/attachments/330612-rangesupport.patch

RincewindsHat:
oh, didn't know that. That make this kinda awkward.

RincewindsHat:
Created an PR, therefore closing the issue

PR #1934, opened 2023-10-09
check_nt: Range support

RincewindsHat:
Transform issue #888 into a PR.

> Submitted by rakhun on 2009-06-11 14:43:20

> Plugin name and version: check_nt v1.4.13.138.g9eab.dirty (nagios-plugins 1.4.13)
> Example commandline: plugins/check_nt -H IP -v USEDDISKSPACE -l C -w 50 -c 5:95
> Tested on: Fedora 10, i686 with compiler gcc (GCC) 4.3.2 20081105 (Red Hat 4.3.2-7)
> The patch makes it possible to determine the status based on ranges, like in the example 
> where the status is critical if disk usage is below 5 or above 95 (not the best example, but > something might be seriously wrong if the usage gets too low)

### DIFF
--- a/lib/utils_base.c
+++ b/lib/utils_base.c
@@ -726,3 +726,48 @@ void np_state_write_string(time_t data_time, char *data_string) {
 
 	np_free(temp_file);
 }
+
+char *fperfdatarange (const char *label,
+		double val,
+		const char *uom,
+		const char *warn,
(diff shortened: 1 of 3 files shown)


RincewindsHat:
NSclient is pretty dead, so is this.

PR #1936, opened 2023-10-15
check_curl: Compiler warnings cast function type

RincewindsHat:
Fixes for `CFLAGS="-Wcast-function-type"`


waja (approved):
Looks reasonable

PR #1941, opened 2023-10-16
only run tests on the master branch and on PRs

datamuc:


waja:
What's the motivation about this?

datamuc:
it's a suggestion for this:
```
<rincewind> does anybody know enough github to disable double tests if something is a
                       PR and a branch of the main repo?

waja:
> <rincewind> does anybody know enough github to disable double tests if something is a
>             PR and a branch of the main repo?

But this PR disables also testing every branch beside master, doesn't it?

datamuc:
> But this PR disables also testing every branch beside master, doesn't it?

correct, you might add `- maint-*` to the branches list. But other branches are only tested if you open a PR for them.

RincewindsHat:
The original reason for my comment there was my annoyance, that creating a PR from a branch of the main repo would trigger the tests two times which seems rather unnecessary to me.
I would rather not have it like this PR suggests it though, since now contributors will not get tests before opening a PR.

Is there a good too just avoid triggering the pipelines multiple times?
Otherwise this could/probably solved organisationaly by just not working in the main repo

Issue #1943, opened 2023-10-18
lib/maxfd.h missing from https://www.monitoring-plugins.org/download/monitoring-plugins-2.3.4.tar.gz


hydrapolic:
Hello, trying to build v2.3.4 from official website, however maxfd.h is missing from the package. It's included in the github source code thought.

RincewindsHat:
Hi @hydrapolic, thanks for reporting, we're on it

hydrapolic:
Thanks!

PR #1944, opened 2023-10-18
extend github action tests

sni:
add some more steps to the test pipeline:

    - make dist
    - builds from the tarball above

this test should fail until #1943 is fixed.


sni:
good, so this PR failed before #1945 and is green afterwards, so it does exactly what expected.

waja (approved):
looks good

PR #1947, opened 2023-10-18
Removing translation files

waja:
Unfortunately there is very less coverage of the translations and was not very much contribution recently this area. On the other side, maintaing the po files consumes a significant effort. For now we drop the translation files, but we can revert this change anytime in the future.


waja:
Looks like `gl/` has issues when removing the translations.

waja:
> Looks like `gl/` has issues when removing the translations.

Help is welcomed. :)

RincewindsHat:
When I find the time :-)

RincewindsHat:
@waja  better now? :-)

waja:
> @waja  better now? :-)

WTF?!? How could I've overseen this?

RincewindsHat:
so, go?

waja:
Technical it looks good and checks are succeeding... so why not? ;)

RincewindsHat (approved):

PR #1948, opened 2023-10-18
Replace most of the boolean variables declared as "int" with "bool"

RincewindsHat:
This PR aims to replace all "old school boolean" (`ints` declared as `ints`) with the C99 `bool` (probably still `int` but the compiler can do whatever it wants here).

This should enhance readability, might prevent some really weird edge-case bugs and allows for some optimizations from the compiler side.

Mostly though the readability thing.

There a quite probably some variables which I did not find on this first try.

Also this is probably what @sni calls a "PR from hell" :-)

I can split it up if this is wished for.

### DIFF
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -50,7 +50,7 @@ unsigned int services = 0;
 int skip_stdout = 0;
 int skip_stderr = 0;
 int warn_on_stderr = 0;
-bool unknown_timeout = FALSE;
+bool unknown_timeout = false;
 char *remotecmd = NULL;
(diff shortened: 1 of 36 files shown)


RincewindsHat:
Fun fact: This actually breaks stuff which I see only just now `-Wall`. There are some preprocessor macros in `check_npt_time`.

waja:
> Also this is probably what @sni calls a "PR from hell" :-)

Wow, what a patch set. 😀 🚀

sni (approved):

sni:
looks good, and since it's only one type of change, it's actually pretty readable :)

PR #1952, opened 2023-10-19
check_smtp: Restore behaviour pre ead5526efa4f713e8001baed409067b0474…

RincewindsHat:
…cb72d regarding -D and TLS


sni (approved):
tested and works as intended

PR #1956, opened 2023-10-26
Fix check mailq

sni:
check_mailq: exit on empty strings and exit early

in case of $utils::PATH_TO_MAILQ beeing an empty string, the "defined" is still true
and leads to misleading error messages.

While on it, rewrite cascaded if/elsifs to more readably exit-early ifs.


sni:
the first two commits only change whitespace, the 3rd makes the ifs more readable and replaces "defined" with a simple string check.

PR #1960, opened 2023-11-09
check_ups: output ups.realpower if supported

sbraz:
Hi,
This adds support for the _ups.realpower_ variable which contains the output power of the UPS.
The output looks like this on my system:
```
UPS OK - Status=Online Utility=221.0V Batt=87.0% Load=32.0% Real power=255.0W |voltage=221000mV;;;0; battery=87%;;;0;100 load=32%;;;0;100 realpower=255W;;;0;
```
I'm not sure whether we should have "Real power" or "Realpower" in the human-readable output, please let me know what the preferred form is.

### DIFF
--- a/plugins/check_ups.c
+++ b/plugins/check_ups.c
@@ -46,12 +46,13 @@ enum {
 
 #define CHECK_NONE	 0
 
-#define UPS_NONE     0   /* no supported options */
-#define UPS_UTILITY  1   /* supports utility line voltage */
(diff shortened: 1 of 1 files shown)


RincewindsHat (approved):
Hi @sbraz,
Thank you very much for your contribution.
I don't have UPS to test right here, but it seems to me like your code does not change any other behaviour, so this is good enough for me :-)

PR #1962, opened 2023-11-13
CI: Switch back to testing

waja:
This reverses 9c1c3fc

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,10 +49,10 @@ jobs:
       fail-fast: false
       matrix:
         distro:
-          - 'debian:stable'
+          - 'debian:testing'
           #...
         include:
-          - distro: 'debian:stable'
+          - distro: 'debian:testing'
             prepare: .github/prepare_debian.sh
           #...
     steps:


RincewindsHat:
@waja sorry for letting this lie around. Question is, should be switch back to testing.
One of the arguments there was, that testing on stable is probably less hassle.
At least seeing this from a dev perspective and less of the package maintainter one.

Maybe a middle way would be nice. doing regular testing on stable and have the same pipeline on manual or time-based trigger for testing?

waja:
> One of the arguments there was, that testing on stable is probably less hassle.

We did not have much hassle in the past all over the years testing against Debian testing, as far as I remember. But had the opportunity to catch up issues with more recent software revisions.

RincewindsHat:
we at least had the issue with the snmp mibs thing. But I don't really have strong opinion here. Might as well use testing.

waja:
> we at least had the issue with the snmp mibs thing. But I don't really have strong opinion here. Might as well use testing.

Don't remember exactly the "fix", but wasn't that a kind of "more recent software revisions"? 🤔

RincewindsHat (approved):

waja:
@RincewindsHat much appreciated

PR #1966, opened 2023-11-23
check_load: cleanup

### LINKED ISSUE #1965: check_load: uses #ifdef HAVE_PROC_LOADAVG which was removed from the configure script
There are several `#ifdef HAVE_PROC_LOADAVG` blocks, e.g. https://github.com/monitoring-plugins/monitoring-plugins/blob/f054ab04e788405932b9bb081747bc7c8328a0d8/plugins/check_load.c#L130-L151

These blocks will never be compiled, since that symbol is not defined.  It used to be a configure check, but that was removed a _long_ time ago, in fe856aa (configure.in line 609).

It seems that using `getloadavg()` is preferred where available, so the `/proc/loadavg` fallback may not be that common.  Regardless, it should either be removed or wired back into configure.ac; I have no idea which would be preferred.

I don't think there's a need to bring back the --with-proc-loadavg=_path_ parameter (what else would it ever be?), but perhaps we should check for the existence of `/proc/loadavg`?

(Noticed via https://bugzilla.suse.com/show_bug.cgi?id=1216148#c9, which turned out to be an unrelated problem.)

RincewindsHat:
closes #1965

Escalated to a refactoring:
 * no more global variables
 * code formatting
 * other "improvements"
 * Updated copyright

### DIFF
--- a/plugins/check_load.c
+++ b/plugins/check_load.c
@@ -67,7 +67,7 @@ static processed_load_config process_arguments(int argc, char **argv);
 static int validate_arguments(const load_config);
 void print_help(void);
 void print_usage(void);
-static int print_top_consuming_processes(const load_config);
+static int print_top_consuming_processes(const int);
 
 /* strictly for pretty-print usage in loops */
 static const int nums[3] = {1, 5, 15};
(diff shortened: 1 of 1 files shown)


sni:
these are cosmetical changes only, right?

RincewindsHat:
That was my intention, yes.

PR #1970, opened 2023-12-06
plugins-root: prevent -lcrypto from showing up in Makefile dependencies

EricFromCanada:
Mirroring [this issue](https://github.com/nagios-plugins/nagios-plugins/issues/508), this change allows building with Homebrew by delaying when `$(LIB_CRYPTO)` is added.

### DIFF
--- a/plugins-root/Makefile.am
+++ b/plugins-root/Makefile.am
@@ -26,7 +26,7 @@ EXTRA_PROGRAMS = pst3
 
 EXTRA_DIST = t pst3.c
 
-BASEOBJS = ../plugins/utils.o ../lib/libmonitoringplug.a ../gl/libgnu.a $(LIB_CRYPTO)
+BASEOBJS = ../plugins/utils.o ../lib/libmonitoringplug.a ../gl/libgnu.a
 NETOBJS = ../plugins/netutils.o $(BASEOBJS) $(EXTRA_NETOBJS)
 NETLIBS = $(NETOBJS) $(SOCKETLIBS)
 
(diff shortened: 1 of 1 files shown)


RincewindsHat:
Hi, thank you for the contribution.
It still builds, so pragmatically this is ok I guess. It bothers me though, since I think this is technically and the right place for `-lssl` is in the current line, since it belongs to the lib and not the individual plugins.
Can we somehow find out what the underlying issue is?

symphorien:
I needed this patch to bump monitoring plugins from 2.3.0 to 2.3.5 on NixOS

waja:
> I needed this patch to bump monitoring plugins from 2.3.0 to 2.3.5 on NixOS

Pardon?

symphorien:
NixOS currently has 2.3.0. I tried to update that to 2.3.5, encountered a build issue about -lcrypto, applied the patch, which fixed the build. I mentioned that because
- this provides the information that this fixes a regression from 2.3.0
- I wanted to show that there is interest/need for this patch.

RincewindsHat (approved):
Not sure if this is the best solution, but it seems to be one, so why not.

Thank you very much.

RincewindsHat:
Ah, I think I found the original issue. In this [line](https://github.com/monitoring-plugins/monitoring-plugins/blob/b1442d73d21cc2ce14328069d4f8532eb2cf7f83/plugins-root/Makefile.am#L92)
and the following:

```
check_dhcp_DEPENDENCIES = check_dhcp.c $(NETOBJS) $(DEPLIBS)
check_icmp_DEPENDENCIES = check_icmp.c $(NETOBJS)
```

`NETOBJS` is introduced as a dependency and since we hat those  [lines:](https://github.com/monitoring-plugins/monitoring-plugins/blob/b1442d73d21cc2ce14328069d4f8532eb2cf7f83/plugins-root/Makefile.am#L29)

```
BASEOBJS = ../plugins/utils.o ../lib/libmonitoringplug.a ../gl/libgnu.a $(LIB_CRYPTO)
NETOBJS = ../plugins/netutils.o $(BASEOBJS) $(EXTRA_NETOBJS)
```

where `-lssl` is added to be used for linking. This will create a Makefile target for `-lssl` then which does not exist.
Now I am wondering why this did work on other setups.

Issue #1874, opened 2023-04-24
check_swap: reports wrong swap size (~ factor 1024)


mrdsam:
Since Version 2.3.3 OR Version 2.3.2 (haven't testet this one) check_swap reports the wrong swap size by about factor 1024. Version 2.3.1 was still OK.

Both, plugin-output and performance-data are wrong.

My systems do have a 1024 MB swap partition, check_swap reports 1204223 MB. Here an example output, same system, 2 different check_swap binaries:

```
#
# New Version, wrong output
#
~$ /usr/local/libexec/nagios/check_swap --version
check_swap v (monitoring-plugins 2.3.3)

~$ /usr/local/libexec/nagios/check_swap -c 25% -w 50%
SWAP OK - 100% free (1204223MB out of 1204223MB) |swap=1262719336448B;631359668200;315679834100;0;1262719336448

#
# Older Version, right output
#
~$ ./check_swap --version
check_swap v (monitoring-plugins 2.3.1)

~$ ./check_swap -c 25% -w 50%
SWAP OK - 100% free (1023 MB out of 1023 MB) |swap=1023MB;511;255;0;1023
```
OS: FreeBSD 13.1-RELEASE-p6

mrdsam:
Note: Because I can't tell whether this is a FreeBSD-only problem, I also opened a bug report at 
[bugs.freebsd.org](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271042) #271042
This bug report also links back here.

sni:
the issue has been introduced by 4621427ba8cbfab4815e65d35a728ad51ad8c391

RincewindsHat:
hm, feared that

sni:
changing the swap_format to %*s %d %*d %d seems to fix it, but is has been there for ages.

```
freebsd13:/tmp/monitoring-plugins %>./plugins/check_swap -vvv -w 5% -c 5%
Command: /usr/sbin/swapinfo -k
Format: %*s %d %*d %d
total=0, free=0
SWAP OK - 100% free (4095MB out of 4095MB) |swap=4293918720B;214695935;214695935;0;4095
```
Then the only wrong value left is the total value at the end which should be bytes as well.

RincewindsHat:
Interesting, I get a different swap format.

RincewindsHat:
any chance, that the output of `swapinfo` changed?

RincewindsHat:
[...]

PR #1972, opened 2023-12-13
Fix issue 1874 (error in free calculation)

Napsty:
This should fix the wrong "free" calculation mentioned in #1874


sni:
This has "just" been changed with
https://github.com/monitoring-plugins/monitoring-plugins/pull/1642
to exclude the cached value. There are examples with and without the cached swap.
I don't want to change that back and forth. But i would agree to a switch which makes
this a user decision. But keep the default like it is right now.

Napsty:
OK thx for the decision. I am therefore closing this PR.

PR #1976, opened 2023-12-20
Refactor check_swap

RincewindsHat:
This PR is the result of refactoring `check_swap`, which was started because the `ifdef` hell in there was annoying me.

The results are the following:

 * The actual fetching of swap data is now outsourced to a separate translation unit (`swap.c`) in a separate directory (`check_swap.d`) although the naming and the positions are up for discussion
 * A first little unit was introduced/started with `plugins/tests/test_check_swap.c` (and the corresponding curl test program) 
 * The `main` in `check_swap` now only deals with computing the result and formatting the output
 * Updated copyright
 * Code was formatted with `clang-format` at least once to achieve better readability

Currently this is probably not in a proper state to be merged, more tests would be appreciated and the structure of the files and so on is open for discussion.


RincewindsHat:
@sni would you like to take a look at this?

sni:
yeah, when i have 3 weeks of vacation ;)

RincewindsHat:
Can I sweet talk your company into giving you that? :-P

PR #1977, opened 2023-12-22
Use debian stable for main tests again

RincewindsHat:
This switches the main testing stuff back to debian stable since testing is broken again.

I would propose to stay with this and add workflows against testing based on manual or timer based triggers on the `master` branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,10 +49,10 @@ jobs:
       fail-fast: false
       matrix:
         distro:
-          - 'debian:testing'
+          - 'debian:stable'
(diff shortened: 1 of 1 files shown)


waja:
Can please anybody point me to a log where I can inspect the breakerage? Didn't see broken pipelines at a first glance.

RincewindsHat:
https://github.com/RincewindsHat/monitoring-plugins/actions/runs/7298953690/job/19890859019

waja:
Okay. This should be just a temporary issue. Anyway, doesn't mind. Just go ahead.

Issue #1416, opened 2016-04-18
output in JSON


FinlayDaG33k:
This is more an enhancement than an issue, but please add a function to give the output in JSON... this makes is alot easier to work with in for example PhP

riskersen:
Hi, what output do you mean? Do you mean in general for all plugins?

FinlayDaG33k:
Yes, as it is right now quite a pain on the brain to use the output of the plugins in for example PHP scripts... I know you can use something my friends call "tokenization" but having the ability to have the script output in JSON is way user-friendlier in my opinion.

RincewindsHat:
this was implemented partially (for most of the `C`-based plugins) with version 3.0.0
The rest will likely follow.

Issue #1098, opened 2013-09-24
drop "SERVICE STATUS: " from all plugins output + the guide [sf#3572875]


monitoring-user:
Submitted by calestyo on 2012-09-29 01:01:00

Hi.

First, the Nagios plug-in development guidlines (http://nagiosplug.sourceforge.net/developer-guidelines.html) seem to be quite outdated (e.g. not really knowing about multi-line output, etc.).

But in general I propose the following (and would also be willing to help with patches, if it was accepted):
- In section 2.1. (http://nagiosplug.sourceforge.net/developer-guidelines.html#AEN33) of the Nagios plug-in development guidlines, drop the recommendation of adding "SERVICE STATUS: " to the output and actually recommend _against_ it.
- In all plugins from the Nagios plugin package, remove the according output.

Reasons and motivation:
1) The output is superfluous:
With respect to the status: All clients/UIs/GUIs I've ever seen (including the majors, Nagios-CGI, Icinga-CGI and Icinga-Web) use (indirectly) the exit status of the programs (as it is defined) and show it anyway in a separate field, and/or marked with colours.
With respect to the "plugin name": All clients/UIs/GUIs I've ever seen (including the majors, Nagios-CGI, Icinga-CGI and Icinga-Web) show that name of the check_command anyway.

2) The ouput is not authoritative (and maybe wrong or at least confusing).
With respect to the status: There is no guarantee, that plugins _print_ the right status; only the exit code is authoriative with respect to Nagios/Icinga.
So at best it's correct (but doubled) informationâ€¦ at worst itâ€™s wrong and confusing.
With respect to the "plugin name": The actual name (encoded in a check program) is not necessarily equal or even related to the way the check is used. A (poor example), check_ping is also used as host-available-check... another example,.. one might even use check_ssh as host-available-check.
Printing SSH OK there is rather confusing; the service name (i.e. the description) is usually much better information on what is checked.

3) It wastes precious space and adds confusing text
"SERVICE STATUS: " wastes text space not only when displaying stuff in moibile devices, but also on normal desktop computer, where one e.g. wants to use the space for displaying more columns with information abot a check.... or more from the "actual" text outpu.

Further it "misaligns" output, e.g. consider check_apt which shows me the number of available updates, which would be about like that:
APT OK: 0 security updates, 0 other updates
APT WARNING: 0 security updates, 2 other updates
APT CRITICAL: 1 security updates, 5 other updates

[...]

PR #1979, opened 2023-12-28
New output infrastructure

RincewindsHat:
This PR introduces a lot of new stuff to `lib` part which is mostly related to generating the results.

The idea is, that a plugin has a main result object (of type `mp_check`) to which you add partial results, called *subchecks* (`mp_subcheck`).
This `mp_subcheck`s can also contain `mp_subcheck`s, so you have a tree of results.

Ideally we remove all of that specific logic/string handling and so on from the individual plugins, which would reduce complexity (in the plugins).
Also this would open the path to centrally structure and format output, which would be cool for me, since I have a hard time regarding the output of some plugins (*cough* `check_disk`).

The handling would be something like

```
mp_check main = mp_check_init();

subcheck1 = check_filesystem_A();
subcheck2 = check_filesystem_B();
subcheck3 = check_something_else();

mp_add_subcheck_to_check(main, subcheck1);
mp_add_subcheck_to_check(main, subcheck2);
mp_add_subcheck_to_check(main, subcheck3);

mp_exit(main);
```

No plugin was changed here, the functionality is implemented in
parallel to the existing stuff (mostly), so it can be included
and plugins can be adapted incrementaly following that (which I plan to do).

## Problems

 * This introduces features from C11/C17, so this will break backwards compatibility


RincewindsHat:
Also, sorry @sni, this is likely going in your direction.

RincewindsHat:
building block for a solution to #1098

RincewindsHat:
starting point for #1416

RincewindsHat:
replaced by #2064

Issue #1523, opened 2017-11-09
[check_mysql] invalid pointer exception while using --extra-opts


Kaelnor:
Hi,

I'm trying to use the --extra-opts option with check_mysql to pass username and password arguments.

I'm using monitoring-plugins 2.1.2-2ubuntu2 from Ubuntu 16.04.3 repos.

Here is the result:

> /usr/lib/nagios/plugins/check_mysql --extra-opts -S
*** Error in `/usr/lib/nagios/plugins/check_mysql': munmap_chunk(): invalid pointer: 0x000000000040acc8 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f589b97e7e5]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x1a8)[0x7f589b98b698]
/usr/lib/nagios/plugins/check_mysql[0x4074bb]
/usr/lib/nagios/plugins/check_mysql[0x406eed]
/usr/lib/nagios/plugins/check_mysql[0x402430]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f589b927830]
/usr/lib/nagios/plugins/check_mysql[0x402e39]
======= Memory map: ========
00400000-0040c000 r-xp 00000000 fc:00 156189                             /usr/lib/nagios/plugins/check_mysql
0060c000-0060d000 r--p 0000c000 fc:00 156189                             /usr/lib/nagios/plugins/check_mysql
0060d000-0060e000 rw-p 0000d000 fc:00 156189                             /usr/lib/nagios/plugins/check_mysql
01f83000-01fb5000 rw-p 00000000 00:00 0                                  [heap]
7f589a970000-7f589ac48000 r--p 00000000 fc:00 136572                     /usr/lib/locale/locale-archive
7f589ac48000-7f589ac5e000 r-xp 00000000 fc:00 391565                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f589ac5e000-7f589ae5d000 ---p 00016000 fc:00 391565                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f589ae5d000-7f589ae5e000 rw-p 00015000 fc:00 391565                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f589ae5e000-7f589af66000 r-xp 00000000 fc:00 391580                     /lib/x86_64-linux-gnu/libm-2.23.so
[...]

PR #1982, opened 2024-01-07
Ini Parser: Avoid freeing symbols from text section

RincewindsHat:
Avoid freeing symbols from text section by remembering where they are from


RincewindsHat:
Closes #1523

sni (approved):

PR #1987, opened 2024-02-12
Improve negate plugin helptext

euniceremoquillo:
This commit updates the negate plugin's helptext as it is currently ambiguous.

### DIFF
--- a/plugins/negate.c
+++ b/plugins/negate.c
@@ -220,8 +220,10 @@ print_help (void)
 
 	printf (COPYRIGHT, copyright, email);
 
-	printf ("%s\n", _("Negates the status of a plugin (returns OK for CRITICAL and vice-versa)."));
-	printf ("%s\n", _("Additional switches can be used to control which state becomes what."));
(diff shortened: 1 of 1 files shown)


plugins/negate.c
```diff
@@ -220,8 +220,10 @@ print_help (void)
 
 	printf (COPYRIGHT, copyright, email);
 
-	printf ("%s\n", _("Negates the status of a plugin (returns OK for CRITICAL and vice-versa)."));
-	printf ("%s\n", _("Additional switches can be used to control which state becomes what."));
+	printf ("%s\n", _("Negates only the return code of a plugin (returns OK for CRITICAL and vice-versa) by default."));
+	printf ("%s\n", _("Additional switches can be used to control:\n"));
+	printf ("- which state becomes what\n");
```
RincewindsHat, line 225: I think this would look better with some indentation, to make it look more like "this is a listing related to the line above".
Would you aggree?

RincewindsHat (approved):
Looks good to me. Thank you very much :-)

PR #1990, opened 2024-02-28
check_icmp: keep performance data order in case of none-reachable hosts

sni:
For responding hosts nothing changes:

    ./check_icmp -H 10.0.0.1
    OK - 10.0.0.1 rta 0.336ms lost 0%|rta=0.336ms;200.000;500.000;0; rtmax=0.471ms;;;; rtmin=0.000ms;;;; pl=0%;40;80;0;100

In case the host is down, the performance data now contain 'U' according to: https://www.monitoring-plugins.org/doc/guidelines.html#AEN201

Old:
    ./check_icmp -H 10.0.0.99
    CRITICAL - 10.0.0.99: Host unreachable @ 10.0.1.138. rta nan, lost 100%|pl=100%;40;80;0;100

New:
    ./check_icmp -H 10.0.0.99
    CRITICAL - 10.0.0.99: Host unreachable @ 10.0.1.138. rta nan, lost 100%|rta=U;;;; rtmax=U;;;; rtmin=U;;;; pl=100%;40;80;0;100

This confuses some performance graphing tools, ex.: pnp.

### DIFF
--- a/plugins-root/check_icmp.c
+++ b/plugins-root/check_icmp.c
@@ -1547,46 +1547,66 @@ finish(int sig)
 	while(host) {
 		if(debug) puts("");
 
-		if (rta_mode && host->pl<100) {
-			printf("%srta=%0.3fms;%0.3f;%0.3f;0; %srtmax=%0.3fms;;;; %srtmin=%0.3fms;;;; ",
-			   (targets > 1) ? host->name : "",
-			   host->rta / 1000, (float)warn.rta / 1000, (float)crit.rta / 1000,
-			   (targets > 1) ? host->name : "", (float)host->rtmax / 1000,
-			   (targets > 1) ? host->name : "", (host->rtmin < INFINITY) ? (float)host->rtmin / 1000 : (float)0);
+		if (rta_mode) {
+			if (host->pl<100) {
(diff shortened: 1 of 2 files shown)


RincewindsHat:
The second "Old" should be a "New" I guess?
And the pnp contains null values instead of empty ones?

sni:
of course, i fixed the typo. Yes, pnp (and others) can handle the U and plot a <null> value instead of a straight line because there was no data at all.

RincewindsHat:
I tend to call this a bug in pnp (at least it should be possible to specify what to do with missing values), but that does not help us, does it?
I am going to take a look later.

sni:
it's only an issue if you use STORAGE_TYPE=multiple for the rrd files, so it's not even a
bug in pnp. Anyway, it is definitely better to precisely tell that there are no performance data, than just leave them away one time and then show them another time.

RincewindsHat (approved):

RincewindsHat:
ok, looks good to me, but in principle we have that problem all over the place and can not properly solve it.
That would be a problem for the monitoring/graphing system.

sni:
no, usually this is no problem. In most cases you either have performance data or not. The specific issue here was, that under some circumstances only a part of the performance data showed up. To my knowledge, this does not happen with other plugins (and did not happen with this plugin previously)

PR #2000, opened 2024-04-12
check_swap: Possibility to run check_swap without thresholds

Napsty:
This PR adds the possibility to run check_swap without having to define the thresholds. Previously to that PR, check_swap could not be run without setting thresholds.

The plugin returns OK if no thresholds are defined. This can be very helpful if Swap Usage is purely monitored for informational (and observability) purposes but without any need to alert.

```
$ ./check_swap
SWAP OK - 3% free (208MB out of 8651MB) |swap=218103808B;0;0;0;9071230976
```

Thresholds still work, of course, but are not mandatory anymore:

```
$ ./check_swap -w 10% -c 5%
SWAP CRITICAL - 3% free (208MB out of 8651MB) |swap=218103808B;907123090;453561545;0;9071230976
```

As thresholds were a must before, the change from this PR will not break any existing configurations, as they already have thresholds defined.

### DIFF
--- a/plugins/check_swap.c
+++ b/plugins/check_swap.c
@@ -401,7 +401,7 @@ check_swap(float free_swap_mb, float total_swap_mb)
 	uint64_t free_swap = free_swap_mb * (1024 * 1024);		/* Convert back to bytes as warn and crit specified in bytes */
 	uint64_t usage_percentage = ((total_swap_mb - free_swap_mb) / total_swap_mb) * 100;
 
-	if (warn.value && crit.value) {                                 /* Thresholds defined */
+	if (warn.value || crit.value) {                                 /* Thresholds defined */
 	        if (!crit.is_percentage && crit.value >= free_swap) return STATE_CRITICAL;
 	        if (!warn.is_percentage && warn.value >= free_swap) return STATE_WARNING;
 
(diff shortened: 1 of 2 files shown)


sni:
i like the idea, but if i read it correctly you now have to either use no thresholds or both. Why not allow only warning or only critical thresholds as well?

Napsty:
> have to either use no thresholds or both

That's the same behaviour as before (both warning and critical were mandatory)

> Why not allow only warning or only critical thresholds as well?

Umm haven't thought about that tbh. But can be done, sure. Want me to adjust the PR?

sni:
> Want me to adjust the PR?

that would be nice.

Napsty:
Seems to be working:

```
$ ./check_swap 
SWAP OK - 3% free (208MB out of 8651MB) |swap=218103808B;0;0;0;9071230976

$ ./check_swap -w 10%
SWAP WARNING - 3% free (208MB out of 8651MB) |swap=218103808B;907123090;0;0;9071230976

$ ./check_swap -c 5%
SWAP CRITICAL - 3% free (208MB out of 8651MB) |swap=218103808B;0;453561545;0;9071230976

$ ./check_swap -w 50% -c 10%
SWAP CRITICAL - 3% free (208MB out of 8651MB) |swap=218103808B;4535615450;907123090;0;9071230976
```

sni:
while on it, could you add those tests to https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/t/check_swap.t

Napsty:
> while on it, could you please

only because it's Friyay :grin: 

Will add shortly

Napsty:
Once I figured out how the tests work, this should now also be working :smile:

sni (approved):

sni:
thank you very much

PR #2005, opened 2024-07-23
Update versions to reflect the actual state, we forgot about this in …

RincewindsHat:
…our last urgent release


waja (approved):
Looks good

PR #2007, opened 2024-07-23
check_http: replace www.mozilla.com with monitoring-plugins.org

sni:
this makes tests more reliable if we test our own sites instead some 3rd party site.


waja (approved):
Looks sane ... LGTM

PR #2010, opened 2024-07-29
check_curl: fix relative redirects on non-standard port

sni:
Having a webserver respond with a relative redirect as for ex. in `Location: /path/to.html` check_curl would use the wrong standard http/https port instead of crafting the absolute url using the given scheme/hostname and port.

Adding a new test case for this for check_http and check_curl. check_http did it correct already, so no fix necessary there.

before:

    %>./check_curl -H 127.0.0.1 -p 50493 -f follow -u /redirect_rel -s redirected -vvv
    **** HEADER ****
    HTTP/1.1 302 Found
    ...
    Location: /redirect2

    ...
    * Seen redirect location /redirect2
    ** scheme: (null)
    ** host: (null)
    ** port: (null)
    ** path: /redirect2
    Redirection to http://127.0.0.1:80/redirect2

Port 80 is wrong here.

fixed:

    %>./check_curl -H 127.0.0.1 -p 50493 -f follow -u /redirect_rel -s redirected -vvv
    **** HEADER ****
    HTTP/1.1 302 Found
    ...
    Location: /redirect2

    ...
    * Seen redirect location /redirect2
    ** scheme: (null)
    ** host: (null)
    ** port: (null)
    ** path: /redirect2
    Redirection to http://127.0.0.1:50493/redirect2

Correctly using the given port to craft the new url.

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -1285,10 +1285,12 @@ redir (curlhelp_write_curlbuf* header_buf)
     }
   }
 
-  if (!uri_strcmp (uri.scheme, "https"))
-    use_ssl = true;
-  else
-    use_ssl = false;
+  if (uri.scheme.first) {
+    if (!uri_strcmp (uri.scheme, "https"))
+      use_ssl = true;
+    else
+      use_ssl = false;
+  }
 
   /* we do a sloppy test here only, because uriparser would have failed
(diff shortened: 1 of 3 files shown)


RincewindsHat (approved):
Looks sane.

PR #2012, opened 2024-08-16
check_curl: Documentation for --certificate, --cookie-jar

oxzi:
From the mere help output for -C / --certificate, I was confused about what its two integer parameters do. Unfortunately, I also missed out on the explaining examples later. Since I like to have basic documentation for each flag, I tried to make the arguments as short as possible.

The other fix was one hyphen too many for the --cookie-jar option.


RincewindsHat (approved):
Thanks a lot :-)

PR #2014, opened 2024-08-19
Check curl regex state

RincewindsHat:
This mainly fixes a problem introduced in #2006 ( f29785f5035e489576d2ba74be774273b73dd149 ), where the option really is `state-regex` but the help says `regex-state`.


andreasbaumann (commented):
Looks good. Thanks for fixing all the things I left out.. :-)

RincewindsHat:
@andreasbaumann would you like to press the "approve" button? :-)

andreasbaumann (approved):
approving (thats a more than just a button) ;-)

PR #2016, opened 2024-08-28
Fix/check ssh variable stuff

RincewindsHat:


sni (approved):

RincewindsHat:
Sorry @sni, thought about it a bit and it looks like `elapsed_time` was previously read before being written to, which was a bug I guess.

sni:
sounds like we are missing some test cases then :-)

RincewindsHat:
ok, investigated a bit. There was only a problem when `-P` (expect a certain remote protocol) was used.

PR #2018, opened 2024-09-08
configure.ac: rewrite PostgreSQL detection using pg_config

orlitzky:
Copy of https://github.com/nagios-plugins/nagios-plugins/pull/787 :)

PostgreSQL installs a pg_config program that spits out its "include" and "library" paths when called with --includedir and --libdir, respectively. Using this we can simplify the detection of PostgreSQL, since in theory the problem is reduced to that of finding pg_config.

This commit replaces the existing PostgreSQL detection with a more straightforward search for pg_config. The change is backwards- compatible, in the sense that --with-pgsql=foo will still find the postgres installation in "foo," except now it will do so through foo/bin/pg_config rather than foo/include and foo/lib.

This solves a rare problem: on systems where only one version of postgres is allowed, or on systems where multiple versions exist but a "default" version can be chosen, the most likely path for postgres is --with-pgsql=/usr. With the current implementation, this leads to -I/usr/include being added to CPPFLAGS, which is fine, and -L/usr/lib being added to LDFLAGS, which is not. On some of those same systems, /usr/lib is for 32-bit libraries while /usr/lib64 is the "normal" (64-bit) library path. So as a result of detecting postgres under /usr, the linker is pointed to the wrong set of libraries. This is known to crash lld (the LLVM linker).


RincewindsHat:
Interesting. I will take a look at it in time.
Thank you for the contribution :-)

PR #2020, opened 2024-09-27
Check snmp fixes

RincewindsHat:
Functionally this changes nothing, code style these changes are an improvement in so far as this way that huge bunch of global variables are marked as "only used in this translation unit".


sni (approved):

sni:
looks good to me

PR #2022, opened 2024-09-30
fix check_curl: OpenSSL SSL_read: error:0A000126:SSL routines::unexpe…

sni:
fix check_curl: OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0

using check_curl on a probably embedded device responding as 'Server: GoAhead-Webs'

    %> check_curl -H ... -S -vvv

    > GET / HTTP/1.1
    Host: ...
    User-Agent: check_curl/v2.4.0 (monitoring-plugins 2.4.0, libcurl/7.76.1 OpenSSL/3.0.7 zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.1 (+libidn2/2.3.0) libssh/0.10.4/openssl/zlib nghttp2/1.43.0)
    Accept: */*
    Connection: close

    * Mark bundle as not supporting multiuse
    * HTTP 1.0, assume close after body
    < HTTP/1.0 302 Redirect
    < Server: GoAhead-Webs
    < Date: Tue Mar 26 17:57:16 2019
    < Cache-Control: no-cache, no-store, must-revalidate,private
    < Pragma: no-cache
    < Expires: 0
    < Content-Type: text/html
    < X-Frame-Options: sameorigin
    < X-XSS-Protection: 1; mode=block
    < X-Content-Type-Options: nosniff
    < Location: https://...
    <
    * OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0
    * Closing connection 0

reading the discussion on https://github.com/openssl/openssl/discussions/22690 suggest to set the option SSL_OP_IGNORE_UNEXPECTED_EOF which makes check_curl behave like check_http at this point. Since this is a rather new flag, fencing it in ifdefs.
And since there can only be one ssl ctx function, we need to move both tasks into one function.

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -214,6 +214,7 @@ char *client_privkey = NULL;
 char *ca_cert = NULL;
 bool verify_peer_and_host = false;
 bool is_openssl_callback = false;
+bool add_sslctx_verify_fun = false;
 #if defined(HAVE_SSL) && defined(USE_OPENSSL)
 X509 *cert = NULL;
 #endif /* defined(HAVE_SSL) && defined(USE_OPENSSL) */
@@ -299,7 +300,7 @@ main (int argc, char **argv)
 
 int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 {
-	(void) preverify_ok;
+  (void) preverify_ok;
   /* TODO: we get all certificates of the chain, so which ones
    * should we test?
    * TODO: is the last certificate always the server certificate?
@@ -324,9 +325,18 @@ int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 
 CURLcode sslctxfun(CURL *curl, SSL_CTX *sslctx, void *parm)
 {
-	(void) curl; // ignore unused parameter
-	(void) parm; // ignore unused parameter
-  SSL_CTX_set_verify(sslctx, SSL_VERIFY_PEER, verify_callback);
+  (void) curl; // ignore unused parameter
+  (void) parm; // ignore unused parameter
+  if(add_sslctx_verify_fun) {
+    SSL_CTX_set_verify(sslctx, SSL_VERIFY_PEER, verify_callback);
+  }
+
+  // workaround for issue:
+  // OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0
+  // see discussion https://github.com/openssl/openssl/discussions/22690
+#ifdef SSL_OP_IGNORE_UNEXPECTED_EOF
+  SSL_CTX_set_options(sslctx, SSL_OP_IGNORE_UNEXPECTED_EOF);
+#endif
 
   return CURLE_OK;
 }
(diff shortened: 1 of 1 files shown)


sni:
Probably totally unrelated to my changes, will retry again tomorrow.

fedory rawhide is broken:
```
   File "/usr/lib64/python3.13/subprocess.py", line 1966, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/dnf-3'
+ cat /tmp/result-srcrpm/root.log /tmp/result-srcrpm/build.log
+ exit 1
```

Maybe we should handle rawhide like debian testing and keep it out of our required test suites but put them in a separate action which maybe runs once a month to make sure we get notified if futur release might break something.
But it should not prevent current development.

sni:
waiting for #2023

sni:
finally all tests are green and i tested this patch on debian11/12 and rhel7/8/9. Looked good,
except on rhel7 but the certificate check did not work there before anyway because of too old libcurl in combination with libnss.

plugins/check_curl.c
```diff
@@ -324,9 +325,18 @@ int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 
 CURLcode sslctxfun(CURL *curl, SSL_CTX *sslctx, void *parm)
 {
-	(void) curl; // ignore unused parameter
-	(void) parm; // ignore unused parameter
-  SSL_CTX_set_verify(sslctx, SSL_VERIFY_PEER, verify_callback);
+  (void) curl; // ignore unused parameter
+  (void) parm; // ignore unused parameter
+  if(add_sslctx_verify_fun) {
```
RincewindsHat, line 330: This is essentially a guard for when TLS backends do not provide verification or no TLS backend is available?

plugins/check_curl.c
```diff
@@ -324,9 +325,18 @@ int verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
...
+    SSL_CTX_set_verify(sslctx, SSL_VERIFY_PEER, verify_callback);
+  }
+
+  // workaround for issue:
+  // OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0
+  // see discussion https://github.com/openssl/openssl/discussions/22690
+#ifdef SSL_OP_IGNORE_UNEXPECTED_EOF
```
RincewindsHat, line 337: Should that error be ignored unconditionaly?

andreasbaumann: Things look fine to me, the only question I would be, how would you decide to set SSL_OP_IGNORE_UNEXPECTED_EOF at runtime? Always? Or only if the webserver is of a specific type?

sni: i'd just always ignore it. Doing this only for specific servers is probably way to much maintenance and adding a extra switch for it seems to as bit over-engineered.
And with that option in place, it behaves like check_http does as well.

plugins/check_curl.c
sni: cert verification depends on the ssl backend and curl version, so this function is not always required. Previously the callback had been added on demand for certificate verification, but there can only be a single callback. That's why i moved that outside
and use a switch to decide whether to add the verify function or not.

andreasbaumann (approved):

PR #2023, opened 2024-10-02
add new github test workflow for debian:testing and fedora:rawhide

sni:
this workflow runs scheduled once a week, so it does not prevent PRs from being merged but still allows us to notice if anything will go wrong with the next release.
Remove rawhide test from the mandatory tests like we did with debian testing already.

### DIFF
--- a/.github/workflows/test-next.yml
+++ b/.github/workflows/test-next.yml
@@ -2,6 +2,7 @@
 name: Tests Debian:Testing and Fedora:Rawhide
 
 on:
+  workflow_dispatch: {}
   push:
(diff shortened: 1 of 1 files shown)


waja (commented):
How about also adding workflow_dispatch (https://docs.github.com/de/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch) too?

sni:
good idea, added that one as well

RincewindsHat (approved):
Looks useful

PR #2026, opened 2024-10-10
Add clang-format config file

RincewindsHat:
# Small commit, huge decision

I propose to define the code style of the C part of the project in the `golang` style.
Whatever `clang-format` does with the config is the right style.
I do get annoyed everytime with mixed indentation in the code and stuff like that.

Will this totally ruin the first level of `git blame`? yes

Nonetheless it makes things better IMHO.
In addition to this change and reformatting stuff, when convenient, we should adapt all relevant documentation.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,10 @@
+BasedOnStyle: LLVM
+UseTab: ForContinuationAndIndentation
+IndentWidth: 4
+TabWidth: 4
+AllowShortIfStatementsOnASingleLine: false
+BreakBeforeBraces: Attach
+AlignConsecutiveMacros: true
+ColumnLimit: 180
+IndentPPDirectives: AfterHash
+SortIncludes: Never


sni:
adding a formater config does not change anything, or do you plan to apply the format for all files afterwards? Or only new files, or only changed files?
Breaking git blame is not so much of an issue, but it also breaks most not yet applied patches.

RincewindsHat:
I plan to apply the formatting piece by piece in the future.
Starting with parts which do not have any pending patches, so we do not get any conflicts there, but the target is to eventually apply that formatting everywhere.

This PR is mostly about making the decision, whether or not this is the "correct" style for the project.

RincewindsHat:
[Our previous coding guidelines](https://www.monitoring-plugins.org/doc/guidelines.html#AEN347) should also be changed then.
Since I did ignore or directly contradict them in the last years, I would like to change those to better communicate with potential developers.

sni:
Fine for me. Would be nice to have a make target or helper script or at least a command to apply formating to a given file.
Also i'd prefer formatting that's close to what the files are already. If that's even possible.

RincewindsHat:
well, I can write `clang-format -i $file` somewhere. I've got that in my editor.

Concerning the part of "keeping the old style" I would have to ask you to define the old style first :-)
But the config in this PR is similar to what the `CODING` file says.

sni (approved):

sni:
fine for me then

sni:
> well, I can write `clang-format -i $file` somewhere. I've got that in my editor.
> 
> Concerning the part of "keeping the old style" I would have to ask you to define the old style first :-) But the config in this PR is similar to what the `CODING` file says.

Would be nice to have that at least in the CODING file then.

PR #2028, opened 2024-10-15
Cleanup check icmp

RincewindsHat:
* apply clang-format
 * mark global variables correctly as `static`
 * update copyright


RincewindsHat:
@sni I changed the formatting a bit in dd649f629e4846313a3dc89fcae72a420f258be4 . I think it is better like this.

PR #2030, opened 2024-10-15
Cleanup/check ups

RincewindsHat:
* clang-format
 * copyright update

### DIFF
--- a/plugins/check_ups.c
+++ b/plugins/check_ups.c
@@ -6,7 +6,7 @@
  * Copyright (c) 2000 Tom Shields
  *               2004 Alain Richard <alain.richard@equation.fr>
  *               2004 Arnaud Quette <arnaud.quette@mgeups.com>
- * Copyright (c) 2002-2023 Monitoring Plugins Development Team
+ * Copyright (c) 2002-2024 Monitoring Plugins Development Team
  *
  * Description:
  *
@@ -33,7 +33,7 @@
  *****************************************************************************/
 
 const char *progname = "check_ups";
-const char *copyright = "2000-2023";
+const char *copyright = "2000-2024";
 const char *email = "devel@monitoring-plugins.org";
 
 #include "common.h"
--- a/plugins/check_ups.c
+++ b/plugins/check_ups.c
@@ -141,7 +141,6 @@ int main(int argc, char **argv) {
 		return STATE_CRITICAL;
 	}
 
-
 	if (supported_options & UPS_STATUS) {
 
 		ups_status = strdup("");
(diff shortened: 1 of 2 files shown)


plugins/check_ups.c
```diff
@@ -233,24 +230,16 @@ int main(int argc, char **argv) {
...
 				result = max_state(result, STATE_WARNING);
 			}
 			xasprintf(&data, "%s",
-					  perfdata("voltage", (long)(1000 * ups_utility_voltage),
-							   "mV", config.check_warn,
-							   (long)(1000 * config.warning_value),
-							   config.check_crit,
-							   (long)(1000 * config.critical_value), true, 0,
-							   false, 0));
+					  perfdata("voltage", (long)(1000 * ups_utility_voltage), "mV", config.check_warn, (long)(1000 * config.warning_value), config.check_crit,
```
sni, line 239: i don't think this is better than the current format.

RincewindsHat:
@sni: Play around a little bit with the options. What do you think about this one?

plugins/check_ups.c
```diff
@@ -233,24 +230,26 @@ int main(int argc, char **argv) {
...
-			} else if (config.check_warn &&
-					   ups_utility_deviation >= config.warning_value) {
+			} else if (config.check_warn && ups_utility_deviation >= config.warning_value) {
 				result = max_state(result, STATE_WARNING);
 			}
-			xasprintf(&data, "%s",
-					  perfdata("voltage", (long)(1000 * ups_utility_voltage),
-							   "mV", config.check_warn,
+			xasprintf(&data,
+					  "%s",
```
sni, line 239: looks better, except these lines with the format on a single line

plugins/check_ups.c
```diff
@@ -706,33 +723,42 @@ void print_help(void) {
...
 	printf("\n");
-	printf("%s\n", _("This plugin attempts to determine the status of a UPS "
-					 "(Uninterruptible Power"));
-	printf("%s\n", _("Supply) on a local or remote host. If the UPS is online "
-					 "or calibrating, the"));
-	printf("%s\n", _("plugin will return an OK state. If the battery is on it "
-					 "will return a WARNING"));
-	printf("%s\n", _("state. If the UPS is off or has a low battery the plugin "
-					 "will return a CRITICAL"));
+	printf("%s\n",
```
sni, line 726: and those...

sni (commented):
small complaints, but i am willing to drop them for the sake of a consistent format.

plugins/check_ups.c
RincewindsHat: hm, i will play a bit more with the binpack parameters

RincewindsHat: so, changed `BinPack`. Arguments are again more compressed. Maybe a little bit too much, but to change that the line length should be reduced I guess.
@sni: How do you like it?

sni (approved):

sni:
i'd say, so far, this is the best variant. Still not perfect...

PR #2031, opened 2024-10-25
Improve readability and reliability of test prepare scripts

RincewindsHat:
While trying to find the reason for failing test in other PRs I found the test scripts to be a bit too obscure and prone to fail in weird ways.
So I added some options to fail early in the pipeline if an error occurs in any of the steps and attempted to fix some common expansion problems.
Incindently this seems to fix the pipeline errors.

### DIFF
--- a/.github/mock.sh
+++ b/.github/mock.sh
@@ -2,6 +2,8 @@
 
 set -x
 
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


waja (approved):
Looks reasonable.

sni (approved):

PR #2037, opened 2024-10-31
Sync with the latest Gnulib code (d4ec02b3cc) (gnulib v1.0)

RincewindsHat:


RincewindsHat:
I think this should be fine, but did I forget something?

sni:
only 400 files changed? What exactly do you want me to review? :-)
Does it still build on centos 7? :-)

RincewindsHat:
fair enough, merge it is

PR #2041, opened 2024-11-06
Add dontfrag / random to fping check

### LINKED ISSUE #2024: check_by_ssh --unknown-timeout does not work as expected
I was really looking forward to the functionality promised by  --unknown-timeout (-U) implemented by #1123 and #1774.
But `./check_by_ssh -H 192.168.x.y -U -C true` still returns: "CRITICAL - Plugin timed out after 10 seconds"

I've checked the source code and found out that it returns UNKNOWN only if SSH returns 255, i.e. it has its own timeout set below the timeout of the plugin, e.g. `./check_by_ssh -H 192.168.x.y -U -oConnectTimeout=9 -C true`.   Without that, ssh times out after 3 minutes (YMMW), while the plugin usually times out much sooner, and returns CRITICAL even if -U is set.

So it would make sense, to set -oConnectTimeout to 1s less than the timeout on the plugin (-t). At least if -U is specified, but preferably always.

OTOH, there are situations, when the connection is established (at least on TCP level), and the processing of the remote command takes too long - i.e. due to high load, or DNS problems.  Those cases are not catched by -oConnectTimeout, and again result in CRITICAL, regardless of -U setting. But this may be desirable in some situations.

Please consider either
a) setting -oConnectTimeout automatically
b) (maybe separate) function to report UNKNOWN even if the remote command takes too long to process

Thank you
Vladislav Kurz

### LINKED ISSUE #2040: RFE: check_fping don't fragment setting
Hi there,

Currently, check_fping allows configuration of the ICMP packet size. It would be good to also have the ability to set the don't fragment and randomise flags so that MTU's can be tested with fping as well

Firstyear:
Fixes #2040 - this adds dontfrag and random flags to the check_fping tool.

### DIFF
--- a/plugins/check_fping.c
+++ b/plugins/check_fping.c
@@ -60,6 +60,8 @@ static int packet_count = PACKET_COUNT;
 static int target_timeout = 0;
 static int packet_interval = 0;
 static bool verbose = false;
+static bool dontfrag = false;
+static bool randomize_packet_data = false;
 static int cpl;
 static int wpl;
 static double crta;
@@ -103,6 +105,11 @@ int main(int argc, char **argv) {
 		xasprintf(&option_string, "%s-S %s ", option_string, sourceip);
 	if (sourceif)
 		xasprintf(&option_string, "%s-I %s ", option_string, sourceif);
+	if (dontfrag)
+		xasprintf(&option_string, "%s-M ", option_string);
+	if (randomize_packet_data)
+		xasprintf(&option_string, "%s-R ", option_string);
+
 
 #ifdef PATH_TO_FPING6
 	if (address_family != AF_INET && is_inet6_addr(server))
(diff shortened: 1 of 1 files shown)


plugins/check_fping.c
```diff
@@ -60,6 +60,8 @@ static int packet_count = PACKET_COUNT;
 static int target_timeout = 0;
 static int packet_interval = 0;
 static bool verbose = false;
+static bool dontfrag = false;
+static bool random = false;
```
RincewindsHat, line 64: Please rename that symbol to avoid a conflict with `random(3)`. I would suggest `randomize_packet_data`

RincewindsHat (changes requested):
Thank you for the contribution. Looks like I can't hurt, I would like to merge it in principle.
But right now, there is at least one problem I would like you to fix.

plugins/check_fping.c
Firstyear: No problem, I've just pushed the update to the PR :)

RincewindsHat (approved):
Looks good to me. Thank you :-)

Issue #1954, opened 2023-10-21
plugins exporting their local symbols


andreasbaumann:
`nm check_curl | grep print`:

```
0000000000005580 T print_help
000000000000b470 T print_revision
000000000000d500 T print_thresholds
0000000000005430 T print_usage
```

Is there a particular reason why we use this style:

```
void print_help (void);
...
void
print_help (void)
{
...
```

instead of:

```
static print_help (void);
...
void
print_help (void)
{
...
```

IMHO it just exports a whole bunch of local symbols into the binary.

RincewindsHat:
Can't say that there is any other reason other than nobody care until just now :-)

RincewindsHat:
After thinking about that for a minute, another reason might be, that we have some unit tests in c which probably need to link against some of those symbols, but I might be wrong

RincewindsHat:
reference #2020 #2021

RincewindsHat:
Apart from uselessly exposing the symbols, _not_ declaring those symbols as static prevents a compiler from detecting unused ones.
IMHO, this is actually a problem, although one with small consequences ("only" wasting dev time)

RincewindsHat:
should mostly be fixed by now.

RincewindsHat:
should be done

PR #2043, opened 2024-11-08
Cleanup/leftovers

RincewindsHat:
Mostly declaring more file local variables `static` plus some subsequent fixes.


RincewindsHat:
Should be all symbols in the C based plugins with this.
Therefore this closes #1954

PR #2056, opened 2024-12-27
check_http: fix documentation for --state-regex

andreasbaumann:


RincewindsHat (approved):
I do wonder how that got through :-)

PR #2058, opened 2024-12-31
Update configure.ac to Autoconf 2.71 syntax

weiss:
Apply all changes suggested by the `autoupdate(1)` tool, except for actually requiring a newer Autoconf version (some of our CI images don't offer 2.71 yet).


RincewindsHat (approved):
I think I started that at some point and didn't dare to go further :-)

But, looks fine I guess.

I anyone knows an autoconf wizard, some helpful comments would be appreciated.

PR #2063, opened 2025-02-17
check_snmp: Make snmpgetnext optional

RincewindsHat:
Alternative to #2062.
This variant removes the `--next` flag for `check_snmp` if `snmpgetnext` can not
be found.


RincewindsHat:
closed, for info see the discussion in #2063

PR #2065, opened 2025-02-19
Opttest tool

RincewindsHat:
Did anybody ever use or at least notice the `opttest.pl` in the main directory?
I did and moved it to `tools`.
Also fixed it and two plugins with diverging exit codes.


RincewindsHat:
@waja That one or a similar one might be worth adding to the tests to ensure, that all plugins comply to our idea about how they should behave

sni:
i never used it and learned about it for the first time now. Might be useful to include this in the automated tests then?

PR #2068, opened 2025-02-24
Check mysql replica check

### LINKED ISSUE #2060: check_mysql supports replica check
As of now the check_mysql supports a "slave check" (-S, --check-slave). This seems to utilize the SHOW SLAVE STATUS command.

Since MySQL 8.0 the "show slave status" command is deprecated and an alias for new "show replica status". Since MySQL 8.4 the "show slave status" has been removed, see [Server and Status Variables and Options Added, Deprecated, or Removed in MySQL 8.4 since 8.0](https://dev.mysql.com/doc/refman/8.4/en/added-deprecated-removed.html).

Now the plugin returns with 8.4 the following error:

> slave query error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'slave status' at line 1

Most likely referenced here: [check_mysql.c#L215](https://github.com/monitoring-plugins/monitoring-plugins/blob/e23325f7c3b34d8950cf27e9ab23af362d0b341b/plugins/check_mysql.c#L215)

RincewindsHat:
fixes #2060

This patchset contains two parts.

 + Replaces 'slave' (almost) everywhere with 'replica' since this seems to be the future naming
   convention for MySQL (and MariaDB too).
   Another cli option `--check-replica` gets introduced and `check_slave` marked as deprecated
   (although it will still work for now)
 + Modifies the replica query in a way where it queries the server version first and depending 
   on the result queries for replicas or for "slaves" if the version is older and does not
   provide that syntax


RincewindsHat:
@seb4web This should fix your problem.

abaddononion:
I'm trying to figure out how to get ahold of this patch... we're attempting to run MySQL 8.4 on an Alma 9 server and encountering this issue. Has there been a release containing this patch yet? I could also build my own nagios-plugins RPMs from scratch if I need... but I wasn't sure how to make sure I'm pulling in this changeset. Any advice?

RincewindsHat:
Hi @abaddononion,
first things first: This project contains the "Monitoring Plugins" of which "nagios-plugins" is a fork. RHEL ships "nagios-plugins", debianish OSes and SLES "monitoring-plugins".

Currently it might be difficult to cherry-pick this patch, since I am refactoring a lot of stuff and there are a lot of intermixed changes. You could use the `master` branch which should work in theory but it is not release ready and I am not sure whether I can recommend to run it in production.

RincewindsHat:
@waja https://github.com/monitoring-plugins/monitoring-plugins/tree/maintenance-2.4.0-check_mysql might be what you need.

waja:
I've created https://salsa.debian.org/nagios-team/monitoring-plugins/-/commit/dd1efc8f1be4d98463c567589a657a0e22fb06d4 for this, many thanks!

RincewindsHat:
well, should have tested that against a real MySQL(tm) server. That also needs repair in the `master` branch, so thanks to the person reporting this.
Anyway, I pushed another commit to that branch, should fix it.

waja:
> Anyway, I pushed another commit to that branch, should fix it.

The latest patchset seems to result into:

```
jon@SERVER:~$ /usr/lib/nagios/plugins/check_mysql -u REDACTED -p REDACTED -S
Slave status unavailable

Server version is 8.4.6 MySQL Community Server - GPL
```

waja:
> > Anyway, I pushed another commit to that branch, should fix it.
> 
> The latest patchset seems to result into:
> 
> ```
> jon@SERVER:~$ /usr/lib/nagios/plugins/check_mysql -u REDACTED -p REDACTED -S
> Slave status unavailable
> 
> Server version is 8.4.6 MySQL Community Server - GPL
> ```

@RincewindsHat it looks like the issue isn't solved in 2.4.0 yet and I also don't know, if that is working in master. We don't have a test against a mysql cluster instance in our test pipelines. :)

PR #2098, opened 2025-03-12
Remove check_overcr

RincewindsHat:
This commit removes the plugin check_overcr and all related files and parts of files.

OverCR seems to be dead and I have never seen a mention of an active usage anywhere.
The website still does exist, but the activity seems to be limited.
Carrying check_overcr forward seems to be a burden without a purpose and more of an opportunity to remove some code.


RincewindsHat:
@sni  @waja This might be a controversial one.

waja:
https://packages.debian.org/search?searchon=contents&keywords=check_overcr&mode=path&suite=stable&arch=any indicates that it's shipped with Debian. Anyway, I've never noted about it when working on the package or seen something on the bugtracker.
So in the end, your technical arguments are still valid and I think there is no argument against it as far as I can judge it by using my smartphone for checking. :)

sni:
tbh, i just heard about check_overcr for the first time ever. So i have no hard feelings removing it.

RincewindsHat:
Thanks for the quick feedback. I only learned about overcr by searching what the plugin might be good for :-)

For documentation purposes: http://molitor.org/overcr/

PR #2100, opened 2025-03-12
Remove check_nwstat

RincewindsHat:
check_nwstat is a plugin which was used to determine the health of things on Novel machines.

Since Novel is quite dead (even more so the product, this can be removed and this commit does just that to reduce ressource usage.

@sni: @waja: not the same thing as OverCR, but do you know any current Novel installations?


waja:
I've no real clue about it's actual relevance, but it's more common then overcr and I've heard about it. ;)
A short search at https://www.google.com/search?q=%22check_nwstat%22+debian+bts doesn't reveal much.

PR #2107, opened 2025-03-12
Refactor/check users

RincewindsHat:


RincewindsHat:
Needs a little testing on other OSes than linux

RincewindsHat:
works on freeBSD in one test :-)

PR #2112, opened 2025-03-17
change error message for missing certificate

klaernie:
The old error message is quite similar to the openssl `failed to retrieve issuer certificate` and can mislead users to troubleshooting certificate stores.

The new message should be distinct enough to make it clear to users that this is not a problem raised by the underlying SSL implementation, but a problem inside monitoring-plugins.

This change is triggered by a problem I have in our infrastructure that leads to missing certificates, where I mistook the error message as coming from openssl, which turned out to be the wrong direction. Then again today I went down the same rabbit hole until I remembered that the problem was outside the openssl certificate stores.
Hopefully I will manage to build a reproducible problem to demonstrate the actual problem that caused me to see this error message, or maybe I'll even get it fixed myself.


plugins/sslutils.c
```diff
@@ -201,7 +201,7 @@ int np_net_ssl_check_certificate(X509 *certificate, int days_till_exp_warn, int
 	time_t tm_t;
 
 	if (!certificate) {
-		printf("%s\n", _("CRITICAL - Cannot retrieve server certificate."));
+		printf("%s\n", _("CRITICAL - Not server certificate present to inspect."));
```
RincewindsHat, line 204: Would "No server certificate" (on `t` less) be better?

klaernie: good point! Sorry, totally made the commit without reading it over at least once.

RincewindsHat (approved):
Thanks, that sounds reasonable.

PR #2115, opened 2025-03-27
Improve handling of -4/-6

Firstyear:
If fping is used with a target that has dual stack v4/v6, then due to
    the logic during command construction, ipv4 will never be checked as v6
    is preferred by fping.

    This explicitly flags -4/-6 when it is requested by the user.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1524,17 +1524,10 @@ AC_PATH_PROG(PATH_TO_FPING6,fping6)
 AC_ARG_WITH(fping_command,
             ACX_HELP_STRING([--with-fping-command=PATH],
                             [Path to fping command]), PATH_TO_FPING=$withval)
-AC_ARG_WITH(fping6_command,
-            ACX_HELP_STRING([--with-fping6-command=PATH],
-                            [Path to fping6 command]), PATH_TO_FPING6=$withval)
-
 if test -n "$PATH_TO_FPING"
 then
 	AC_DEFINE_UNQUOTED(PATH_TO_FPING,"$PATH_TO_FPING",[path to fping])
 	EXTRAS="$EXTRAS check_fping\$(EXEEXT)"
-	if test x"$with_ipv6" != xno && test -n "$PATH_TO_FPING6"; then
-		AC_DEFINE_UNQUOTED(PATH_TO_FPING6,"$PATH_TO_FPING6",[path to fping6])
-	fi
 else
 	AC_MSG_WARN([Get fping from http://www.fping.com in order to make check_fping plugin])
 fi
--- a/plugins/check_fping.c
+++ b/plugins/check_fping.c
@@ -79,7 +79,29 @@ int main(int argc, char **argv) {
 	server = strscpy(server, config.server_name);
 
 	char *option_string = "";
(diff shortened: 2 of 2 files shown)


Firstyear:
No problem, will fix that shortly :)

Firstyear:
@RincewindsHat No problems, I have fixed the issue and commented the logic a bit better now.

RincewindsHat:
Sorry for the delay, I did not forget this, but there are too many things to do ...

Anyway, I did some reading and at least current versions of `fping` (5.1 on my machine) are deprecating the `fping6` symlink due there being no good reasons for it's existance anymore, since there are explixit `-4` and `-6` flags available (the sames seems to be case for `ping`, at least on a Debian Linux).
That brings me to the thought where the whole `PATH_TO_FPING6` should be dropped for simplicities sake.

Additionaly we had a small discussion in the IRC about whether IPv6 is optional anymore and the result (from my side at least) is, that IPv6 is not optional anymore and we can remove all occurences of `USE_IPV6` and go on with dual stack code everywhere.

This is probably a lot to ask of you, since this changes everything here, but would that be something you would like to adapt?
Or (since you care a bit at least :-) ) do you disaggree with the points above, which I would happily discuss with you.

Firstyear:
I actually agree with all these points - I didn't want to overstep and just "remove things" without checking first.

So I'm happy to remove both the `USE_IPv6` and the use of the fping6 alias. I'll update the PR shortly.

Firstyear:
Looks like this fails on your CI as debian:stable lacks "is_inet6_addr" (see plugins/netutils.h line 83). What did you want me to do here about that one?

RincewindsHat:
IPv6 was disabled statically in the CI. I reenabled it in #2123. Could you update your branch to the current master?

Firstyear:
No problem, updated.

RincewindsHat:
Nice, thank you.
Now, I have a question about the logic:
As it is, `-4`/`-6` is always added to the `fping` arguments, there is no way to not set and let fping figure out what is appropriate.
I am not sure if there is not a scenario in which this might be problem.

Would it be better, if we get `AF_UNSPEC` (user did not select anything) to just let fping handle it without setting `-4` or `-6`?

Firstyear:
The logic goes:

```
	 * If the user requested -6 OR the user made no assertion and the address is v6 or dualstack
	 *   -> we use ipv6
	 * If the user requested -4 OR the user made no assertion and the address is v4 ONLY
	 *   -> we use ipv4
```

So if the user make *no assertion* about the v4 or v6, then *if* the address is v6 OR dualstack -> v6, if the user made no assertion and the address is v4 *only* then we use v4 as a flag. This is to ensure that how we resolve the address matches what fping will do - for example, we resolve the address as v6 and then fping attempts v4 (because we passed a hostname, not an ip). 

The goal was to make this just a bit more deterministic and clear about what we are actually testing, rather than leaving it as "undefined".

RincewindsHat (approved):
I see, maybe I had a knot in my thoughts before.
Anyway, thank you for your work and for adapting to my wishes so quickly :-)

Firstyear:
No problem at all! Happy to help :)

PR #2117, opened 2025-04-14
Add -u option to output age in hours or days (vs seconds) in check_file_age

sigio:
This patch adds a -u option to check_file_age.

This can be used to change the units in the human-readable output of the plugin to report file age in either hours (h) or days (d).

old:
FILE_AGE WARNING: /var/www/nextcloud_v30.0.5 is 5493635 seconds old and 4096 bytes | age=5493635s;5184000;8640000 size=4096B;0;0;0

new: (with -u d arguments)
FILE_AGE WARNING: /var/www/nextcloud_v30.0.5 is 63 days old and 4096 bytes | age=5493021s;5184000;8640000 size=4096B;0;0;0

### DIFF
--- a/plugins-scripts/check_file_age.pl
+++ b/plugins-scripts/check_file_age.pl
@@ -32,8 +32,9 @@ use utils qw (%ERRORS &print_revision &support);
 sub print_help ();
 sub print_usage ();
 
-my ($opt_c, $opt_f, $opt_w, $opt_C, $opt_W, $opt_h, $opt_V, $opt_i);
+my ($opt_c, $opt_f, $opt_w, $opt_C, $opt_W, $opt_h, $opt_u, $opt_V, $opt_i);
 my ($result, $message, $age, $size, $st, $perfdata);
+my ($units, $udiv, $uage );
 
 $PROGNAME="check_file_age";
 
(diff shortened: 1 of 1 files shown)


nicktencate (approved):

RincewindsHat:
That sounds like a nice addition, but it still requires the operator to know the probable range of values in advance.
Would it be a nicer solution to "humanize" that value automatically? Meaning, to convert it to a appropriate and easily understandable time unit?

sigio:
> That sounds like a nice addition, but it still requires the operator to know the probable range of values in advance. Would it be a nicer solution to "humanize" that value automatically? Meaning, to convert it to a appropriate and easily understandable time unit?

That would be nice, but in my usecase I only needed 'days' basically, we use it to check how old an installed application is, and warn when this is a few months old. Here output in seconds is completely useless, and anything over days isn't really needed. I'd welcome someone writing it, but I haven't written much perl the last 15 years anymore.

RincewindsHat:
Hm, @sni any opinions (since it's the Perl part)?

For my part, and this makes me feel ungrateful, I tend to reject this suggestion. Additional flags means more complexity and more maintenance burden (not that much in this case, but still). The gain is rather limited to this specific use case and I think the other path with autoformated time intervals provides more advantages and does not increase the user interface.

I would be happy if you could unbury your Perl wisdom and go for the other solution though :-)

PR #2123, opened 2025-05-08
Do not explicitely disable IPv6 in the CI

RincewindsHat:
Previously IPv6 functionality was explicitely disabled in the Github CI,
although it is not clear to me why that was the case,
since IPv6 was not used there but there is no good reason
to not compile the functionality at all.

This should fix some of the problems with #2115

### DIFF
--- a/.github/workflows/test-next.yml
+++ b/.github/workflows/test-next.yml
@@ -38,7 +38,7 @@ jobs:
               ${{ matrix.distro }} \
               /bin/sh -c '${{ matrix.prepare }} && \
                 tools/setup && \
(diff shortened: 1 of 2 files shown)


waja:
> although it is not clear to me why that was the case,

It was disabled because of infrastructure issues at GitHub.

RincewindsHat:
Thanks @waja, that makes sense. Was it necessary though to deactivate compilation of IPv6 support completely?

Anyways, the tests seem to run properly. I guess we can continue with these changes then?

waja:
I don't remember for the exact reason anymore.

PR #2124, opened 2025-05-09
changed filename in cmd_file_read to const char * (check_apt warning)

andreasbaumann:
The warning I got on gcc 15.1.1 (Arch) was:

```
check_apt.c: In function ‘run_upgrade’:
check_apt.c:319:50: warning: passing argument 1 of ‘cmd_file_read’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  319 |                 result.errorcode = cmd_file_read(input_filename, &chld_out, 0);
      |                                                  ^~~~~~~~~~~~~~
In file included from runcmd.h:28,
                 from check_apt.c:38:
../lib/utils_cmd.h:23:19: note: expected ‘char *’ but argument is of type ‘const char *’
   23 | int cmd_file_read(char *, output *, int);
      |                   ^~~~~~
```

The man page of glibc states than `int open(const char *path, int oflag, ...);` is the
prottype. `cmd_file_read` uses open here `if ((fd = open(filename, O_RDONLY)) == -1) `

### DIFF
--- a/lib/utils_cmd.c
+++ b/lib/utils_cmd.c
@@ -346,7 +346,7 @@ int cmd_run_array(char *const *argv, output *out, output *err, int flags) {
 	return _cmd_close(fd);
 }
 
-int cmd_file_read(char *filename, output *out, int flags) {
+int cmd_file_read(const char *filename, output *out, int flags) {
 	int fd;
 	if (out)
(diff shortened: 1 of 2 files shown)


RincewindsHat (approved):
Good catch. Thank you :-)

PR #2131, opened 2025-06-25
check_curl: fix sticky redirection with ipv6 address

mjeanson:
When running the check on an ipv6 address with a sticky onredirect policy like in this example:

  check_curl -6 -H example.com -I ::1 -f sticky

It results in a getaddrinfo error:

  HTTP CRITICAL - Unable to lookup IP address for '[::1]': getaddrinfo returned -3 - Temporary failure in name resolution

This happens because in check_http() if the content of server_addr is an ipv6 address enclosing brackets are added and on redirection a subsequent call to check_http() will pass this now bracketed value to getaddrinfo resulting in the error.

To work around this, keep a pointer to the original server_addr value and use it when a non-bracketed address is expected.


RincewindsHat:
Hi @mjeanson,
thanks for the contribution. I will take a look at it when time and energy allows it.
In the meantime, could you take a look at what causes the tests to fail?

mjeanson:
It should pass the tests now.

PR #2133, opened 2025-07-11
Fix check ssh buffer overflow

rlaager:
A buffer overflow was occurring when the server responded with:
Exceeded MaxStartups\r\n

glibc would then abort() with the following output:
*** buffer overflow detected ***: terminated

It was the memset() that was overflowing the buffer.  But the memmove()
needed fixing too.

First off, there was an off-by-one error in both the memmove() and
memset().  byte_offset was already set to the start of the data _past_
the newline (i.e. len + 1).  For the memmove(), incrementing that by 1
again lost the first character of the additional output.  For the
memset(), this causes a buffer overflow.

Second, the memset() has multiple issues.  The comment claims that it
was NULing (sic "null") the "rest".  However, it has no idea how long
the "rest" is, at this point.  It was NULing BUFF_SZ - byte_offset + 1.
After fixing the off-by-one / buffer overflow, it would be NULing
BUFF_SZ - byte_offset.  But that doesn't make any sense.  The length of
the first line has no relation to the length of the second line.

For a quick-and-dirty test, add something like this just inside the
while loop:
memcpy(output,
  "Exceeded MaxStartups\r\nnext blah1 blah2 blah3 blah4\0",
  sizeof("Exceeded MaxStartups\r\nnext blah1 blah2 blah3 blah4\0"));

And, after the memmove(), add:
  printf("output='%s'\n", output);

If you fix the memset() buffer overflow, it will output:
output='ext blah1 blah2 blah3 '

As you can see, the first character is lost.

If you then fix the memmove(), it will output:
output='next blah1 blah2 blah3'

Note that this is still losing the "blah 4".

After moving the memset() after byte_offset is set to the new strlen()
of output, then it works correctly:
output='next blah1 blah2 blah3 blah4'

I also fixed a variable type while I was there. This is a second commit.


RincewindsHat:
Hi @rlaager,
Thanks, that looks reasonable on a first glance. I will take a look into it in time.

PR #2139, opened 2025-08-01
(Re)construct PLATFORM_ID as it's droped since Fedora 43

waja:
See https://fedoraproject.org/wiki/Changes/Drop_PLATFORM_ID?#Drop_PLATFORM_ID


RincewindsHat:
Fine for me if it works :-)
But what do we need it for (if Fedora does not?)

waja:
> But what do we need it for (if Fedora does not?)

It's used in [.github/mock.sh](https://github.com/monitoring-plugins/monitoring-plugins/blob/master/.github/mock.sh).

PR #2151, opened 2025-09-15
check_by_ssh: Ignore output on stderr by default

### LINKED ISSUE #2147: check_by_ssh: OpenSSH 10.1 Post-Quantum Cryptography Warning Breakage
Starting with the upcoming version 10.1, OpenSSH will report big warnings if the opposite server does not support post-quantum cryptography, <https://www.openssh.com/pq.html>.

Yesterday, I got a report from someone using current OpenSSH w/ `check_by_ssh` and Icinga. Their monitored server did not support PQ, so `ssh(1)` printed the warning text on stderr, but `check_by_ssh` only reported the first line and exits with UNKNOWN.

---

Personally also using an unstable OpenSSH version on my monitoring server, I was able to reproduce this with a server not supporting any PQ KexAlgorithms.

Using `ssh(1)` shows the mentioned warning on stderr (`2>/dev/null` hides it), but executes the  command and exits with `0`.

```
$ ssh icinga@example.com '/usr/lib/nagios/plugins/check_disk -c 10% -w 20% /'
** WARNING: connection is not using a post-quantum key exchange algorithm.
** This session may be vulnerable to "store now, decrypt later" attacks.
** The server may need to be upgraded. See https://openssh.com/pq.html
DISK OK - free space: / 7183MiB (54% inode=91%);| /=6355419136B;11723079680;13188464640;0;14653849600

$ echo $?
0
```

When using `check_by_ssh`, only the first line of the warning text from stderr is shown and the commands exits with `3`. However, the actual command was executed, but it's output was completely ignored due to the presence of something on stderr.

https://github.com/monitoring-plugins/monitoring-plugins/blob/f69aba3129d672b88f5bae187fb971fc0ac228ed/plugins/check_by_ssh.c#L113-L127

```
$ /usr/local/libexec/nagios/check_by_ssh -C '/usr/lib/nagios/plugins/check_disk -c 10% -w 20% /' -H example.com -l icinga
Remote command execution failed: ** WARNING: connection is not using a post-quantum key exchange algorithm.

$ echo $?
3
```

---

As stated in the linked "OpenSSH Post-Quantum Cryptography" document, the new `ssh_config(5)` option [`WarnWeakCrypto`](https://man.openbsd.org/ssh_config.5#WarnWeakCrypto) was introduced. Passing `-o WarnWeakCrypto=no` to `check_by_ssh` mitigates this issue.

```
$ /usr/local/libexec/nagios/check_by_ssh -C '/usr/lib/nagios/plugins/check_disk -c 10% -w 20% /' -H example.com -l icinga -o WarnWeakCrypto=no
DISK OK - free space: / 7183MiB (54% inode=91%);| /=6355419136B;11723079680;13188464640;0;14653849600

$ echo $?
0
```

FTR, it is also possible to mitigate this via `check_by_ssh`'s `--skip-stderr` option.

---

Since OpenSSH 10.1 will be released soon and will slowly be shipped to downstream distributions, this will lead to issues within the next months if the monitoring server is up to date, but has to monitor more ancient machines.

oxzi:
check_by_ssh no longer returns UNKNOWN if ssh(1) returns data on stderr. But it can be enforced again by the new "--unknown-on-stderr" option.

---

The default logic of check_by_ssh results in an UNKNOWN state if the ssh(1) process produces output on stderr. Using the "--skip-stderr=[n]" option allows ignoring a certain amount of lines or disabling this check altogether. Furthermore, passing the "--warn-on-stderr" option reduces the exit code to WARNING.

The "--help" output does not document this behavior, only states that "--warn-on-stderr" will result in the WARNING, but does not mention the UNKNOWN by default.

The man page of ssh(1) mentions that debug information is logged to stderr. This conflicts with the described logic, resulting in check_by_ssh to go UNKNOWN, unless additional options are set.

Starting with OpenSSH version 10.1, ssh(1) will report warnings to stderr if the opposite server does not support post-quantum cryptography, <https://www.openssh.com/pq.html>.

This change, slowly being rolled out throughout the next months/years, might result in mass-breakages of check_by_ssh.

By introducing a new "--unknown-on-stderr" option, enforcing the prior default logic of an UNKNOWN state for data on stderr, and ignoring output on stderr by default, check_by_ssh will continue to work. One might even argue that this change converges actual implementation and the documented behavior, as argued above.

---

$ ssh example '/usr/lib/nagios/plugins/check_dummy 0 demo' ** WARNING: connection is not using a post-quantum key exchange algorithm. ** This session may be vulnerable to "store now, decrypt later" attacks. ** The server may need to be upgraded. See https://openssh.com/pq.html OK: demo

$ echo $?
0

$ ./check_by_ssh -H example -C '/usr/lib/nagios/plugins/check_dummy 0 demo' OK: demo

$ echo $?
0

$ ./check_by_ssh -H example -C '/usr/lib/nagios/plugins/check_dummy 0 demo' --warn-on-stderr Remote command execution failed: ** WARNING: connection is not using a post-quantum key exchange algorithm.

$ echo $?
1

$ ./check_by_ssh -H example -C '/usr/lib/nagios/plugins/check_dummy 0 demo' --unknown-on-stderr Remote command execution failed: ** WARNING: connection is not using a post-quantum key exchange algorithm.

$ echo $?
3

---

Fixes #2147.

### DIFF
--- a/plugins/check_by_ssh.c
+++ b/plugins/check_by_ssh.c
@@ -119,13 +119,14 @@ int main(int argc, char **argv) {
 		skip_stderr = config.skip_stderr;
 	}
 
-	/* UNKNOWN or worse if (non-skipped) output found on stderr */
-	if (chld_err.lines > (size_t)skip_stderr) {
+	/* Allow UNKNOWN or WARNING state for (non-skipped) output found on stderr */
+	if (chld_err.lines > (size_t)skip_stderr && (config.unknown_on_stderr || config.warn_on_stderr)) {
 		printf(_("Remote command execution failed: %s\n"), chld_err.line[skip_stderr]);
-		if (config.warn_on_stderr) {
+		if (config.unknown_on_stderr) {
+			return max_state_alt(result, STATE_UNKNOWN);
+		} else if (config.warn_on_stderr) {
 			return max_state_alt(result, STATE_WARNING);
 		}
-		return max_state_alt(result, STATE_UNKNOWN);
 	}
 
 	/* this is simple if we're not supposed to be passive.
@@ -191,12 +192,13 @@ check_by_ssh_config_wrapper process_arguments(int argc, char **argv) {
 		{"name", required_argument, 0, 'n'},
 		{"services", required_argument, 0, 's'},
 		{"identity", required_argument, 0, 'i'},
-		{"user", required_argument, 0, 'u'},
+		{"user", required_argument, 0, 'u'}, /* backwards compatibility */
 		{"logname", required_argument, 0, 'l'},
 		{"command", required_argument, 0, 'C'},
 		{"skip", optional_argument, 0, 'S'}, /* backwards compatibility */
 		{"skip-stdout", optional_argument, 0, 'S'},
 		{"skip-stderr", optional_argument, 0, 'E'},
+		{"unknown-on-stderr", no_argument, 0, 'e'},
 		{"warn-on-stderr", no_argument, 0, 'W'},
 		{"proto1", no_argument, 0, '1'},
 		{"proto2", no_argument, 0, '2'},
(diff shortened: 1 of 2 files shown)


RincewindsHat (approved):
Like it. Thank you :-)

Issue #2152, opened 2025-09-15
Refactor/check users broke OpenBSD


oxzi:
```
./plugins $ make check_users
  CC       check_users.o
  CC       check_users.d/users.o
In file included from check_users.d/users.c:116:
check_users.d/../popen.h:6:1: error: unknown type name 'FILE'
    6 | FILE *spopen(const char *);
      | ^
check_users.d/../popen.h:7:13: error: unknown type name 'FILE'
    7 | int spclose(FILE *);
      |             ^
check_users.d/../popen.h:10:1: error: unknown type name 'pid_t'
   10 | pid_t *childpid = NULL;
      | ^
check_users.d/../popen.h:12:1: error: unknown type name 'FILE'
   12 | FILE *child_process = NULL;
      | ^
check_users.d/../popen.h:13:1: error: unknown type name 'FILE'
   13 | FILE *child_stderr = NULL;
      | ^
5 errors generated.
*** Error 1 in [...]/monitoring-plugins/plugins (Makefile:2869 'check_users.d/users.o': @echo "  CC ...)
```

After adding the required imports.

```diff
diff --git a/plugins/popen.h b/plugins/popen.h
index e318ce25..64a94294 100644
--- a/plugins/popen.h
+++ b/plugins/popen.h
@@ -3,6 +3,8 @@
  *
  *****************************************************************************/
 
+#include "common.h"
+
 FILE *spopen(const char *);
 int spclose(FILE *);
 void popen_timeout_alarm_handler(int);
```

```
./plugins $ make check_users
  CC       check_users.o
  CC       check_users.d/users.o
  CCLD     check_users
ld: error: duplicate symbol: child_process
>>> defined at popen.h:14 (./popen.h:14)
>>>            check_users.o:(child_process)
>>> defined at popen.h:14 (check_users.d/../popen.h:14)
>>>            check_users.d/users.o:(.bss+0x10)

ld: error: duplicate symbol: child_stderr_array
>>> defined at popen.h:13 (./popen.h:13)
>>>            check_users.o:(child_stderr_array)
>>> defined at popen.h:13 (check_users.d/../popen.h:13)
>>>            check_users.d/users.o:(.bss+0x8)

ld: error: duplicate symbol: child_stderr
>>> defined at popen.h:15 (./popen.h:15)
>>>            check_users.o:(child_stderr)
>>> defined at popen.h:15 (check_users.d/../popen.h:15)
[...]

PR #2157, opened 2025-09-26
fixes check_users build on OpenBSD (without utmpx)

sthen:
This fixes building check_users on OpenBSD (ref #2152)


RincewindsHat (approved):
Thank you! Since I don't have an OpenBSD around I trust you that this works :-)

PR #2161, opened 2025-09-27
plugins: move duplicate input validation to shared utility

UiP9AV6Y:
both check_curl and check_http have implemented the same validation function. this change relocates the implementation to a shared location to be used by both.

### DIFF
--- a/plugins/Makefile.am
+++ b/plugins/Makefile.am
@@ -105,8 +105,9 @@ noinst_PROGRAMS = @EXTRA_PLUGIN_TESTS@
 # These two lines support "make check", but we use "make test"
 check_PROGRAMS = @EXTRA_PLUGIN_TESTS@
 
-libnpcommon_a_SOURCES = utils.c netutils.c sslutils.c runcmd.c	\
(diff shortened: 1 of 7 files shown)


UiP9AV6Y:
i intend add support for certificate authentication (`-J` and `-K`) to `check_tcp` and `check_smtp` in the near future where i intend to use the same validation code again; this code is merely in preparation for that. sorry for not stating my full intentions with this MR in the first place.

andreasbaumann:
Ah, that's different then. :-)

RincewindsHat:
I guess I should have responded here earlier. Sorry for that.

I don't want to touch `check_http` or create any connections from it to other code. I consider it to be dead and don't want put any work into it or create later work by somehow connecting it closer to other code.

That being said, I am totally for shared functions, if they make sense otherwise.
As for this concrete PR I would reject it on the basis of not wanting to touch `check_http` in any meaningful way.

But feel free to combine the changes here with the changes to `check_tcp` and `check_smtp` if that makes more sense.

Issue #1097, opened 2013-09-24
check_dig: option to require flag is set [sf#3569503]


monitoring-user:
Submitted by ulrik_haugen on 2012-09-19 15:45:48

I'd like an option to check_dig to require that a flag is set in the reply. This would be of use when checking signed records to insist that the ad flag is set in the reply. While you can just use a validating resolver and expect check_dig to complain about answer missing this gives no warning if the resolver should stop validating for any reason.

PR #2165, opened 2025-10-07
check_dig: add -E/--require-flags and -X/--forbid-flags

Decstasy:
Hello maintainers,

this PR adds two optional switches to validate DNS header flags from `dig` output:

* `-E, --require-flags=LIST` — flags that must be present (e.g. `aa,qr`)
* `-X, --forbid-flags=LIST` — flags that must not be present (e.g. `rd`)

Includes small helpers to parse the `flags:` line and CSV lists. Help/usage updated. No behavior change if the options are not used.

### Why

Operational/SecOps cases often need to assert `aa` on authoritative servers or ensure `rd` isn’t set. So I guess we need such a feature.

### Usage

```bash
# require aa and qr
check_dig -H <dns> -l <name> -E aa,qr

# forbid rd (e.g., with +norecurse it should be absent)
check_dig -H <dns> -l <name> -X rd -A "+norecurse"
```

### Testing

* Manually tested against internal authoritative servers with various `dig` options (`+norecurse`, time/retry).
* Two colleagues tested independently; no issues found.
* Note: presence of `rd` depends on your `+norecurse`/recursion settings, as expected.

### Review

I don’t write C here every day, so please take a close look, especially around memory handling (e.g., `realloc`/`strdup`). I’m happy to adjust style/i18n/helper placement, or switch to aggregated error messages for multiple flags if preferred.

Related / closes monitoring-plugins/monitoring-plugins#1097

Thanks!
Dennis

### DIFF
--- a/plugins/check_dig.c
+++ b/plugins/check_dig.c
@@ -36,6 +36,7 @@ const char *progname = "check_dig";
 const char *copyright = "2002-2024";
 const char *email = "devel@monitoring-plugins.org";
 
+#include <ctype.h>
 #include "common.h"
 #include "netutils.h"
 #include "utils.h"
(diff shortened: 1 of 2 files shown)


Decstasy:
Hey @RincewindsHat,
thank you for your review and the valuable findings. The pointers were indeed a bit messy.

A few energy drinks later, it’s done. I refactored all the helpers and the corresponding calls in main. I also added inline comments and proper function documentation.
I tested everything as thoroughly as I could, and I believe the code is in good shape now. Structs really make everything easier - it’s been a while since I wrote C, so please have a good look at my changes :)

Check out the commit message for more details, and feel free to adjust smaller things if you think it’s necessary.

RincewindsHat (approved):

RincewindsHat:
Thanks again for the contribution and the cooperation

PR #2172, opened 2025-11-05
check_ntp_time: add cli option for output format

RincewindsHat:


RincewindsHat:
forgot that previously

PR #2194, opened 2025-12-02
Remove check_ntp

RincewindsHat:
check_ntp was marked as deprecated for forever (SVN times). In the spirit of cleaning up and removing the dead bodies, this commit removes check_ntp and all its connections.


RincewindsHat:
@waja  @sni  I opened this to have a paper trail and a room for discussion.
The deprecation note has been there forever and it seems like nobody knows anymore how and why it is there.
This confuses me a lot (and probably this is true for other people).
I would like to avoid shipping faulty software and don't want to invest ressources (my time, computing ressources for compilation and testing, space for storage) in a dead horse.
That's why I propose to remove it altogehter close this chapter.

Issue #2192, opened 2025-12-01
Check_curl onredirect=follow strips the URL parameters, while check_http / check_curl with onredirect=curl does not.


inqrphl:
Tested on a local build with the latest commit: fd42290d4a0aafd0d3c154cdd8fad53bfcc79046

During migration from check_http to check_curl, I found out that the results do not match for some checks with a lot of redirects. Since there was some changes to the check_curl recently I also wanted to test it on the latest version.

To test this, I ran three checks on the "mail.google.com/chat"

### Check_http with --onredirect=follow

```bash
$ ~/repositories/monitoring-plugins/plugins/check_curl --version
check_curl v2.4.0.872.gfd42 (monitoring-plugins 2.4git)
libcurl/8.14.1 OpenSSL/3.5.4 zlib/1.3.1 brotli/1.1.0 zstd/1.5.7 libidn2/2.3.8 libpsl/0.21.2 libssh2/1.11.1 nghttp2/1.64.0 nghttp3/1.8.0 librtmp/2.3 OpenLDAP/2.6.10
```

```
$ ~/repositories/monitoring-plugins/plugins/check_http -H mail.google.com -u '/chat' -t 60 -w 10 -c 10 --ssl --sni --onredirect=follow -4 -vvv | tee check_http.log
SSL initialized
GET /chat HTTP/1.1
User-Agent: check_http/v2.4.0.872.gfd42 (monitoring-plugins 2.4git)
Connection: close
Host: mail.google.com


https://mail.google.com:443/chat is 405 characters
STATUS: HTTP/1.1 302 Found
**** HEADER ****
Content-Type: application/binary
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Date: Mon, 01 Dec 2025 14:25:07 GMT
Location: https://mail.google.com/chat/
Server: ESF
Content-Length: 0
X-XSS-Protection: 0
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Alt-Svc: clear
Connection: close
**** CONTENT ****

Redirection to https://mail.google.com:443/chat/
SSL initialized
GET /chat/ HTTP/1.1
User-Agent: check_http/v2.4.0.872.gfd42 (monitoring-plugins 2.4git)
Connection: close
Host: mail.google.com


https://mail.google.com:443/chat/ is 381 characters
[...]

PR #2198, opened 2025-12-04
check_curl: append the query string from parsed uri

inqrphl:
Check the UriUriA object, and if query string exists append it to the new_url.

Function redir parses the new location header value using the uriParseUriA function already, which populates the query field. This field was already being printed, but it was not being appended to the new_url during its construction.

Redirection chain of check_curl --onredirect=follow now mimics the chain of check_http --onredirect=follow. Tested on the url: mail.google.com/chat

Related issue #2192


sni:
looks good to me. The failed test seems unrelated.

PR #2209, opened 2025-12-16
add proxy argument and improve dns cache usage

inqrphl:
add proxy argument that useing the -x and --proxy argument. add it to the static curl config struct, command usage and help outputs of the cli.

parse these argument together with the environment variables like http_proxy before setting the CURLOPT_PROXY in the curl configuration option. this is required, as there is no easy way to ascertain/get what the CURLOPT_PROXY that libcurl will use. by the point it is set by libcurl, we have no control over it anymore, and need it for the other steps in the configuration.

if the CURLOPT_PROXY is set, skip the DNS cache population which would set the CURLOPT_RESOLVE. this is currently not perfect however. if a proxy is set with socks4 or socks5 scheme, the host should be resolving the hostname.

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -239,39 +239,9 @@ mp_subcheck check_http(const check_curl_config config, check_curl_working_state 
 	// ==============
 	CURLcode res = curl_easy_perform(curl_state.curl);
 
-	if (verbose > 1) {
-		printf("* curl_easy_perform returned: %s\n", curl_easy_strerror(res));
-	}
-
 	if (verbose >= 2 && workingState.http_post_data) {
 		printf("**** REQUEST CONTENT ****\n%s\n", workingState.http_post_data);
 	}
-
-	/* If SSL cert checking is enabled, check the certs before checking for curl errors.
-	 * curl_state is updated after curl_easy_perform, and a cert check can be done at this point
-	 * check_http tries to check certs as early as possible, and returns afterwards unless
-	 * continue_after_check_cert is enabled. there may be servers with valid certificates
-	 * that return erroneus responses, but check_http returns OK since it only concerned with
-	 * certificates. Behave similarly here and check CURLcode after certificates.
-	 */
-#ifdef LIBCURL_FEATURE_SSL
-	if (workingState.use_ssl && config.check_cert) {
-		if (verbose > 1) {
-			printf("* adding a subcheck for the certificate\n");
-		}
-		mp_subcheck sc_certificate = check_curl_certificate_checks(
-			curl_state.curl, cert, config.days_till_exp_warn, config.days_till_exp_crit);
-
-		mp_add_subcheck_to_subcheck(&sc_result, sc_certificate);
-		if (!config.continue_after_check_cert) {
-			if (verbose > 1) {
-				printf("* returning after adding the subcheck for certificate, continuing after "
-					   "checking the certificate is turned off\n");
-			}
-			return sc_result;
-		}
-	}
-#endif
 
 	mp_subcheck sc_curl = mp_subcheck_init();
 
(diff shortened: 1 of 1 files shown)


plugins/check_curl.c
```diff
@@ -1009,6 +1010,10 @@ check_curl_config_wrapper process_arguments(int argc, char **argv) {
 			strncpy(result.config.curl_config.user_auth, optarg, MAX_INPUT_BUFFER - 1);
 			result.config.curl_config.user_auth[MAX_INPUT_BUFFER - 1] = 0;
 			break;
+		case 'x': /* proxy info */
+			strncpy(result.config.curl_config.proxy, optarg, DEFAULT_BUFFER_SIZE - 1);
+			result.config.curl_config.user_auth[DEFAULT_BUFFER_SIZE - 1] = 0;
```
RincewindsHat: why change `user_auth`? C&P error from above?

RincewindsHat:
Sorry, took me some time.
A lot of code there is to get the environment variables (_HTTP_RPOXY_ and so on), but [`man 3 CURLOPT_PROXY`](https://curl.se/libcurl/c/CURLOPT_PROXY.html) states that

> The CURLOPT_PROXY(3) and CURLOPT_NOPROXY(3) options override environment variables.

Am I correct to assume, that we can just leave the env things as they are right now?
the environment variables are working correctly at the moment.

inqrphl:
> Sorry, took me some time. A lot of code there is to get the environment variables (_HTTP_RPOXY_ and so on), but [`man 3 CURLOPT_PROXY`](https://curl.se/libcurl/c/CURLOPT_PROXY.html) states that
> 
> > The CURLOPT_PROXY(3) and CURLOPT_NOPROXY(3) options override environment variables.
> 
> Am I correct to assume, that we can just leave the env things as they are right now? the environment variables are working correctly at the moment.

Yes this commit handles the environment variables.

The idea is to get proxy environment variables early in the function and set CURLOPT_PROXY manually before the request. Then we determine whether the proxy or the requestor resolves the target hostname, using the proxy scheme and the target protocol. This has to be done before populating the DNS cache, which is just bellow the additions to the code.

I did not find another way of determining what proxy libcurl would end up using, unless I set it manually myself by using the same environment variables discovery methods. Libcurl does not expose a function/variable afaik. This also led to me adding the proxy argument.

That would technically consititute as an override of proxy variables by manually setting CURLOPT_PROXY, but using the same values.

More importantly I realized now that I am not considering no-proxy environment variables when setting the proxy_resolves_hostname variable. There could be a default proxy set, but the target host could be included in no-proxy list. Then the requestor should resolve the hostname. 

I am not getting the no-proxy variables at all, or iterating through the no proxy hosts for a match with the target hostname.

inqrphl:
@RincewindsHat  No worries, I was just finished with the code changes. Testing it on the customer fleet now.

I have the subdomain and IP CIDR detection for no_proxy. They are now also used in the codepath that prematurely determines if proxy or host resolves the DNS address, and optionally skipping to populate DNS cache.

In the tests I wrote, I set up local apache2 to work with IPv6 unconditionally, and stopped removing [::1] from /etc/hosts. This was intended for check_curl IPv6 subdomain noproxy detection tests, but it seems like IPv6 is being used unconditionally now: #2219  . I can leave it at that.

Other than that, I have written some tests about its behavior. I wrote both t/check_curl.pl and tests/check_curl.pl for testing.

t/check_curl.pl tests require a working proxy server. The debian testing container was already setting up squid proxy, which supports http proxies on IPv4 and IPv6. I configured it to use IPv6 as well and enabled proxy usage for ACL localhost. 

I did not install or set up another proxy server that supports socks4/5, but it is not necessarily required, this codebase only has to determine if proxy resolves hostname or not and I check it by looking at the verbose output. Libcurl will handle the proxy specifics after its set.

tests/check_curl.pl runs its own server, so tests in that file are mainly about argument parsing, like uppercase/lowercase environment variables and direct cli arguments overriding eachother for CURLOPT_PROXY. I also check some no_proxy cases there.

PR #2211, opened 2025-12-23
use %lld for time_t and cast parameters

sthen:
some OS define time_t as 64-bit on all architectures including 32-bit ones; adjust format strings to allow for this


sthen:
oops, the check_disk.c patch isn't time_t, but similar. parameter_list in plugins/check_disk.d/utils_disk.h defines these three (used_bytes, free_bytes, total_bytes) as uint64_t so need a long long type on 32-bit.

PR #2212, opened 2025-12-24
use pkg-config XX netsnmp instead of net-snmp-config XX

sthen:
`net-snmp-config --cflags` returns the flags used to build Net-SNMP, which in 5.9.5+ includes `-Werror=declaration-after-statement`; code in check_snmp.c is not ready for this. (I looked at fixing that instead but it's rather intrusive). "pkg-config --cflags netsnmp" just returns flags needed to build against the library.

Not super clean as it doesn't degrade nicely if pkg-config is not available; not sure how much of a problem that is?

### DIFF
--- a/.github/monitoring-plugins.spec
+++ b/.github/monitoring-plugins.spec
@@ -37,6 +37,7 @@ BuildRequires:  gettext-devel
 BuildRequires: gcc
 BuildRequires: make
 BuildRequires: automake
+BuildRequires:  pkgconf
 
 ### Check Plugin specific build requirements
 # check_by_ssh


RincewindsHat:
ok, the error in the CI are an error of mine. Missed `-lm` (math lib) previously and that only worked since `netsnmp-config` listed it too.
Will fix that.

RincewindsHat:
hm, no `pkg-conf` files on rocky linux 8 :-(

waja:
> hm, no `pkg-conf` files on rocky linux 8 :-(

https://rockylinux.pkgs.org/8/rockylinux-baseos-x86_64/pkgconf-pkg-config-1.4.2-1.el8.i686.rpm.html ?

Issue #2201, opened 2025-12-11
Bump Gnulib to stable-202507


oxzi:
To build v3.0.0-rc1 on OpenBSD, I had to update the ports patch on the vendored Gnulib in `gl` as follows.

```diff
--- gl/fseeko.c.orig
+++ gl/fseeko.c
@@ -101,6 +101,8 @@ fseeko (FILE *fp, off_t offset, int whence)
 #elif defined EPLAN9                /* Plan9 */
   if (fp->rp == fp->buf
       && fp->wp == fp->buf)
+#elif defined __OpenBSD__ /* OpenBSD */
+  if (0) /* fseeko() is compatible with what gnulib expects */
 #elif FUNC_FFLUSH_STDIN < 0 && 200809 <= _POSIX_VERSION
   /* Cross-compiling to some other system advertising conformance to
      POSIX.1-2008 or later.  Assume fseeko and fflush work as advertised.
```

As I wanted to bring this upstream, I realized that this change is [already there](https://gitweb.git.savannah.gnu.org/gitweb/?p=gnulib.git;a=commit;h=1594e825f4b65c95fe8a715f57ea54ed03a8f5b5). Furthermore, it is also [present](https://gitweb.git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob;f=lib/fseeko.c;hb=refs/heads/stable-202507#l107) in the `stable-202507` tag.

Please consider updating the Gnulib. Thanks.

PR #2213, opened 2025-12-28
Sync with Gnulib stable-202507 code (a8ac9f9ce5)

RincewindsHat:
solves #2201


RincewindsHat:
hm, that was gnulib-`master`, probably should go to `*-stable` instead

PR #2225, opened 2026-01-15
OpenBSD: pledge(2) some network-facing checks

oxzi:
OpenBSD's [`pledge(2)`](https://man.openbsd.org/pledge.2) system call allows the current process to self-restrict itself, being reduced to promised pledges. For example, unless a process says it wants to write to files, it is not allowed to do so any longer.

This change starts by calling `pledge(2)` in some network-facing checks, removing the more dangerous privileges, such as executing other files.

My initial motivation came from `check_icmp`, being installed as a setuid binary and (temporarily) running with root privileges. There, the `pledge(2)` calls result in `check_icmp` to only being allowed to interact with the network and to `setuid(2)` to the calling user later on.

Afterwards, I went through my most commonly used monitoring plugins directly interacting with the network. Thus, I continued with `pledge(2)`-ing `check_curl` - having a huge codebase and all -, `check_ntp_time`, `check_smtp`, `check_ssh`, and `check_tcp`.

For most of those, the changes were quite similar: start with network-friendly promises, parse the configuration, give up file access, and proceed with the actual check.

---

While this change might only address a niche, I would argue it might help a lot by only a few lines in case of disaster. But I am open for other opinions.

### DIFF
--- a/plugins-root/check_icmp.c
+++ b/plugins-root/check_icmp.c
@@ -812,6 +812,15 @@ void parse_address(const struct sockaddr_storage *addr, char *dst, socklen_t siz
 }
 
 int main(int argc, char **argv) {
+#ifdef __OpenBSD__
+	/* - rpath is required to read --extra-opts (given up later)
+	 * - inet is required for sockets
+	 * - dns is required for name lookups (given up later)
+	 * - id is required for temporary privilege drops in configparsing and for
+	 *   permanent privilege dropping after opening the socket (given up later) */
+	pledge("stdio rpath inet dns id", NULL);
+#endif // __OpenBSD__
+
 	setlocale(LC_ALL, "");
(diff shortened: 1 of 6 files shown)


RincewindsHat:
Love it. I had a workshop about that recently and it sounds good :-)
I was thinking about asking for some explanatory comments, but now I am not sure whether that would add more than `man 2 pledge` would explain.

oxzi:
Rebased the PR branch and added a short explanation about the chosen promises - next to the omnipresent `stdio` and `error` promises - above the first `pledge(2)` call in each check.

sthen:
btw I've added the patches from this PR to the OpenBSD -current packages, have seen no problems myself and had no reports of problems.

RincewindsHat:
If it works for you, it's fine for me :-)

PR #1856, opened 2023-03-20
check_procs: Implement --exclude-process to exclude specific processes

ckujau:
Signed-off-by: Christian Kujau <lists@nerdbynature.de>

PR #2235, opened 2026-02-18
Fix check_procs -T option

mgagne:
Previous PR #1856 introduced an issue where -T is now expecting an argument.

This change fixes the issue.


RincewindsHat (approved):
Thanks a lot, I did not notice that one before :-)

PR #2241, opened 2026-03-16
Remove deprecated autoconf macro (autoupdate)

RincewindsHat:


RincewindsHat:
has to wait for distros to pick up.

PR #2259, opened 2026-04-17
Error summary

### LINKED ISSUE #2258: First line of plugin output lacks descriptive status summary
Hi,

I just learned that the output of the plugins is going to change drastically.

```
$ ./check_tcp -H cb.rbfh.de -p 444
[CRITICAL] - ok=0, warning=0, critical=1, unknown=0
        \_[CRITICAL] - Connection to cb.rbfh.de on port 444 was REFUSED
```

Basically, I like that tree, but I think it should live entirely after the first line of output. The first line of output is special, at least in Naemon and Nagios-based systems. It (the first line) should contain a brief description of the problem.

This way, it also violates https://www.monitoring-plugins.org/doc/guidelines.html#AEN29

Also, many notification commands just send the first line because the other lines may contain sensitive data. The first line of the tree is not very helpful; if I received that on a pager, I would be upset.

RincewindsHat:
This change uses the output of the first non-OK subcheck as the summary for the whole check

At least improves the situation regarding #2258.

@datamuc  @lgmu this one would be the "easy" solution without touching everything again.
I am hereby guilt tripping you into testing this to give me feedback :-)


datamuc:
Funny, I have a free month of codex, and it looks pretty much like the first iteration I had with it[^1].

But that has problems:

```
$ ./check_curl -H cb.rbfh.de -p 443 
[WARNING] - Testing http://cb.rbfh.de:443/
        \_[WARNING] - Testing http://cb.rbfh.de:443/
                \_[OK] - HTTP/1.1 400 Bad Request - 661 bytes in 0.051 second response time
                \_[OK] - Total connection time: 0.051480s
                \_[OK] - Status line output matched "HTTP/"
                \_[WARNING] - HTTP return code: 400|'time'=0.051480s;;;
```

Which isn’t any better.

So I told it to pick the deepest node with the worst condition[^2]. That works quite well so far, but I haven’t done very extensive testing yet:

```
$ ./check_curl -H cb.rbfh.de -p 443  -C 400 --continue-after-certificate -r xtml
[CRITICAL] - Regex "xtml" in body matched not
[CRITICAL] - ok=0, warning=0, critical=1, unknown=0
        \_[CRITICAL] - Testing https://cb.rbfh.de:443/
                \_[WARNING] - Certificate 'cb.rbfh.de' expires in 52 day(s) (Wed 10 Jun 2026 04:56:41 AM GMT +0000)
                \_[OK] - HTTP/1.1 200 OK - 10990 bytes in 0.100 second response time
                \_[OK] - Total connection time: 0.100337s
                \_[OK] - Status line output matched "HTTP/"
                \_[OK] - HTTP return code: 200
                \_[CRITICAL] - Regex "xtml" in body matched not|'time'=0.100337s;;;   
```
EDIT: Just saw warning=0 even though there is a warning in that tree.

Maybe that mp API also needs something like:

```
mp_set_ok_message(overall, "text that is displayed in the first line if everything is fine")
```

This is probably only needed once per plugin and could be optional.

BTW: I believe "matched not" is incorrect grammar.

[^1]: Well, not the code, but the result and my approach (the first non-OK subcheck as the summary). The code actually didn’t look that great, I think, but I’m not very good at judging C code.
[^2]: I can push that slope somewhere if you’re brave and want to take a look.

RincewindsHat:
that `warning=0` is correct in that only the first level of nodes is counted and that node is `CRITICAL`. Arguably that is confusing, but IMHO better than recursively counting every node which is not really useful since the aggregate nodes are not counted separately.
One could think about only counting leave nodes though...

I am open to suggestions to replace "matched not", it is asthetically not pleasing (although grammatically correct I think).

A default OK message sounds not bad.

I wouldn't mind to take a look at the recursive search approach, was it sloperated?

datamuc:
> Was it sloperated?

Of course, https://github.com/monitoring-plugins/monitoring-plugins/compare/master...KIBruh:monitoring-plugins:codex?expand=1

RincewindsHat:
@datamuc Implemented the recursive approach with the last commit. Would you care to test it a bit?

datamuc:
Hi,

Thank you, I think it is much better now.

The first line of an OK check is still a bit pathetic though:

```
$ ./check_curl -H cb.rbfh.de 
[OK] - ok=1
...
```
it's nicer to get

> TCP OK - 0.030 second response time on cb.rbfh.de port 80

as a recovery message than

> [OK] - ok=2

Well, i think I could live with that if i would still do on call. I will try to deploy some of these checks somewhere so i can see them in the GUI and see how it feels there.

RincewindsHat:
@datamuc Did you evaluate this a bit? Would be nice to have some feedback before the merge

datamuc:
Sorry, completely forgot about that.

2.x:
Screenshot 2026-05-19 110316

main/master branch:
Screenshot 2026-05-19 111311


error_summary branch:
Screenshot 2026-05-19 111543


It's an improvement I guess. But feels still quite disruptive. I really would like to hear more opinions ...

sni:
Would it hurt to do the same thing for OKs? The "ok=1" does not add any value but instead prevents users from seeing the actual plugin output.

lgmu:
Honestly I still prefer the 2.X output...
It provided everything you wanted to know in one line without having to look at multiple lines.

RincewindsHat:
hm, I'll put "nice OK message" on the list then.

But i tend to go continue with the approach, the old output was borderline useless in some cases (`check_disk`)

lgmu:
I agree, but that could just be fixed by parsing the default output differently like `OK - /opt 1.690 GiB/7.593 GiB (22.3%)`

RincewindsHat:
I honestly think, that putting everything in a single line has its constraints which can not be overcome.
And I don't want to waste the opportunity to have a useful multi-line output.
The frontend there (Thruk?) could just strip the newlines out for a singe line or someone(tm) implements a single line output format here.

@sni didn't we shred the single line output mode, because nobody felt it was necessary some time ago?

sni:
> @sni didn't we shred the single line output mode, because nobody felt it was necessary some time ago?

well yes, we agreed that everything supports multiline output. But i'd still stick to the format of the first line with a useful summary and details on subsequent lines.

Sure, i could add a output formatter in Thruk and transform the plugin output into a readable format. But the output is also used in notifications, mails, etc... I'd say it's much easier for the plugin to provide a reasonable output.

RincewindsHat:
Ok, I am going to merge this since I get the vibe that this here is an improvement so far.
I am going to work on changing the OK output in a separate set of changes.

Issue #2265, opened 2026-05-08
check_http and check_curl: custom return states for timeouts


inqrphl:
Currently, the timeouts always return CRITICAL. This does not help when discerning if a request got a reply that went above the critical threshold, or if a request timed out since the server did not respond in time. At least that is the case when someone looks at the return state, and does not inspect the message.

Defaulting to CRITICAL when timing out works for most cases, but not all of them. Raising a CRITICAL has consequences related when using naemon /+ Thruk , compared to UNKNOWN.

An example where a timeout may be seen as UNKNOWN if the check is looking for a string in the response body. The did not get a body due to timing out, so it cannot determine if something is in the body or not. Or we are looking for requests that need to take a while to respond, so using a small timeout and returning OK when it times out is the way to go.

Therefore, adding a new argument like ```--timeout-custom-result-state [ok,warning,critical,unknown]``` can help to customize the return code when requests time out.

I have a working example that works on both check_curl and check_http in this [commit](https://github.com/inqrphl/monitoring-plugins/commit/4f35a424b9385e4c360bb3c61af6e8a99a9ddd42) . 

The required changes are small, in check_http we parse the argument and set the socket_timeout_state found in netutils.c . This is used in the SIGALRM handler function socket_timeout_alarm_handler.

In check_curl, we parse the argument and save the given state into the config as on_timeout_result_state . After performing a curl request, we check if it returned CURLE_OPERATION_TIMEDOUT , then return a subcheck result using on_timeout_result_state as its state.

sni:
i agree, i would make the argument name a bit shorter, like ex.: `--timeout-result`
I would also move the argument parsing to a central helper function, because it could be useful
for other plugins as well.
Also i would add support for numeric return codes, like 1,2 or 3

PR #2266, opened 2026-05-12
check_http and check_curl: custom timeout return state

inqrphl:
As discussed in #2265

Added a new argument to both check-curl and check-http  called ```--timeout-state``` . It customizes the return state when a timeout occurs. Takes either a return code ```0,1,2,3``` or mixed case return state name ```ok,warning,critical,unknown``` .

check_http:  parse the argument and set the ```socket_timeout_state found``` in netutils.c . This is used in the ```SIGALRM``` handler function socket_timeout_alarm_handler. The ```SIGALRM``` is scheduled using ```alarm()``` call.

check_curl: parse the argument and save the given state into the config as on_timeout_result_state . After performing a curl request, check if it returned curl return code ```CURLE_OPERATION_TIMEDOUT``` . If so, change the ```sc_curl``` subcheck to use the ```on_timeout_result_state``` from config, and return ```sc_result``` check immediately.

Clang-format is applied

### DIFF
--- a/plugins/check_curl.c
+++ b/plugins/check_curl.c
@@ -276,7 +276,7 @@ mp_subcheck check_http(const check_curl_config config, check_curl_working_state 
 		xasprintf(&sc_curl.output, _("cURL got a timeout error"));
 		sc_curl = mp_set_subcheck_state(sc_curl, config.on_timeout_result_state);
 		mp_add_subcheck_to_subcheck(&sc_result, sc_curl);
-		return sc_curl;
+		return sc_result;
 	}
 
 	/* Curl errors, result in critical Nagios state */
@@ -898,7 +898,7 @@ check_curl_config_wrapper process_arguments(int argc, char **argv) {
 		STATE_REGEX,
 		OUTPUT_FORMAT,
 		NO_PROXY,
-		TIMEOUT_CUSTOM_RESULT_STATE,
+		TIMEOUT_RESULT,
 	};
 
 	static struct option longopts[] = {
@@ -948,7 +948,7 @@ check_curl_config_wrapper process_arguments(int argc, char **argv) {
 		{"cookie-jar", required_argument, 0, COOKIE_JAR},
 		{"haproxy-protocol", no_argument, 0, HAPROXY_PROTOCOL},
 		{"output-format", required_argument, 0, OUTPUT_FORMAT},
-		{"timeout-custom-result-state", required_argument, 0, TIMEOUT_CUSTOM_RESULT_STATE},
+		{"timeout-result", required_argument, 0, TIMEOUT_RESULT},
 		{0, 0, 0, 0}};
 
 	check_curl_config_wrapper result = {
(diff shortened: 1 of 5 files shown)


inqrphl:
Example output:

```
:~/repositories/monitoring-plugins$ ./plugins/check_curl --hostname 127.0.0.1 --port 9613 --timeout-result unknown --timeout 5
[UNKNOWN] - ok=0, warning=0, critical=0, unknown=1
        \_[UNKNOWN] - Testing http://127.0.0.1:9613/
                \_[UNKNOWN] - cURL returned 28 - Operation timed out after 5002 milliseconds with 0 bytes received

                
:~/repositories/monitoring-plugins$ ./plugins/check_http --hostname 127.0.0.1 --port 9613 --timeout-result 0 --timeout 5
[OK] - ok=1, warning=0, critical=0, unknown=0
        \_[OK] - Socket timeout after 5 seconds
```

plugins/check_http.c
```diff
@@ -298,6 +300,21 @@ bool process_arguments(int argc, char **argv) {
 				socket_timeout = atoi(optarg);
 			}
 			break;
+		case TIMEOUT_RESULT:
+			if (!strcmp(optarg, "0") || !strcmp(optarg, "ok")) {
```
RincewindsHat: I would prefer `strcasecmp` for the string version of the state.
This would allow for "OK"

plugins/check_http.c
```diff
@@ -1883,6 +1900,10 @@ void print_help(void) {
 
 	printf(UT_CONN_TIMEOUT, DEFAULT_SOCKET_TIMEOUT);
 
+	printf(" %s\n", "--timeout-result=RESULT|INTEGER");
+	printf("    %s\n", _("Timeouts default to returning STATE_CRITICAL."));
+	printf("    %s\n", _("This argument changes the return state on timeouts."));
```
RincewindsHat, line 1905: Listing all the available states would not hurt here.

RincewindsHat (changes requested):

plugins/check_http.c
inqrphl: Printout out this now:   --timeout-result=ok|warning|critical|unknown|0|1|2|3

inqrphl:
On check_curl, i am getting the output "Operation timed out" instead of "Connection timed out" on my Debian 13 system after timeout. This is written to variable ```extern char errbuf``` somewhere. 

Let me investigate this before merging.

plugins/check_http.c
inqrphl: Done, thanks

plugins/check_curl.c
```diff
@@ -271,6 +271,15 @@ mp_subcheck check_http(const check_curl_config config, check_curl_working_state
 
 	mp_subcheck sc_curl = mp_subcheck_init();
 
+	/* Custom handling for timeouts */
```
RincewindsHat: could you move this block a few lines inside that `if (res != CURLE_OK)` block?
I think it would make more sense to have all of the error handling in one place.

RincewindsHat (changes requested):
One small change, otherwise I approve

plugins/check_curl.c
inqrphl: Moved it inside ```if (res != CURLE_OK)``` block.

I check if its ```CURLE_OPERATION_TIMEDOUT``` or something else with an if-else and not a switch, since there are only two cases for now.

RincewindsHat (approved):
Thank you very much :-)

PR #2270, opened 2026-05-30
Ok summary

### LINKED ISSUE #2258: First line of plugin output lacks descriptive status summary

RincewindsHat:
Second part to fix #2258

This add functionality to the library to set an OK summary (meaning: a function to set the first line of output if the overall state is OK).

### DIFF
--- a/lib/output.c
+++ b/lib/output.c
@@ -105,6 +105,7 @@ static inline char *fmt_subcheck_perfdata(mp_subcheck check) {
  */
 mp_check mp_check_init(void) {
 	mp_check check = {
+		.ok_summary = NULL,
 		.evaluation_function = &mp_eval_check_default,
 		.default_output_override = NULL,
 		.default_output_override_content = NULL,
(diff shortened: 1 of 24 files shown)


RincewindsHat:
Implemented for all the plugins with the new output functionality.

PR #2277, opened 2026-06-11
Check ntp time delay

### LINKED ISSUE #1295: check_ntp has to respect ntpd-parameters like average and minimum
actually check_ntp/check_tnp_time submits udp packest to the ntp-server with an interval of about one second. In the documentation of ntpd servers it menttioned, that ntp-requests are/can be discarded when appearing with this high frequency: 

From ´http://www.eecis.udel.edu/~mills/ntp/html/accopt.html#discard: 

discard [ average avg ][ minimum min ] [ monitor prob ]
    Set the parameters of the rate control facility which protects the server from client abuse. If the limited flag is present in the ACL, packets that violate these limits are discarded. If, in addition, the kod flag is present, a kiss-o'-death packet is returned. See the Rate Management page for further information. The options are:

```
average avg
    Specify the minimum average interpacket spacing (minimum average headway time) in log2 s with default 3.
minimum min
    Specify the minimum interpacket spacing (guard time) in seconds with default 2.
monitor
    Specify the probability of being recorded for packets that overflow the MRU list size limit set by mru maxmem or mru maxdepth. This is a performance optimization for servers with aggregate arrivals of 1000 packets per second or more.
```

For this, check_ntp will have to adopt its own request frequency or there have to command line parameters to adjust average and minimum.

RincewindsHat:
This patch set mainly adds a delay to requests made with `check_ntp_time` to avoid
triggering rate limits on NTP servers.
Additionaly to a default delay, it adds an option to set the delay.

Another change in here validates whether a given socket path fits into the
target struct and aborts execution if not (which might cause a segfault oder other
problems otherwise.

The original patch set comes from a Paul Crawford which I hope is adequatly reflected
in the authorship of the commits.

This should fix #1295 and #1142

### DIFF
--- a/plugins/check_ntp_time.c
+++ b/plugins/check_ntp_time.c
@@ -43,6 +43,7 @@
 #include "thresholds.h"
 #include "check_ntp_time.d/config.h"
 #include <netinet/in.h>
+#include <string.h>
 #include <sys/socket.h>
 
 static int verbose = 0;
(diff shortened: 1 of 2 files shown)


RincewindsHat:
actually I forgot the effectual implementation here....

PR #2279, opened 2026-06-17
check_mysql_query: add option to hide query in output

### LINKED ISSUE #1600: [check_mysql_query] Add option to hide query
I use this check a few times with a very large SQL query, so it would be nice to be able to hide the query from the output.

Maybe replace it with some custom text?

RincewindsHat:
Adds an option to `check_mysql_query` which hides the query in the plugin output.

fixes #1600


datamuc:
in `check_pgsql` we have `--queryname` for that

RincewindsHat:
ha, good idea.

PR #2283, opened 2026-06-18
check_mysql: Fix MariaDB casing

C0rn3j:
Correctly stylizes `MariaDB` (from `Mariadb`) in the `check_mysql` plugin output.


RincewindsHat (approved):

RincewindsHat:
Thank you :-)

PR #2287, opened 2026-06-19
Fix/check smtp successfull auth

### LINKED ISSUE #2284: 3.0.0: check_smtp: SIGABRT (assert subcheck.output != NULL) on successful AUTH LOGIN
`check_smtp -A LOGIN -U <user> -P <pass>` aborts with SIGABRT **when the SMTP authentication succeeds**. A *failed* login reports a normal CRITICAL/WARNING, so the crash paradoxically only happens on success.

### Steps to reproduce

```
check_smtp -H <host> -p 587 -S -A LOGIN -U <user> -P <correct-password>
```

against a server that accepts the credentials (replies `235`).

### Actual output

```
check_smtp: output.c:131: mp_add_subcheck_to_check: Assertion `subcheck.output != NULL' failed.
Aborted (exit via signal 6)
```

### Expected output

A normal `SMTP OK` result, e.g. the auth subcheck reporting that authentication succeeded.

### Root cause

In `plugins/check_smtp.c`, the AUTH LOGIN subcheck is created with `mp_subcheck_init()`, which leaves `sc_auth.output == NULL`. Every **error** branch sets `sc_auth.output` via `xasprintf()`, but the **success** path (server replies `235`) falls through to the final `break;` without ever setting `sc_auth.output`:

```c
if (strncmp(buffer, "235", 3) != 0) {
    xasprintf(&sc_auth.output, "invalid response received after authpass");
    sc_auth = mp_set_subcheck_state(sc_auth, STATE_CRITICAL);
    break;
}
break;                 // <-- success path: sc_auth.output is still NULL
} while (false);
...
mp_add_subcheck_to_check(&overall, sc_auth);   // asserts output != NULL -> abort()
```

`mp_add_subcheck_to_check()` (`lib/output.c:131`) then hits `assert(subcheck.output != NULL)` and aborts.

### Suggested fix

Set an output string and an explicit OK state on the success path before the final `break;`:

```c
}

/* Authentication succeeded (server replied 235) */
xasprintf(&sc_auth.output, "Authentication succeeded");
sc_auth = mp_set_subcheck_state(sc_auth, STATE_OK);
break;
} while (false);
```

RincewindsHat:
The subcheck for authentication in `check_smtp` had not output set for successful authentication. This patch set fixes this.

Additionaly it removes an unused variable.

And the test for an expected response was not used at all.

fixes #2284

### DIFF
--- a/plugins/check_smtp.c
+++ b/plugins/check_smtp.c
@@ -481,6 +481,10 @@ int main(int argc, char **argv) {
 	if (config.authtype != NULL) {
 		mp_subcheck sc_auth = mp_subcheck_init();
 
+		// set sucess values here, failure will be set below
+		sc_auth = mp_set_subcheck_default_state(sc_auth, STATE_OK);
+		xasprintf(&sc_auth.output, "Login succeded");
+
 		if (strcmp(config.authtype, "LOGIN") == 0) {
 			char *abuf;
 			int ret;
--- a/plugins/check_smtp.c
+++ b/plugins/check_smtp.c
@@ -483,9 +483,9 @@ int main(int argc, char **argv) {
 	if (config.authtype != NULL) {
 		mp_subcheck sc_auth = mp_subcheck_init();
 
-		// set sucess values here, failure will be set below
+		// set success values here, failure will be set below
 		sc_auth = mp_set_subcheck_default_state(sc_auth, STATE_OK);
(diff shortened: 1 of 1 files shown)


RincewindsHat:
@d--j this should fix #2284. Could you test whether it fixes your problem?

Issue #2288, opened 2026-06-19
3.0.0: check_disk tries to access `/sys/kernel/debug/tracing` and crashes


C0rn3j:
This command works fine on 2.3.5, crashes on 3.0.0:

```bash
sudo -u nagios '/usr/lib/nagios/plugins/check_disk' '-K' '10%' '-W' '15%' '-c' '10%' '-w' '20%' '-X' 'none' '-X' 'tmpfs' '-X' 'sysfs' '-X' 'proc' '-X' 'configfs' '-X' 'devtmpfs' '-X' 'devfs' '-X' 'mtmfs' '-X' 'tracefs' '-X' 'cgroup' '-X' 'fuse.*' '-X' 'fuse.gvfsd-fuse' '-X' 'fuse.gvfs-fuse-daemon' '-X' 'fuse.portal' '-X' 'fuse.sshfs' '-X' 'fdescfs' '-X' 'overlay' '-X' 'nsfs' '-X' 'squashfs' '-A' '-I' '^/var/lib/docker' '-I' '^/backup'

DISK CRITICAL - /sys/kernel/debug/tracing is not accessible: Permission denied
```

Note `-X tracefs` being present which I suppose I'd expect to mean that the tracing interface shouldn't be accessed?

I see that Icinga [excluded cgroup and tracefs by default](https://github.com/Icinga/icinga2/issues/4184#issuecomment-273099847) to prevent this error, so I suppose there's a regression in 3.0.0 where it's checking them anyways?

If anything, I'd hope the error would show the device it was checking to be able to debug what it failed on better.

I am also not exactly sure why the third entry here is present on some 24.04 Ubuntu servers but not others (where it works), but I suppose that's beside the point.
```bash
# mount | grep -E 'debug|trace'
debugfs on /sys/kernel/debug type debugfs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/debug/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
```

RincewindsHat:
hm, it might be that the filtering moved AFTER `stat`ing the fs, which would cause this.

RincewindsHat:
as a workaround `--ignore-missing` would prevent the crash.

RincewindsHat:
and setting `-X` actually prevents calling `stat` on unwanted fs types.
[...]

PR #2289, opened 2026-06-25
check_disk: prevent using stat on inaccessible filesystems

RincewindsHat:
This patch comes from @d--j in issue #2288.


RincewindsHat:
well, seems to work.

PR #2296, opened 2026-07-04
configure.ac,plugins/common.h: update sys/poll.h -> poll.h

orlitzky:
Make these warnings go away:

```
In file included from ../plugins/common.h:119,
                 from check_icmp.c:46:
/usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
  1 | #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
    |  ^~~~~~~
```

The bare `<poll.h>` is correct according to POSIX (and not just my libc): https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/poll.h.html

There's one more sys/poll.h hiding behind an `#if` for Solaris and HPUX that I left alone since I know nothing about those.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -598,7 +598,7 @@ dnl Checks for header files.
 dnl
 
 AC_HEADER_SYS_WAIT
-AC_CHECK_HEADERS(signal.h syslog.h uio.h errno.h sys/time.h sys/socket.h sys/un.h sys/poll.h)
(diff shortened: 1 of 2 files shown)


RincewindsHat (approved):
I don't know much about this, but if it still works and the standard tells one to do this ....

Thank you :-)

Issue #2160, opened 2025-09-26
TODO: Build tests with different features


RincewindsHat:
It would be nice to have build tests for more different configurations. Ideally even on more Unixes (although that might be tricky with Github).

This is an issue to discuss the topic and/or track progress

Some configurations I thought of:

 - `gnutls` instead of `openssl`
 - no TLS at all (probably doesn't work right now and I am not sure if that is needed/wanted)

sni:
i am using this https://github.com/vmactions/freebsd-vm in other projects to run tests on freebsd. Works ok most of the times, sometimes it hangs during provisioning...

PR #2304, opened 2026-07-29
CI: Add build and unit tests on FreeBSD, OpenBSD and NetBSD

neilpang:
Refs #2160.

Adds `.github/workflows/test-unix.yml` with one job per platform. Each boots a
QEMU VM on the normal `ubuntu-latest` runner using the `vmactions` actions @sni
mentioned in the issue, then runs `tools/setup` -> `./configure --enable-libtap`
-> `make` -> the `lib/tests` unit tests.

Triggers are `push` and `pull_request` plus `workflow_dispatch`, the same as
`test.yml`.

### What it covers, and what it deliberately does not

The top level `make test` is **not** run in the VMs. It depends on the service
zoo that `.github/prepare_debian.sh` builds (apache, squid, slapd, mariadb,
postfix, snmpd, sshd), which is not reasonable to reproduce three more times.
What these jobs cover is the autotools bootstrap, the full C build of every
plugin, and the self-contained libtap unit tests under `lib/tests`.

### Verification

https://github.com/neilpang/monitoring-plugins/actions/runs/30911208131

| | FreeBSD 15.1-RELEASE-p1 | OpenBSD 7.9 | NetBSD 11.0 |
|---|---|---|---|
| `CC`/`CCLD` invocations | 196 | 201 | 211 |
| libtap | Files=10, Tests=265, PASS | same | same |

### The two fixes that come with it

**`check_icmp` does not build on NetBSD.**

```
check_icmp.d/./config.h:88:17: error: field 'stime' has incomplete type
   88 |  struct timeval stime;
```

`plugins-root/check_icmp.d/config.h` uses `struct timeval` without including
`<sys/time.h>`. Linux and FreeBSD pull it in transitively; NetBSD does not.

**`lib/Makefile.am` recursed with a bare `make`.**

It used `cd tests && make $@` where every other recursive rule in the tree uses
`$(MAKE)`, so the recursion swapped make implementation halfway through.

The committed `lib/tests/test_generic_output` binary and its missing
`.gitignore` entry were in an earlier version of this PR; #2305 handled both
upstream, so those hunks are gone and `test_generic_output` now builds and runs
on all three BSDs -- which is where the jump from 246 to 265 subtests comes
from.

### OpenBSD versions are resolved at run time

Per @oxzi: nothing is pinned. `pkg_add` installs the newest `autoconf` and
`automake` the mirror offers, and `AUTOCONF_VERSION`/`AUTOMAKE_VERSION` are read
back off the binaries it installed, which is what the `metaauto` wrappers
dispatch on. The run above resolved autoconf 2.72 and automake 1.18.

### Not covered

- **Solaris**: a separate PR, as agreed. The job is green on Solaris 11.4; it
  needed fixes in four files, two of which are not Solaris-specific --
  `lib/tests/test_cmd.c` never NULL-terminates its argv array and so reads
  uninitialised memory on every platform, and `check_swap`'s SVR4 branch calls
  `getSwapFromSwapctl_SRV4()` while the function is named
  `getSwapFromSwap_SRV4()`, so that path has never been compiled by anything.
- **`gnutls` instead of `openssl`, and no-TLS builds**: not attempted here.

### DIFF
--- a/.github/workflows/test-unix.yml
+++ b/.github/workflows/test-unix.yml
@@ -19,6 +19,10 @@ on:
         description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
         required: false
         default: false
+  push:
+    branches:
+      - '*'
+  pull_request:
 
 jobs:
   freebsd:
--- a/.github/workflows/test-unix.yml
+++ b/.github/workflows/test-unix.yml
@@ -1,0 +1,102 @@
+---
+name: Tests on other Unixes
+
+# Build and run the self-contained unit tests on the Unixes that
+# configure.ac has explicit support for but that GitHub does not offer as a
+# hosted runner. Each job boots a QEMU VM via the vmactions actions.
+#
+# These jobs deliberately do NOT run the top level "make test": that suite
+# needs a full service zoo (apache, squid, slapd, mariadb, postfix, snmpd,
+# sshd, ...) which .github/prepare_debian.sh sets up only for the Linux job.
+# What is exercised here is the autotools bootstrap, the C build of every
+# plugin, and the standalone libtap unit tests under lib/tests.
+
+on:
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        type: boolean
+        description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
+        required: false
+        default: false
+
+jobs:
+  freebsd:
+    name: Running build and unit tests on FreeBSD
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v7
+      - name: Build and test in a FreeBSD VM
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          prepare: |
+            pkg install -y autoconf automake gmake
+          run: |
+            set -e
+            uname -a
+            tools/setup
+            ./configure --enable-libtap
+            gmake
+            cd lib && gmake test
+
+  openbsd:
+    name: Running build and unit tests on OpenBSD
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v7
+      - name: Build and test in an OpenBSD VM
+        uses: vmactions/openbsd-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          # OpenBSD carries about eighteen parallel autoconf packages and a
+          # dozen automake ones, so pkg_add needs an exact version and the
+          # metaauto wrappers then dispatch on AUTO*_VERSION. Both are
+          # resolved at run time -- the newest the mirror offers, and the
+          # version read back off the binary that pkg_add installed -- so
+          # nothing here needs revisiting when the next release lands.
+          prepare: |
+            set -e
+            newest() {
+              pkg_info -Q "$1" | grep -E "^$1-[0-9]" | sort -V | tail -1
+            }
+            pkg_add -I "$(newest autoconf)" "$(newest automake)" gmake
+          run: |
+            set -e
+            AUTOCONF_VERSION=$(ls /usr/local/bin/autoconf-* | sed 's:.*/autoconf-::' | sort -V | tail -1)
+            AUTOMAKE_VERSION=$(ls /usr/local/bin/automake-* | sed 's:.*/automake-::' | sort -V | tail -1)
+            export AUTOCONF_VERSION AUTOMAKE_VERSION
+            echo "autoconf $AUTOCONF_VERSION, automake $AUTOMAKE_VERSION"
+            uname -a
+            tools/setup
+            ./configure --enable-libtap
+            gmake
+            cd lib && gmake test
+
+  netbsd:
+    name: Running build and unit tests on NetBSD
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v7
+      - name: Build and test in a NetBSD VM
+        uses: vmactions/netbsd-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          prepare: |
+            /usr/sbin/pkg_add -u autoconf automake gmake perl
+          run: |
+            set -e
+            uname -a
+            tools/setup
+            ./configure --enable-libtap
+            gmake
+            cd lib && gmake test
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -27,4 +27,4 @@ libmonitoringplug_a_SOURCES += parse_ini.c extra_opts.c
 endif USE_PARSE_INI
 
 test test-debug:
-	cd tests && make $@
+	cd tests && $(MAKE) $@
--- a/plugins-root/check_icmp.d/config.h
+++ b/plugins-root/check_icmp.d/config.h
@@ -3,6 +3,7 @@
 #include "../../config.h"
 #include "../../lib/states.h"
 #include <stddef.h>
+#include <sys/time.h>
 #include <netinet/in_systm.h>
 #include <netinet/in.h>
 #include <netinet/ip.h>


RincewindsHat:
Hi,
Thank you for this. I thought it would be more of a hassle to integrate more operating systems here.
And thank you for the other findings.
You are correct about `lib/tests/test_generic_output` the binary should not be there.

I am going to do some research about that and maybe pull it out of this PR to keep things a bit separated.

Running heavy weight tests is probably a bit much to do it all the time, so a schedule (as you suggested) might be the way (and a possibility to trigger them manually).

Regarding the OpenBSD version I have to ask some other people whether this is a bad/good/great idea to do it like this.

Regarding Solaris: I would be interested to add that later if it is possible in a reasonable manner.

neilpang:
Thanks for looking at it so quickly.

On triggering: `workflow_dispatch` is already in there (with the same
`debug_enabled` tmate input `test.yml` uses), so the manual half exists today.
The only change your suggestion needs is swapping `push` + `pull_request` for a
`schedule:`.

One trade-off worth naming before you pick: schedule-only means these jobs stop
gating anything, so a BSD regression surfaces on the next scheduled run rather
than on the PR that caused it -- and at that point it is a later, unrelated PR
that shows red. Keeping `pull_request:` and dropping only the `push:` trigger
would cut most of the cost while preserving the pre-merge signal. I am happy
with either; it is a couple of lines. Just say which and I will push it.

Pulling the `test_generic_output` removal into its own change makes sense. One
thing that may save you some research: it cannot simply be added to
`EXTRA_TEST`, because `lib/output.c:498` calls `xasprintf()`, which is defined
in `plugins/utils.c`. `lib/tests` links only `libmonitoringplug.a` and
`libgnu.a`, not the `libnpcommon.a` that carries it, so it fails at link time on
Linux too, not just on the BSDs. I tried that first and backed it out.

On the OpenBSD pin, for whoever you ask: it is not about wanting one particular
OpenBSD release. OpenBSD ships ~18 parallel `autoconf` packages and dispatches
between them through the `metaauto` wrappers, so `pkg_add` needs an exact
`autoconf-2.71p0`-style name and `AUTOCONF_VERSION`/`AUTOMAKE_VERSION` have to
agree with it. Pinning `release:` is only what keeps those in sync as the
action's default OpenBSD moves. If someone knows a tidier way to say "newest
autoconf" there, I would happily switch to it.

Solaris: the offer stands whenever you want it.

RincewindsHat:
ok, pls rebase, I fixed the test.

For now I would like to have the jobs on manual dispatch only.

I would like to have solaris too, but in a new PR then.

Bonus points for answering in middle english poetry from now on.

.github/workflows/test-unix.yml
```diff
@@ -0,0 +1,100 @@
...
+          # dispatches between them through the metaauto wrappers, so both the
+          # package and the AUTO*_VERSION below have to name one explicitly.
+          # The release is pinned to keep the two in sync.
+          prepare: |
+            pkg_add -I autoconf-2.71p0 automake-1.16.5p0 gmake
+          run: |
+            set -e
+            AUTOCONF_VERSION=2.71
+            AUTOMAKE_VERSION=1.16
+            export AUTOCONF_VERSION AUTOMAKE_VERSION
```
oxzi: Can the `AUTO{CONF,MAKE}_VERSION` be replaced with the `gnu` `CONFIGURE_STYLE` as the OpenBSD ports did upstream for the `monitoring-plugins` a while ago?

https://github.com/openbsd/ports/commit/9f9dd90196d1081138df997f4d9c848225e79f6d

Doing so would allow getting rid of pinning the OpenBSD version.

neilpang:
Rebased. The binary and .gitignore hunks are gone -- #2305 handled both
upstream -- so this is the workflow plus two portability fixes:
sys/time.h in check_icmp.d/config.h, and $(MAKE) instead of a bare make
in lib/Makefile.am.

Triggers are workflow_dispatch only. Solaris in a separate PR.

Since dispatch-only means pushing runs nothing, I verified with a copy
that differs in one place: it still has the push/pull_request block
under "on:". Everything from "jobs:" onward is byte-identical to this
commit.

  https://github.com/neilpang/monitoring-plugins/actions/runs/30870072279

  FreeBSD 15.1-RELEASE-p1    Files=10, Tests=265, Result: PASS
  OpenBSD 7.9                Files=10, Tests=265, Result: PASS
  NetBSD 10.1                Files=10, Tests=265, Result: PASS

.github/workflows/test-unix.yml
oxzi: @neilpang: Thanks for rebasing this PR. Could you also address this comment? Otherwise, the OpenBSD version will need to be updated every six months, which, let's be honest, is not going to happen.

RincewindsHat:
please address the concerns of @oxzi above, too.

neilpang:
@oxzi Done -- the OpenBSD job pins nothing now.

One thing worth flagging about that ports commit: it dropped
CONFIGURE_STYLE=autoreconf in the same change that moved the distfile
from the GitHub tag to the release tarball. That tarball is an autotools
dist, so it already carries a generated configure and the port stopped
needing autoreconf at all. CI builds a git checkout, where tools/setup
has to run autoreconf, so the autotools versions still matter here.

What replaces the pins: pkg_add takes the newest the mirror offers, and
AUTOCONF_VERSION/AUTOMAKE_VERSION are read back off the binary it
installed, since metaauto dispatches on exactly that name. Nothing to
bump per release.

  newest() {
    pkg_info -Q "$1" | grep -E "^$1-[0-9]" | sort -V | tail -1
  }
  pkg_add -I "$(newest autoconf)" "$(newest automake)" gmake

The pins were already behind, incidentally: the mirror carries
autoconf-2.72p0 and automake-1.18.1 against the 2.71p0 and 1.16.5p0 that
were in the branch.

Verified, all three green:
https://github.com/neilpang/monitoring-plugins/actions/runs/30911208131

  OpenBSD 7.9    Files=10, Tests=265, Result: PASS
  FreeBSD 15.1   Files=10, Tests=265, Result: PASS
  NetBSD 11.0    Files=10, Tests=265, Result: PASS

The OpenBSD job resolved autoconf 2.72 and automake 1.18 at run time.
The NetBSD image moved from 10.1 to 11.0 between this morning's runs and
this one, which is the same drift a pinned release needs hand-holding
for.

RincewindsHat:
@oxzi does this sufficiently address your concerns?

oxzi:
> @oxzi does this sufficiently address your concerns?

While it wasn't done quite the way I had in mind, it works. So who am I to complain?

---

Another thing: I just wanted to check the action output and was surprised to not find it here in the PR. Compared to the other actions, this one has neither `on.push` nor `on.pull_request`.

Compare the following:
https://github.com/monitoring-plugins/monitoring-plugins/blob/7149adb6745cf04acf8d3ee4c84a72e4681c3f2d/.github/workflows/test.yml#L4-L15

https://github.com/neilpang/monitoring-plugins/blob/92ac9a483654465b87e4f77ae84df4d3e5b829a5/.github/workflows/test-unix.yml#L14-L21

Is this intended? Does it still magically work or does one has to invoke this by hand (which will never happen)?

RincewindsHat:
@oxzi As discussed earlier in this thread there is not automatic trigger for workflows (per my wishes) since they are rather heavy and I did not want to run them all the time for everything.
I do see your point on them not being used at all, but I did not yet see/find a good solution (say "necessary before a PR gets merged")

neilpang:
Numbers first, since "heavy" is the premise here. Per-job wall time,
measured the same way for both:

  test-unix.yml              test.yml, which runs on every push and PR
    FreeBSD     2 min          debian:stable    7 min
    OpenBSD     4 min          rockylinux:8     5 min
    NetBSD     14 min          almalinux:9      3 min
                               fedora:latest    2 min
    total      20 min          total           17 min

So the three BSDs cost about what the Linux set already costs, and that
one runs on every push and every pull request today. NetBSD is the only
wildcard: it took 3 minutes in the run before that one and 14 in this
one.

If you want it gated rather than continuous, merge_group is the literal
form of "necessary before a PR gets merged". The docs: "Runs your
workflow when a pull request is added to a merge queue". It fires once
per merge attempt rather than once per push, and it can be a required
check for the queue. On the workflow side that is one line next to
workflow_dispatch:

  merge_group:

The rest is a repo setting, a merge queue on master.

If you would rather not touch repo settings, test-next.yml in this repo
is already the pattern for heavy jobs:

  push:
    branches-ignore:
      - '*'
  schedule:
    - cron: '0 9 * * 1'

No new concepts, but it gates nothing: a BSD regression turns up the
following Monday, in whatever PR happens to be open then, which is the
thing @oxzi is pointing at.

Or per-PR but opt-in:

  pull_request:
    types: [opened, synchronize, labeled]

with a job-level

  if: contains(github.event.pull_request.labels.*.name, 'test-bsd')

You add the label when a PR touches something that could break a BSD.
No repo settings, no schedule.

Say which one and I will push it.

RincewindsHat:
ok, we had a short internal discussion about that and countrary to my earlier statements we will go for "just run it everytime", so same as the other tests.

Sorry about the confusion.

neilpang:
Done, pushed. Triggers are now push + pull_request + workflow_dispatch,
the block copied from test.yml.

Verified on a branch first, so this PR gets one commit rather than a red
run: the push fired the workflow, all three green, and the repo's own
Tests and Spellcheck were green on the same commit.
https://github.com/neilpang/monitoring-plugins/actions/runs/31174082357

  FreeBSD   13 min 41 s
  NetBSD     3 min 26 s
  OpenBSD    4 min 28 s

One correction to what I said this morning: I called NetBSD the wildcard,
and this run has it finishing in 3 minutes while FreeBSD took 14. So it
is not one particular system that is slow, it is whether the prepared-VM
cache is warm. The total is the same either way -- about 21 runner-minutes
here against the 20 I quoted.

RincewindsHat (approved):

RincewindsHat:
Thanks a lot :-)

PR #2307, opened 2026-07-30
Fix check_ups (broken since v3.0.0)

sbraz:
Hi @RincewindsHat,
#2105 broke  `check_ups`. The following  commits fix it:

- **check_ups: fix inverted check for missing UPS name**

  The refactor in 29b1be07 dropped the negation when moving
  validate_arguments() over to the config wrapper, so the check now fires
  when a UPS name *is* given, making -u unusable.

- **check_ups: actually report load, temperature, real power and status**

  Status, load, temperature and real power are only ever handed to
  mp_add_subcheck_to_check() in their error branches, so unless the query
  itself fails, they are computed and then thrown away; the plugin reports
  a single battery line and nothing else. Load, temperature and real power
  lose their perfdata too.

  Also:

    - sc_retrieve_status has no state at all, and an unset subcheck state
      means UNKNOWN, which propagates to the whole check. Every run comes
      back UNKNOWN regardless of the UPS.
    - the real power branch writes its output into sc_temperature, dropping
      the temperature text.
    - mp_set_subcheck_default_state() returns the subcheck by value, but
      the load one ignores the return, so the default never applies.
    - the status string is built by appending to itself and starts out
      NULL, giving "Status: (null)Online, Charging".
    - stray spacing in the load label.

  All of this came in with 29b1be07.

### DIFF
--- a/plugins/check_ups.c
+++ b/plugins/check_ups.c
@@ -653,7 +653,7 @@ check_ups_config_wrapper process_arguments(int argc, char **argv) {
 }
 
 check_ups_config_wrapper validate_arguments(check_ups_config_wrapper config_wrapper) {
-	if (config_wrapper.config.ups_name) {
+	if (config_wrapper.config.ups_name == NULL) {
 		printf("%s\n", _("Error : no UPS indicated"));
 		config_wrapper.errorcode = ERROR;
 	}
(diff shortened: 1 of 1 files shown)


RincewindsHat:
Hi @sbraz,
Mea culpa I guess, I should have been more careful.
Thank you for the changes, I will take a look at it.

sbraz:
No problem. After patching it, the check works just fine, so my monitoring is green again!
And at least I learnt about the new multiline output format :)

RincewindsHat (approved):
Looks good to me :-)
Thank you

PR #2319, opened 2026-08-08
check_snmp: fix two build failures against net-snmp

daeho-ro:
Two independent build failures in `check_snmp`, reproduced with net-snmp 5.9.5.2 on macOS.
Neither is platform specific. One commit each.

**1.** `_np_state_generate_key()` still tests `USE_OPENSSL`, which #2253 renamed to `MOPL_USE_OPENSSL`. `lib/utils_base.h` includes `sha256.h` only when `MOPL_USE_OPENSSL` is undefined, so an OpenSSL build compiles the gnulib branch without its declarations:

```
check_snmp_helpers.c:918:20: error: variable has incomplete type 'struct sha256_ctx'
check_snmp_helpers.c:921:3: error: use of undeclared identifier 'sha256_process_bytes'
```

**2.** `DEFAULT_SNMP_VERSION` is a legacy alias that net-snmp defines only when `NETSNMP_NO_LEGACY_DEFINITIONS` is unset:

```
check_snmp_helpers.c:130:41: error: use of undeclared identifier 'DEFAULT_SNMP_VERSION'
```

It also expands to the version number (1/2/3), not the `SNMP_VERSION_*` encoding the field expects. Compile-only change: the value is overwritten in every path, by `-P` or by the existing v2c fallback.

### DIFF
--- a/plugins/check_snmp.d/check_snmp_helpers.c
+++ b/plugins/check_snmp.d/check_snmp_helpers.c
@@ -898,7 +898,7 @@ state_key np_enable_state(char *keyname, int expected_data_version, const char *
 char *_np_state_generate_key(int argc, char **argv) {
 	unsigned char result[256];
 
-#ifdef USE_OPENSSL
+#ifdef MOPL_USE_OPENSSL
 	/*
 	 * This code path is chosen if openssl is available (which should be the most common
 	 * scenario). Alternatively, the gnulib implementation/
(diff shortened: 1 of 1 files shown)


RincewindsHat (approved):
Correct in both points, thank you.
Verifying this I actually found another `USE_OPENSSL` in `plugins/check_curl.d/check_curl_helpers.h` which I did not notice before, so thanks for that, too.

PR #2333, opened 2026-09-01
sslutils: Linter fixes and streamlining symbol names

RincewindsHat:


RincewindsHat:
once again this escalated slightly.

Basically this renames all the symbols in netutils and sslutils so they share a common prefix to be identified more easily and get more structure into the project.

While doing that I discovered some unused stuff and "improved" other components

PR #2334, opened 2026-09-04
Homogenice utils functions

RincewindsHat:
use a proper prefix for all the functions and remove some unused ones.


martialblog:
Very homoge-nice work

PR #2336, opened 2026-09-04
check_users: Discard sessions with invalid or missing pid in utmp

hillu:
The `signal(pid, 0)` check matches what who(1) from coreutils does.

(I have checked versions 9.4 and 9.7.)

### DIFF
--- a/plugins/check_users.d/users.c
+++ b/plugins/check_users.d/users.c
@@ -83,6 +83,8 @@ get_num_of_users_wrapper get_num_of_users_systemd() {
 
 #	ifdef HAVE_UTMPX_H
 #		include <utmpx.h>
+#		include <errno.h>
+#		include <signal.h>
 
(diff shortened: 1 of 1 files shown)


RincewindsHat (approved):
Looks good to me :-)

RincewindsHat:
just to satisfy my curiosity: What kind of error could occur that would trigger this case?

hillu:
All I can say for certain is that I observed this on a system where many entries had been made to the `utmp` file and the processes had been killed shortly after. My assumption is that the processes getting killed led to the entries not getting cleaned up.

`who(1)` only showed us a small number of sessions while `check_users` saw well over 100, leading to somebody being paged off-hours and getting properly confused by the different values.

RincewindsHat:
well, if this fixes that it sounds good to me :-)

PR #2343, opened 2026-09-11
check_{disk,load,ping,procs,swap}: OpenBSD Hardening

oxzi:
Add OpenBSD hardening via pledge(2) and via unveil(2) for check plugins executing other commands. Before an unveil(2) call is made, the command path is extracted from the command defines, allowing only this command to be executed.


RincewindsHat (approved):
Looks sane to me :-) Thanks

PR #2345, opened 2026-09-14
check_ping: Remove HTML Output Leftovers

oxzi:
Remove the single last HTML output from check_ping - a single `</A>` without an opening tag - and drop every mentioning of this feature. However, the flags are still accepted, but silently discarded.


RincewindsHat (approved):
+1

PR #2348, opened 2026-09-15
check_ping: Drop incomplete multiple hostname support

### LINKED ISSUE #2346: check_ping: Incomplete multiple hostname behavior
The `check_ping` check command has an incomplete and undocumented implementation to accept multiple hostnames. These hostnames might either be passed via multiple `-H` flags or as a single, comma-separated value.

https://github.com/monitoring-plugins/monitoring-plugins/blob/027dfe5129f59b68b03bf2ac3ec2cd743d969804/plugins/check_ping.c#L278-L298

The behavior of multiple hostnames is a bit unexpected; at least for me. The first ping result which is not UNKNOWN wins, all other check results, including CRITICAL, are fine. Furthermore, the check result is being rewritten to OK.

https://github.com/monitoring-plugins/monitoring-plugins/blob/027dfe5129f59b68b03bf2ac3ec2cd743d969804/plugins/check_ping.c#L177-L179

In practice, this does not really work - at least on my setup - due to the racing `timeout_alarm_handler`, which `exit(3)` on its own.

https://github.com/monitoring-plugins/monitoring-plugins/blob/027dfe5129f59b68b03bf2ac3ec2cd743d969804/lib/utils_cmd.c#L668-L684

I started a small patch trying to beautify the output, when I realized the shortcomings. Due to its both broken and undocumented state, I would suggest removing it. However, I would like to have at least some conses first.

oxzi:
The check_ping check command had an incomplete and undocumented implementation to accept multiple hostnames. These hostnames might either be passed via multiple "-H" flags or as a single, comma-separated value. The first hostname takes precedence over the following, which were only tried in case of an UNKNOWN (!) error for the prior. However, due to the timeout logic - ending the check command -, the later hostnames were not even tried.

As being both undocumented and not functional, this feature was removed. Note, the - also undocumented - option to pass ordered options after the flags was left intact, also for a single hostname.

Most parts of the diff are just a change in the indentation level, as the loop in the main function was removed.

Fixes #2346.


RincewindsHat (approved):
Looks good to me, thank you :-)

PR #2353, opened 2026-09-17
Clang format everything

RincewindsHat:


oxzi:
Just an idea, but how about adding a `clang-format` check to the GitHub Actions to ensure everything is being formatted correctly from the start? This would get rid of these reoccurring PRs and ease `git blame`.

RincewindsHat:
yes, that would be good at this point (or should have been there after the first run).

PR #2354, opened 2026-09-18
check_dig: fix strcasestr declaration

Devstellar:
`check_dig.c` includes `<ctype.h>` before `common.h`.

`common.h` includes the generated `config.h`, which defines `_GNU_SOURCE`. On glibc systems this is too late if a system header has already included `<features.h>`, causing `strcasestr()` to be used without its prototype.

With recent GCC versions this causes the build to fail with:

```text
error: implicit declaration of function ‘strcasestr’
```

Include `common.h` before `<ctype.h>` so the feature-test macros from `config.h` are defined before any system headers are included.

Tested on Linux/glibc: `check_dig` failed to compile before this change and compiles successfully after it.


RincewindsHat:
sounds reasonable. Which distro are you using? This does build on Debian for me.

Devstellar:
It's a source-built Linux system using glibc rather than a packaged distro.
The relevant difference may be the toolchain: I'm using a recent GCC (14+), where implicit function declarations are errors by default.
In this case config.h does define _GNU_SOURCE, but check_dig.c includes <ctype.h> before common.h, so glibc's <features.h> has already been processed before _GNU_SOURCE is defined.
That is why I think the include order is worth fixing even if the issue does not reproduce on your Debian build.

RincewindsHat (approved):

PR #2355, opened 2026-09-18
check_icmp: fix positional host arguments

Devstellar:
`check_icmp` still accepts positional host arguments, but after the 3.x refactor
their parsed targets are discarded.

The code currently calls `add_target()` for each positional argument without
storing the returned target information, leaving `number_of_targets` at zero
and causing:

    check_icmp: No hosts to check

Process positional hosts through `add_host()` in the same way as `-H`, updating
the host list, target list, target count, and address-family flags.

Tested with:
- one positional host
- one host using `-H`
- multiple positional hosts

All cases work correctly after the change.


Devstellar:
The Debian test job appears to have hit an unrelated/flaky check_ssh test and was eventually cancelled after 360 minutes. All other CI jobs passed. Could someone please re-run the Debian job?

PR #2356, opened 2026-09-18
Fix inclusive upper bound for negative-infinity ranges

Devstellar:
`mp_check_range()` treats the upper bound of ranges starting at negative
infinity as exclusive.

For example, with the range `~:2`, a value of `2` is currently considered
outside the range and triggers an alert.

This differs from both the documented Monitoring Plugins range semantics and
the legacy `check_range()` implementation, where range endpoints are
inclusive.

Change the comparison from `cmp_perfdata_value(...) == -1` to `<= 0`,
making the upper bound inclusive.

 And add regression tests for:
- value below the upper bound
- value equal to the upper bound
- value above the upper bound

This also fixes threshold checks such as `-C 2` incorrectly alerting when the
measured value is exactly `2`.


RincewindsHat (approved):
Indeed, got something wrong here.
Thanks

PR #2357, opened 2026-09-18
check_ntp_time: add stratum thresholds

Devstellar:
Add optional stratum threshold checks to `check_ntp_time`.

This allows monitoring the stratum advertised directly by an NTP server
without relying on ntpd mode 6 queries, making the check usable with servers
such as chronyd.

New options:

- `-W, --swarn=THRESHOLD`
- `-C, --scrit=THRESHOLD`

The stratum is taken from the NTP response of the server selected by
`check_ntp_time` as the best candidate.

The stratum subcheck is only enabled when `-W` or `-C` is specified, so the
default output and behavior remain unchanged.

Tested with:

- no stratum thresholds
- `-C 2` with stratum 2 -> OK
- `-C 1` with stratum 2 -> CRITICAL
- `-W 1 -C 2` with stratum 2 -> WARNING
- long option `--scrit=2`

Correct inclusive upper-bound handling for ranges such as `~:2` is addressed
separately in #2356.


RincewindsHat:
For clarification: This allows `check_ntp_time` to issue a warning if a server is not the stratum we expect?

Devstellar:
Yes. More precisely, this adds standard Monitoring Plugins threshold handling for the stratum advertised directly by the NTP server.
For example:
- -C 2 means stratum 1 and 2 are accepted, while stratum 3 or higher results in CRITICAL.
- -W 2:2 could be used if the intention is to warn whenever the server is not exactly stratum 2.

The main use case that motivated this is monitoring chronyd servers. check_ntp_peer relies on ntpd mode 6 control queries, which chronyd does not implement. check_ntp_time, on the other hand, already receives the server's stratum in the normal NTP response, so it can check it without any ntpd-specific control protocol.

One important distinction is that this checks the stratum advertised by the server being queried, not the stratum of that server's synchronization peer. For example, a server synchronized to a stratum 1 source will normally advertise itself as stratum 2, and this change checks that advertised value.

The stratum check is optional and is only enabled when -W or -C is specified, so the existing default behaviour of check_ntp_time is unchanged.

RincewindsHat:
Ok, so this basically is useful if "Am I talking to a stratum x server?" is my question?
Sorry to be annoying, but I am not a time admin and I don't know what the relevance is.

Devstellar:
> Ok, so this basically is useful if "Am I talking to a stratum x server?"

Yes. The main use case is detecting an unexpected fallback in the NTP hierarchy, not measuring clock accuracy.

For example, a server may normally synchronize from several stratum 1 sources and advertise itself as stratum 2, while also having higher-stratum sources available as fallback. If the preferred sources are lost, its advertised stratum will increase.

That allows different severity levels, for example:
- -W 2 to warn if the server is no longer at its normal stratum 2 level.
- -C 5 to make it critical if it falls back far enough to advertise stratum 6 or higher.

So the offset check answers “is the clock currently accurate?”, while the stratum thresholds answer “is the server still synchronizing at an expected level of the NTP hierarchy?”

RincewindsHat:
Ah, thanks for the clarification. This indeed does sound useful.

